PR #9, opened 2018-07-28
Fix plugin files to use root dir

lbb:
See: https://github.com/GoogleContainerTools/krew/pull/11

### DIFF
--- a/plugins/krew.yaml
+++ b/plugins/krew.yaml
@@ -7,11 +7,11 @@ spec:
   - uri: https://storage.googleapis.com/krew-test/krew-437272f4f52b6d114f87b366cc3d069f54ab7451dc530c9b6517139fc150f908.zip
     sha256: 437272f4f52b6d114f87b366cc3d069f54ab7451dc530c9b6517139fc150f908
     files:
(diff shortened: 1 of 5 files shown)


plugins/krew.yaml
```diff
@@ -7,35 +7,35 @@ spec:
...
     selector:
       matchExpressions:
       - {key: os, operator: In, values: [darwin]}
   - uri: https://storage.googleapis.com/krew-test/krew-437272f4f52b6d114f87b366cc3d069f54ab7451dc530c9b6517139fc150f908.zip
     sha256: 437272f4f52b6d114f87b366cc3d069f54ab7451dc530c9b6517139fc150f908
     files:
-    - from: "unix/plugin.yaml"
-      to: plugin.yaml
-    - from: "krew-linux"
+    - from: "build/unix/plugin.yaml"
```
ahmetb: the only thing I noticed is the `/foo` vs `foo/` notation.

I assume technically we could do `*/foo` as well. I think ideally we should not allow parts starting with `/` since they're not actually absolute.

PR #11, opened 2018-08-04
import of plugins from danisla

danisla:
Converting several of my kubefunc bash functions that I use frequently to plugins:

- [gke-credentials](https://github.com/danisla/kubefunc/blob/v1.0.1/gke-credentials/plugin.yaml)
- [kube-node-admin](https://github.com/danisla/kubefunc/tree/v1.0.1/kube-node-admin/plugin.yaml)
- [kube-shell](https://github.com/danisla/kubefunc/tree/v1.0.1/kube-shell/plugin.yaml)
- [pod-logs](https://github.com/danisla/kubefunc/tree/v1.0.1/pod-logs/plugin.yaml)
- [pod-shell](https://github.com/danisla/kubefunc/tree/v1.0.1/pod-shell/plugin.yaml)


plugins/pod-logs.yaml
```diff
@@ -0,0 +1,20 @@
...
+    sha256: c548f94d8efd1fb8aa35a54b7e5cdb64ae012447122d9704f3cee4163bbbe9c3
+    uri: https://github.com/danisla/kubefunc/archive/v1.0.1.zip
+    files:
+    - from: "/*/pod-logs/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  shortDescription: Get logs from a list of pods
```
ahmetb: the description doesn't make it clear how's this different than kubectl-logs.
have you checked out mtail? does it differentiate enough?

plugins/pod-shell.yaml
```diff
@@ -0,0 +1,20 @@
...
+    sha256: c548f94d8efd1fb8aa35a54b7e5cdb64ae012447122d9704f3cee4163bbbe9c3
+    uri: https://github.com/danisla/kubefunc/archive/v1.0.1.zip
+    files:
+    - from: "/*/pod-shell/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  shortDescription: Run shell command in pod
```
ahmetb: how's the use case for this look like? (trying to see what it adds over kubectl exec). do you mind providing a long description field with some examples for when the shortDescription isn't explanatory enough like this?

plugins/install-helm.yaml
```diff
@@ -0,0 +1,19 @@
...
+    sha256: c548f94d8efd1fb8aa35a54b7e5cdb64ae012447122d9704f3cee4163bbbe9c3
+    uri: https://github.com/danisla/kubefunc/archive/v1.0.1.zip
+    files:
+    - from: "/*/install-helm/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  shortDescription: Install helm package manager
```
ahmetb: Similarly, plugins ideally should not be one-off use like installing Helm.

If there's something like `curl https://helm.sh/install.sh | bash`, users should use that.

plugins/install-cert-manager.yaml
```diff
@@ -0,0 +1,20 @@
...
+    sha256: c548f94d8efd1fb8aa35a54b7e5cdb64ae012447122d9704f3cee4163bbbe9c3
+    uri: https://github.com/danisla/kubefunc/archive/v1.0.1.zip
+    files:
+    - from: "/*/install-cert-manager/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  shortDescription: Install cert-manager helm chart
```
ahmetb: I think just installing a helm chart doesn't warrant a plugin, especially because it's mostly an one-off operation and not repetitive. Not to mention, it won't scale to have a plugin per helm chart :)

danisla: I use this one a lot because it also installs the cluster issuer which I always have to look up. Ya if it was just installing the chart there isn’t much value added.

plugins/install-helm.yaml
danisla: This one is really useful for me because it bundles all of the commands for installing helm with RBAC.

plugins/pod-shell.yaml
danisla: This is more convenient because it shows you a list of pods that you choose from interactively and tells you if there are multiple pods. Again, super convenient for daily stuff I find my self repeating.

plugins/pod-logs.yaml
danisla: This is very different than mtail. It presents a list to choose from. 

You should try these plugins first to gauge how they work and differ.

plugins/install-helm.yaml
danisla: Is there a rubric or documentation with guidelines on the usefulness of a plugin so that it can be admitted to the Krew-index?

lbb:
I agree with @ahmetb; some plugins are too specific.

* I like `kube-node-admin`. I think this is very useful for debugging.
* One-time install plugins don't give enough value to be installed. This might be suitable for multiple clusters. For the majority, this doesn't scale.
* I like the idea of the *-shell plugins. If we would use them they need a 
  different name. I as a user can't figure out the difference between kube,pod- 
  shell. Maybe name it debug-shell and pod-shell. But this can also be solved with a 
  simple bash alias. The question is if it brings enough value to have a custom 
  command. @ahmetb ?

We should write a rubic for plugin contributions. I'll create an issue.

danisla:
I agree that the one time plugins may seem like they don’t give enough value. Many of my use cases are in the DevOps realm and involve creating and interacting with many clusters. I would hope that others would find that useful too. 

kube-shell creates a new pod used for debugging services or the cluster itself and is not dependent on any existing pods. Whereas pod-shell locates an existing pod and runs a shell in it. 

I think we should prioritize the tapping of other Krew-indices so that others can use Krew to manage their plugins without a centralized authority deciding a plugin’s  relevance.

plugins/install-cert-manager.yaml
ahmetb: I think this is still a helm chart problem.

I've brought this up in https://github.com/jetstack/cert-manager/issues/259 so please participate in the discussion and support resolution of this issue there.

plugins/install-helm.yaml
ahmetb: Is the RBAC problem because GKE does not automatically create `cluster-admin` binding to your account (I believe this will be fixed soon) OR is it because helm installation is not creating the `tiller` service account for you?

plugins/pod-shell.yaml
ahmetb: That makes sense, we should get this in.

plugins/pod-logs.yaml
ahmetb: Makes sense, we should get this in. Do you foresee a more original name for this or are you happy to use this name (hint: this will impede discoverability for this plugin via search engines)

ahmetb:
@danisla I think the problems in this PR are:

1. We don't have well defined guidelines on how to enforce style on the centralized krew-index repo. Therefore we're trying to be careful of people putting dibs on generic words/names, or plugins that aren't super useful (such as two one-off commands combined together). 

  Note that the plugin system is in its SUPER EARLY days, therefore we need input from "what is a plugin and what shouldn't be a plugin". My opinion is tipping the scale towards "tools that are used frequently as opposed to one-off install scripts".

2. We're evaluating 7 plugins at once in one PR. :) I think some of them are fine. If you want to send them one by one I think we can get them in sooner.

ahmetb:
I created https://github.com/GoogleContainerTools/krew/issues/23 to keep track of custom index repository support. Please subscribe.

plugins/install-helm.yaml
danisla: it's both the cluster-admin binding and the tiller service-account and rbac role.

plugins/install-cert-manager.yaml
danisla: It sounds like it is a helm problem, no clear simple resolution in sight other than having 2 helm charts for the issuers... I would still use my one-liner plugin to do it all.

ahmetb:
@danisla I still don't think the one-off installer plugins should go in. Plugins are more directed towards workflows and we probably want to have the first few plugins follow the best practices. 

I'm not against you creating a plugin to simplify an installation step, but I think as long as someone copies a cmd from somewhere to `install-cert-manager`, they might as well just the 3 commands (not to mention 3 separate commands would explain what's happening better for tutorial/workshop purposes etc).

I think we should prioritize custom index repos for this. Feel free to remove one-off plugins and we can merge the remaining.

plugins/kube-node-admin.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - from: "/*/kube-node-admin/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  shortDescription: List nodes and run privileged pod with chroot
+  caveats: |
+    This plugin needs the following programs:
+    * kubectl
```
lbb: `kubectl` should not be listed as a dependency.

ahmetb:
Due to lack of CI/CD I'm not sure if the plugins are working correctly. :) So please `krew update` and try installing them.

PR #12, opened 2018-08-05
Add a new plugin "service"

superbrothers:
This PR adds a new plugin "service". This kubectl plugin opens the Kubernetes URL(s) for the specified service in your browser.

- https://github.com/superbrothers/kubectl-service-plugin

image

### DIFF
--- a/plugins/service.yaml
+++ b/plugins/service.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha1
+kind: Plugin
+metadata:
+  name: service
+spec:
+  platforms:
+  - uri: https://github.com/superbrothers/kubectl-service-plugin/releases/download/v1.1.0/service-darwin-amd64.zip
+    sha256: 8197bc5506c628381cf1ae73ec03cc44d0513f6ca102d8b7165ba7dff034208e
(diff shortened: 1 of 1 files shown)


lbb:
I like the plugin, as this uses a Go lib for the browser support. Can it be used under Windows?

IHMO service is a too generic name. Maybe the plugin should be called `open-svc` or `browse-service`?

superbrothers:
If it is possible, I want to continue to use this name. Because some people already use this plugin with this name. Furthermore I have to make many changes such as source code, repository name, and more.

ahmetb:
I'm ok with this name in the index.

But, I *strongly recommend* renaming the plugin to something original. This is a personal suggestion as I think anyone searching for "kubernetes service" will not probably see this plugin at all.

An original name like "serve" or "open-svc" can perform much better. The repo doesn't look like it has a ton of popularity yet. So please consider finding a good name for the plugin in its early days –that will help you get traction as well.

superbrothers:
> So please consider finding a good name for the plugin in **its early days** that will help you get traction as well.

You're completely right. I rename this plugin to `open-svc`, so I'll close this PR. I'll recreate a PR after renamed.

Thanks.

PR #19, opened 2018-09-03
Add a manifest for pod restart plugin

achanda:
Source here https://github.com/achanda/kubectl-restart

### DIFF
--- a/plugins/restart.yaml
+++ b/plugins/restart.yaml
@@ -1,0 +1,19 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha1
+kind: Plugin
+metadata:
+  name: "restart"
+spec:
+  platforms:
+  - uri: https://github.com/achanda/kubectl-restart/archive/v0.0.3.zip
+    sha256: d7079b79bf4e10e55ded435a2e862efe310e019b6c306a8ff04191238ef4b2b4
+    files:
+    - from: "/kubectl-restart-*/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v0.0.1"
+  shortDescription: "Restarts a pod with the given name"
+  description: |
+    Restarts a pod with the given name. The existing pod
+    will be deleted and created again, not a true restart.


plugins/restart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha1
+kind: Plugin
+metadata:
+  name: "restart"
+spec:
+  platforms:
+  - head: https://github.com/achanda/kubectl-restart/archive/v0.0.1.zip
```
ahmetb: Since you have a version tag, it would be better if you use `url` and therefore `sha256` fields.

`head` is for moving targets (like `master` branch).

plugins/restart.yaml
```diff
@@ -0,0 +1,20 @@
...
+    files:
+    - from: "/kubectl-restart-0.0.1/*"
```
ahmetb: You can probably do `/kubectl-restart-*/*` here so the newer versions can easily be maintained without updating this. But don't forget to test this with `kubectl plugin install --source=[...]`

plugins/restart.yaml
```diff
@@ -0,0 +1,20 @@
...
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v0.0.1"
+  shortDescription: "Restarts a pod with the given name"
+  description: |
+    Restarts a pod with the given name. The existing pod
+    will be deleted and created again, not a true restart.
+  caveats: |
+    This plugin needs bash
```
ahmetb: I think you can leave this caveat out since we show this post-installation and it's extremely likely darwin/linux will have bash.

However you should list if you need a specific minimum bash version like 4.x.

achanda:
Thanks @ahmetb , updated the PR.

plugins/restart.yaml
```diff
@@ -0,0 +1,24 @@
...
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v0.0.1"
+  shortDescription: "Restarts a pod with the given name"
+  description: |
+    Restarts a pod with the given name. The existing pod
+    will be deleted and created again, not a true restart.
+  flags:
+    - name : "namespace"
```
ahmetb: You should not define this flag. kubectl will provide this to you as an environment variable, when called like `kubectl plugin -n=FOO restart`. 

See [docs](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/):

> `KUBECTL_PLUGINS_CURRENT_NAMESPACE`: The current namespace that is the context for this call. This is the actual namespace to be used, meaning it was already processed in terms of the precedence between what was provided through the kubeconfig, the --namespace global flag, environment variables, and so on.

This way, you don't need to handle `-n` vs `--namespace` and its position etc. Kubectl is supposed to do this for you.

achanda:
Thanks, tagged a new release and updated here.

ahmetb:
Thanks! Hoping to see more from this plugin!

If we get any pushback regarding the "too broad" name, I might suggest you to rename it to "restart-pod" but I reckon that won't happen anytime soon.

achanda:
Fir enough, thanks!

PR #28, opened 2018-10-17
krew 0.2.0

ahmetb:


plugins/krew.yaml
```diff
@@ -1,47 +1,76 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
   name: krew
 spec:
+  version: "652e113"
```
lbb: Why is the version a git tag? What do you think about making it the tag version? I think it is more human readable

ahmetb: @lbb this was a build configuration mistake and was fixed in https://github.com/GoogleContainerTools/krew/pull/121.

If we release a newer version of krew, it'll be picked up. So for now it's not worth diverging from the version of this file included in the release, mostly b/c it's not visible to users.

PR #32, opened 2018-11-02
Update rbac-view

jasonrichardsmith:


jasonrichardsmith:
Please disregard, this was a test for automation of updating the index when plugin gets updated.

PR #36, opened 2018-11-17
adding sniff yaml file after testing

eldadru:


eldadru:
I signed it!

plugins/sniff.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: sniff
+spec:
+  version: "v1.0.0"
```
ahmetb: This should be 1.1.0 but it’s optional. Feel free to drop it.

plugins/sniff.yaml
```diff
@@ -0,0 +1,53 @@
...
+        os: linux
+        arch: amd64
+  - uri: https://github.com/eldadru/ksniff/releases/download/v1.1.0/ksniff.zip
+    sha256: afb1a2e2cd093f1c8f8fff511f48cc5a290d2c6ecd18d9f51f9c66500710297b
+    bin: kubectl-sniff.exe
+    files:
+    - from: ./kubectl-sniff-windows
+      to: kubectl-sniff.exe
+    - from: ./static-tcpdump
+      to: static-tcpdump
```
ahmetb: This probably should be an exe

plugins/sniff.yaml
```diff
@@ -0,0 +1,53 @@
...
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  - uri: https://github.com/eldadru/ksniff/releases/download/v1.1.0/ksniff.zip
+    sha256: afb1a2e2cd093f1c8f8fff511f48cc5a290d2c6ecd18d9f51f9c66500710297b
+    bin: kubectl-sniff.exe
+    files:
+    - from: ./kubectl-sniff-windows
```
ahmetb: Are you sure your binary doesn’t have an exe in the zip?

Can you please test install this manifest with KREW_OS=windows override OR on a windows machine?

plugins/sniff.yaml
eldadru: static-tcpdump is always a ELF, it's not executed locally but in a remote container running on kubernetes.

plugins/sniff.yaml
eldadru: The binary on the zip for windows is "kubectl-snifff-windows" (which is a PE without the exe suffix)
I tested it with KREW_OS=windows:
➜  ~ KREW_OS=windows k krew install --manifest=sniff.yaml
Installing plugin: sniff
CAVEATS:
\
 |  This plugin needs the following programs:
 |  * wireshark (optional, used for live capture)
/
Installed plugin: sniff

As you can see here:

➜  afb1a2e2cd093f1c8f8fff511f48cc5a290d2c6ecd18d9f51f9c66500710297b ls
kubectl-sniff.exe  static-tcpdump

the files copied with the exe suffix which will do the work for windows machines.

plugins/sniff.yaml
eldadru: changed to 1.1.0

PR #41, opened 2018-12-22
Add cssh plugin v1.0.0

normanjoyner:
This PR adds [cssh plugin](https://github.com/containership/kubectl-cssh) v1.0.0 to krew as "cssh", which can be invoked with `kubectl cssh`.

-----

### DIFF
--- a/plugins/cssh.yaml
+++ b/plugins/cssh.yaml
@@ -1,0 +1,25 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: cssh
+spec:
+  version: "v1.0.0"
+  platforms:
+  - uri: https://github.com/containership/kubectl-cssh/archive/v1.0.0.tar.gz
+    sha256: c638dcc802d2d52ce0ec2283a2f9d217eaa17855f3edd702229571da39c1ab84
+    bin: kubectl-cssh
+    files:
+    - from: "/*/parse-args.sh"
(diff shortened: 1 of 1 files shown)


normanjoyner:
I unfortunately don’t have a more original name in mind. I am struggling to come up with something more specific than “ssh”, as that’s exactly what the plugin does.

ahmetb:
Found another plugin by the name `ssh` already. https://github.com/jordanwilson230/kubectl-plugins/blob/23da33ff9d827a178bf131644ffe3dd6ef44b724/kubectl-ssh This is sort of a problem with plugins with very common names, and we are trying to prevent first-comer advantage with this rule.

For example, there are plugins like `kail` or `mtail` to show logs, they don't say "logs" but they still do the job. I recommend going that route.

normanjoyner:
hey @ahmetb any update on this? would you prefer `ssh-node`?

ahmetb:
Let's go with "cssh" if you don't mind. It has more branding to it. Please make sure to document it well on you repo, and include a link to docs on how it works from "caveats" field in krew plugin spec.

We're very likely to go with a repo-centric model in the future which won't result in such hassles to resolve naming collisions.

jordanwilson230:
@normanjoyner What are your thoughts on https://github.com/containership/kubectl-ssh/issues/18 ?

normanjoyner:
@ahmetb updated to `cssh` and added documentation to the plugin repository.

PR #44, opened 2019-01-09
Add stem plugin

ynqa:
Add a tailing logs plugin like [stern](https://github.com/wercker/stern/tree/master/stern) which is named stem.
- https://github.com/ynqa/stem, v0.1.0

Checked installation with the following command:
```
$ kubectl krew install --manifest=plugins/stem.yaml
```

-----

### DIFF
--- a/plugins/stem.yaml
+++ b/plugins/stem.yaml
@@ -1,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: stem
+spec:
+  version: v0.1.0
+  shortDescription: Tailing logs for multiple pods
+  platforms:
+  - uri: https://github.com/ynqa/stem/releases/download/v0.1.0/stem_0.1.0_darwin_amd64.tar.gz
+    sha256: 84be42645c16f5857316fdd3e7b0b0251aca8e4a61f22f410f39ba3b46d57a45
+    bin: kubectl-stem
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  - uri: https://github.com/ynqa/stem/releases/download/v0.1.0/stem_0.1.0_linux_amd64.tar.gz
+    sha256: eea0364aaaa4c3b9e0b4405836ae93d323c3de422dd4687dd79d4aae012640bc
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks a lot! 👋 

I just installed this and run `kubectl stem` it just hanged there.

I recommend you use the `caveats:` field (see other plugins) to give a quick note:

- giving an example command
- pointing to the documentation (can be showing --help or linking to the repo)

we show the caveats section just after the installation. Do you mind making this change?

ahmetb:
By the way, is this command just `stern`? How does it differ than `stern`?

If so, we should package the stern repo and not your fork of it.

ahmetb:
> The main difference from the above projects is that it is able to use as sub-command of kubectl via kubectl plugin mechanism.

This should not be a reason to fork tools. stern works fine as a standalone tool. you can `brew install stern` or `go get github.com/foo/stern` and just call it like `stern [...]`.

Also when I ran --help, it seemed different set of flags than stern. So I'm not sure if the only difference is the kubectl packaging.

I'm inclined to close this as it stands right now.

ynqa:
@ahmetb I'd like to tail logs with kubectl. But if you feel unfavorable for this, please close it. Thanks.

> it seemed different set of flags than stern
It is owing to kubernetes/cli-runtime, which adds basic flags for kubectl.

ahmetb:
@ynqa Can you please explain the story behind the project? Is this a direct fork of stern (github doesn't show so)? What % of the code from stern, and is the difference limied to command-line flags?

ahmetb:
> Almost stern codes, differences are command-line flags now.

If all you did was copying the source code and integrating the cli-runtime, you probably should have forked `stern` project on GitHub and made the change.

Right now, your repo only has 1 commit adding 731,000+ lines of code, so I don't think your users will be able to go audit and trust the software you're distributing. 

I'm hesitant about merging this. I understand a tool using `cli-runtime` is definitely greater than blindly integrating stern into krew.

But I don't think you should fork "stern", change just the cmdline flags, and distribute it in a different name. I recommend working with the "stern" maintainer and proposing them to:

1. convert project to cli-runtime
2. distribute as plugin (there are not a lot of reasons to do this maybe)

ynqa:
Right. Thanks for closing PR, and I'll also delete this repository.

PR #45, opened 2019-01-09
Add exec-as plugin

jordanwilson230:
Adds the ```exec-as``` plugin seen [here](https://github.com/jordanwilson230/kubectl-plugins#kubectl-ssh). Note, in the link above, it is named 'ssh'.

Verified installation via release url.

This plugin allows the user to specify a username (such as root) when exec'ing into a pod.

### DIFF
--- a/plugins/connect.yaml
+++ b/plugins/connect.yaml
@@ -1,0 +1,20 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: connect
+spec:
+  platforms:
+  - head: https://github.com/jordanwilson230/kubectl-plugins/archive/krew.zip
+    uri: https://github.com/jordanwilson230/kubectl-plugins/archive/v1.0.0-krew.zip
+    sha256: 21099ad3c3a6509a4ddb94aa0a88c055437e369c41c0e5c0cae49dc6db697241
+    bin: kubectl-connect
+    files:
+    - from: "**/kubectl-connect"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.0-krew"
+  caveats: The node which the pod is running on cannot have more than one taint.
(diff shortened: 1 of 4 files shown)


jordanwilson230:
I signed it!

jordanwilson230:
> I thought you were considering of a name like `cssh`? Don't you think `connect` word is too broad? A name with a noun like `connect-node` might work better here (see Naming Guide).

@ahmetb I think you're confusing this with https://github.com/GoogleContainerTools/krew-index/pull/41

In the meantime, I will look for a new name for connect and update.

ahmetb:
Let's hold off onto these. These words are _really_ far too broad and common. I don't want this to end in a major name-grab.

cc: @juanvallejo

jordanwilson230:
@ahmetb no prob.  I can switch it to pssh or pod-exec. Asside from "connect", what are you refering to when you say "these words"?

ahmetb:
> what are you refering to when you say "these words"?

sorry I conflated the "ssh" plugin with this one, my bad.

> Renamed plugin to `pod-exec`

I recommend empathizing with the potential user here.

You look at the list of plugins, you see `pod-exec`, what would you think? I'd personally say "how's this different than `exec`?".

You actually answered it `shortDescription: Like kubectl exec, but offers a `user` flag to exec as root or any other user`.

So why not call it `exec-as` or `exec-as-user`?

juanvallejo:
> So why not call it exec-as or pod-exec-as-user?

Agree, it'd be nice to disambiguate the added functionality provided by this plugin for the user. I'd be favor of either of these as well (maybe even `user-exec`?)

jordanwilson230:
Thanks, @juanvallejo and @ahmetb, user-exec works.  I'll update once its changed.  @ahmetb I believe that a names like 'exec-as' violate the naming convention that kubectl has in place, which prevent plugin names from colliding with their commands.

jordanwilson230:
@ahmetb @juanvallejo Changes pushed.  New name is user-exec.

ahmetb:
Thanks for accommodating the changes. Welcome!

PR #46, opened 2019-01-09
add prompt plugin v1.0.0-krew

jordanwilson230:
Adds the prompt plugin seen [here](https://github.com/jordanwilson230/kubectl-plugins#kubectl-prompt).

Verified installation via release url.

This plugin allows a user to flag a certain namespace or cluster, which will then prompt for confirmation when executing commands in the given environment.

-----

- [ ✔️ ] Read the [Plugin Naming Guide](https://github.com/GoogleContainerTools/krew/tree/master/docs/NAMING_GUIDE.md) (for new plugins)
- [ ✔️  ] Verify the installation from URL or a local archive works (`kubectl krew install --manifest=[...] --archive=[...]`)

### DIFF
--- a/plugins/prompt.yaml
+++ b/plugins/prompt.yaml
@@ -6,7 +6,7 @@ spec:
   platforms:
   - head: https://github.com/jordanwilson230/kubectl-plugins/archive/krew.zip
     uri: https://github.com/jordanwilson230/kubectl-plugins/archive/v1.0.0-krew.zip
-    sha256: d12d3c0c9cf2559100adbd7bd032ecf1fc01e09cb8731726ead42a49d6104a4d
+    sha256: 708c3c91cf0be5d859797b13e85ba4a0b9f74921d4e3a4db65de914a07133a17
     bin: kubectl-prompt
     files:
     - from: "*/kubectl-prompt"
--- a/plugins/prompt.yaml
+++ b/plugins/prompt.yaml
@@ -6,7 +6,7 @@ spec:
   platforms:
   - head: https://github.com/jordanwilson230/kubectl-plugins/archive/krew.zip
     uri: https://github.com/jordanwilson230/kubectl-plugins/archive/v1.0.0-krew.zip
-    sha256: 21099ad3c3a6509a4ddb94aa0a88c055437e369c41c0e5c0cae49dc6db697241
+    sha256: c020d4d283461ef9f103fd0baa218e9c72612395ff936fd337e165d054cee699
     bin: kubectl-prompt
     files:
     - from: "*/kubectl-prompt"
--- a/plugins/prompt.yaml
+++ b/plugins/prompt.yaml
@@ -6,7 +6,7 @@ spec:
   platforms:
   - head: https://github.com/jordanwilson230/kubectl-plugins/archive/krew.zip
     uri: https://github.com/jordanwilson230/kubectl-plugins/archive/v1.0.0-krew.zip
-    sha256: c020d4d283461ef9f103fd0baa218e9c72612395ff936fd337e165d054cee699
+    sha256: d12d3c0c9cf2559100adbd7bd032ecf1fc01e09cb8731726ead42a49d6104a4d
     bin: kubectl-prompt
     files:
     - from: "*/kubectl-prompt"
--- a/plugins/prompt.yaml
+++ b/plugins/prompt.yaml
@@ -15,6 +15,18 @@ spec:
       matchExpressions:
       - {key: os, operator: In, values: [darwin, linux]}
   version: "v1.0.0-krew"
-  caveats: When removing this plugin, delete the line beginning with `function kubectl()` in your ~/.bash_profile
+  caveats: |
+    This plugin requires bash.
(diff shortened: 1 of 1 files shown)


plugins/prompt.yaml
```diff
@@ -0,0 +1,20 @@
...
+    uri: https://github.com/jordanwilson230/kubectl-plugins/archive/v1.0.0-krew.zip
+    sha256: 21099ad3c3a6509a4ddb94aa0a88c055437e369c41c0e5c0cae49dc6db697241
+    bin: kubectl-prompt
+    files:
+    - from: "*/kubectl-prompt"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.0-krew"
```
ahmetb: I recommend just doing this v1.0.0. we may start using semver soon to determine recentness (right now we rely on sha256).

plugins/prompt.yaml
```diff
@@ -0,0 +1,20 @@
...
+  caveats: When removing this plugin, delete the line beginning with `function kubectl()` in your ~/.bash_profile
+  shortDescription: Prompts for user confirmation when executing commands in critical namespaces or clusters, i.e., production.
+  description: Protect your production environments! When issuing certain commands inside a specified namespace/cluster, a warning prompt appears, telling the user they are in a flagged environment and asks to confirm the command. Ex., `kubectl prompt`, `kubectl add -n production`, kubectl add -c `prod-cluster`. `kubectl prompt list`.
```
ahmetb: `add` doesn't exist as a plugin?

also you can use yaml multi-line features to split this to shorter lines

plugins/prompt.yaml
jordanwilson230: Since I'm cutting these releases off a branch of my kubectl-plugins repo, that might be a bit confusing to be releasing a v1.0.0 (vs v1.0.0-krew) that is specific only for Krew.  Does Krew recommend housing these contributions in a separate repo (i.e., outside jordanwilson230/kubectl-plugins)?

plugins/prompt.yaml
jordanwilson230: resolved. good eye.

plugins/prompt.yaml
juanvallejo: > Since I'm cutting these releases off a branch of my kubectl-plugins repo, that might be a bit confusing to be releasing a v1.0.0 (vs v1.0.0-krew) that is specific only for Krew

I think it should be fine, given that the `platforms.uri` field still points to the actual `v1.0.0-krew` location.

plugins/prompt.yaml
```diff
@@ -0,0 +1,20 @@
...
+    sha256: efa1c2122ecaeb0a43ba1ed9e54c1ff1274d20de76934f7bb91dd154f2327aa1
+    bin: kubectl-prompt
+    files:
+    - from: "*/kubectl-prompt"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.0-krew"
+  caveats: When removing this plugin, delete the line beginning with `function kubectl()` in your ~/.bash_profile
```
ahmetb: can you add a caveat here saying

> This plugin only works for bash.

plugins/prompt.yaml
```diff
@@ -0,0 +1,20 @@
...
+  shortDescription: Prompts for user confirmation when executing commands in critical namespaces or clusters, i.e., production.
+  description: Protect your production environments! When issuing certain commands inside a specified namespace/cluster, a warning prompt appears, telling the user they are in a flagged environment and asks to confirm the command. Ex., `kubectl prompt`, `kubectl prompt add -n production`, kubectl prompt add -c `prod-cluster`. `kubectl prompt list`.
```
ahmetb: can you wrap these lines by using yaml multiline string https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

plugins/prompt.yaml
ahmetb: also the `export KUBECTL_*_PROMPT` env var?

ahmetb: also you can say

> Visit https://github.com/jordanwilson230/kubectl-plugins#kubectl-prompt for documentation.

jordanwilson230:
Thanks for the suggestions @ahmetb.  They changes have been pushed.

plugins/prompt.yaml
jordanwilson230: Done

plugins/prompt.yaml
jordanwilson230: Done

plugins/prompt.yaml
jordanwilson230: Done

ahmetb:
Thanks for pushing through! Welcome to krew.

jordanwilson230:
> Thanks for pushing through! Welcome to krew.

Thanks, @ahmetb!

PR #62, opened 2019-02-05
Adding kube-capacity

robscott:
I'm not really sure what to name this. The upstream repo is just [kube-capacity](https://github.com/robscott/kube-capacity) which the naming guide specifically says to avoid for a plugin. Capacity is probably not specific enough. Happy to change this to something like `resource-capacity` here if that makes more sense.

### DIFF
--- a/plugins/resource-capacity.yaml
+++ b/plugins/resource-capacity.yaml
@@ -1,0 +1,30 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: resource-capacity
+spec:
+  platforms:
+  - uri: https://github.com/robscott/kube-capacity/releases/download/0.1.3/kube-capacity_0.1.3_Darwin_x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
I think resource-capacity or resource-summary makes sense. Definitely not kube-* as it’s listed in the naming guide.

plugins/kube-capacity.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capacity
```
ahmetb: This must match the filename.

ahmetb:
Do you mind doing `git commit --amend --no-edit && git push -f` sometime, so that the newly added testing system tests this manifest?

robscott:
Thanks for the feedback! I ended up going with resource-capacity for the sake of consistency with the standalone name. The new automated testing is pretty great.

juanvallejo:
lgtm

PR #64, opened 2019-02-05
ignore: trigger validation on existing manifests

ahmetb:
please ignore


ahmetb:
All pass. 🆗 That said we're not splitting the output right, now so it's hard to tell which manifest is being tested. It's ok for now since almost all PRs edit one manifest at a time.

PR #65, opened 2019-02-08
Add ctx and ns plugins

weibeld:

---

Two plugins for changing the current context and the namespace of the cluster of the current context, respectively, by fuzzy finding them with fzf ([here](https://github.com/weibeld/kubectl-ctx) and [here](https://github.com/weibeld/kubectl-ns)).

The plugin names are maybe not compatible with the naming guidelines. Do you have any suggestions what the plugins could be named? I chose them this way because it's short to type, and kubectl unfortunately doesn't yet auto-complete plugin names.

There's also  a `change-ns` plugin already in the index but it doesn't provide any means to quickly select the target namespace (e.g. by completion or fuzzy finding). So, could be good to have another plugin with a better UX at its place.

### DIFF
--- a/plugins/ctx.yaml
+++ b/plugins/ctx.yaml
@@ -1,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ctx
+spec:
+  version: v0.0.2
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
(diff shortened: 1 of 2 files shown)


ahmetb:
As the author of [`kubectx` and `kubens`](https://github.com/ahmetb/kubectx), I don't even add my own 3000-star project, because it would be a name grab. Sorry, but these names are too generic.

If we were to allow namegrabs, we'd err on the side of a well-established/popular projects –and in this case kubectx is both several years older, popular, and has a well-established user base.

PR #68, opened 2019-02-12
Add latest-release plugin

MQasimSarfraz:
I have a [simple plugin](https://github.com/MQasimSarfraz/kubectl-release-plugin) that lists the latest git releases for k8s projects. I have tested it locally and works well for me. It would be great if we can add it to `krew`

Also, do let me know if you think I need to rename anything here.

### DIFF
--- a/plugins/latest-release.yaml
+++ b/plugins/latest-release.yaml
@@ -1,0 +1,30 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: latest-release
+spec:
+  version: "v0.0.4"
+  shortDescription: List the latest release of kubernetes projects.
+  description: |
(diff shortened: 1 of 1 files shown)


ahmetb:
I think this tool is suitable as a generic tool that's not specific to kubernetes ecosystem. You can probably repurpose this tool to just search the repo with that name that has highest stars, and then look up its releases, like:

```
$ latest-release git
$ latest-release kubectx
$ latest-release fortio
$ latest-release ahmetb/kubectl-aliases
...
```

I think this is a super cool idea, but I doubt its usefulness as a kubectl plugin. I'd be happy to accept it anyway, but I suggest you re-evaluate if you want to limit this to be a kubectl plugin.  It would be way more useful as a generic to me personally.

MQasimSarfraz:
@ahmetb thanks for the feedback. I agree with the idea of making this tool generic. I will rework this soon but for now I will go ahead and close the PR! :+1:

PR #87, opened 2019-03-04
add init-tiller plugin

yogeek:
-----

### DIFF
--- a/plugins/init-tiller.yml
+++ b/plugins/init-tiller.yml
@@ -1,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: init-tiller
+spec:
+  version: "v1.0.0"       # optional, only for documentation purposes
+  platforms:
+    # url for downloading the package archive:
+  - uri: https://github.com/yogeek/kubectl-plugins/releases/download/1.0.0/kubectl-init-tiller-v1.0.0.tar.gz
+    sha256: "15e0b59882eab68265812b46c5d937a512130e37ca95fe84db02a15f74ba1423"
+    files:                     # copy the used files out of the archive
+    - from: "/init_tiller/*.sh" # path to the files extracted from archive
+      to: "."               # '.' refers to the root of plugin install directory
+    bin: "init_tiller.sh"  # path to the plugin executable after copying files above
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  shortDescription: Deploy and configure Tiller with RBAC in a given namespace.
+  # (optional) use caveats field to show post-installation recommendations
+  caveats: |
+    This plugin needs the following programs:
+    * kubens
+    * helm
+  description: |
+    This plugin deploys and configures Tiller with RBAC in a given namespace.


ahmetb:
I think tiller is going away in new version of Helm. In this case, we probably should not add this plugin.

I recommend you consider proposing this as a plugin to helm (they have a plugin mechanism, too) or as a feature?

yogeek:
@ahmetb thank you for your quick reply. However, as stated in the [Helm Roadmap](https://github.com/helm/helm/wiki/Roadmap#3x-roadmap), the version 3 _may be added in the indefinite future_. In the meantime, numerous teams will still be using V2 version and I think this plugin may be useful.
Maybe we can add it and it will be removed when Helm V3 makes it obsolete ?

ahmetb:
> Maybe we can add it and it will be removed when Helm V3 makes it obsolete?

I believe right now removing a plugin from the index might be breaking `krew upgrade` and `krew delete` commands. We haven't tested such a scenario thoroughly.

yogeek:
@ahmetb it would be a shame to refuse plugins because of this abnormal behaviour. Is it possible to test this removal with a fake pugin and to validate this doubt of yours ? If you need help, we can do some tests...

ahmetb:
I agree with you. Remove seems to work when manifest is gone from the index. However `upgrade` fails miserably. It's on the roadmap to add a "build receipt" feature that copies the installed manifest.

<details>
<summary>Details</summary>
<pre>
I0306 11:13:29.015547   45312 scanner.go:74] Reading plugin "warp"
F0306 11:13:29.031695   45312 root.go:50] open /Users/ahmetb/.krew/index/plugins/warp.yaml: no such file or directory
failed to load the index file for plugin
github.com/GoogleContainerTools/krew/cmd/krew/cmd.glob..func4
	/go/src/github.com/GoogleContainerTools/krew/cmd/krew/cmd/upgrade.go:59
github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra.(*Command).execute
	/go/src/github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra/command.go:762
github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra.(*Command).ExecuteC
	/go/src/github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra/command.go:852
github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra.(*Command).Execute
	/go/src/github.com/GoogleContainerTools/krew/vendor/github.com/spf13/cobra/command.go:800
github.com/GoogleContainerTools/krew/cmd/krew/cmd.Execute
	/go/src/github.com/GoogleContainerTools/krew/cmd/krew/cmd/root.go:48
main.main
	/go/src/github.com/GoogleContainerTools/krew/cmd/krew/main.go:23
runtime.main
	/usr/local/go/src/runtime/proc.go:201
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1333
</pre>
</details>

----

**More importantly,** I have a concern that plugins can be one-off install/bootstrap scripts aren't suitable for common availability here. As opposed to a kubectl plugin that users would call frequently, they would call it once per cluster (assuming they aren't going through a lot of clusters).

Historically, I've _turned down_ certain specialized "installation" plugins. Here's an example that (not so surprisingly) installs Helm (in a way [very similar to yours](https://github.com/danisla/kubefunc/blob/84b3e685d887c1c64486be1eddbb64f27731b779/install-helm/install-helm.sh)). https://github.com/GoogleContainerTools/krew-index/pull/11#discussion_r207714318

image

yogeek:
> they would call it once per cluster 
In fact, I created this plugin because we want to install Tiller once per _namespace_, and this for each cluster we have. In this case, it quickly becomes a repetitive tasks... (either for our CI that has to initialize every namespace, or for developers that want to create an ephemeral test namespace to test their helm chart)
But, I get your point and the arguments advanced in the other PRs are defendable...

ahmetb:
Apologies for this situation. Right now me being the only maintainer of the index repo isn't helping but I have an outstanding proposal to transfer the project to Kubernetes SIG CLI (https://groups.google.com/forum/#!topic/kubernetes-sig-cli/9fWm_aZ-AC8).

I don't want to accept plugins that might be ineligible later if/when the community comes up with criteria or acceptance rubric that prohibits such bootstrap/installation plugins. And accepting this plugin would set a bad precedent.

Your plugin submission is still a great input to the criteria process the community would come up with later on. So thanks for that. I think you may still be able to distribute your plugin with krew's `--manifest`, `--archive` flags but you would not get any upgrades.

yogeek:
@ahmetb No worry, I understand ;) Let's put this in standby for now until the community decide the direction of the project. 
now, it is quite disturbing for a newcomer to both solutions coexisting : the `kubectl plugin` feature on on hand (documented [officially](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/)), and `krew` on the other hand, that it mush more easier to use but not offically supported. So let's hope krew will become soon a part of k8s native ecosystem !
Thank you for your help :-)

bacongobbler:
Even if Helm 3 is released today, there will still be a large number of users running Helm 2 in production, and Helm 2 would continue to be supported for a given amount of time until we decide to drop support. I don't see a reason why this shouldn't be available for those users still running Helm 2.

ahmetb:
@bacongobbler Reiterating the discussion above, the reason I'm objecting to this isn't because Tiller is going away vs not; it's because I don't think we should have one-off plugins.

Plugins are meant to encapsulate common day-to-day workflows, and installation/bootstrapping operations are more suitable for separate tasks.

PR #90, opened 2019-03-10
Add homepage to open-svc

superbrothers:
This PR adds the homepage field to open-svc plugin manifest.

If there is no problem, I will create PRs to add the homepage field to other plugin manifests.

-----


ahmetb:
Yeah adding to all would be cool. 

I think it’s ok as we disabled unknown field validation in v0.2 release so we can upgrade from earlier versions.

PR #113, opened 2019-03-13
Add homepage fields.

jordanwilson230:
Adds a homepage field to the exec-as and prompt plugin manifests.
The homepage README has been revised to provide clear instructions, specific for the Krew plugins.

-----


ahmetb:
Is there a reason why you’re pointing to a branch and not the master?

jordanwilson230:
@ahmetb The master branch has plugins that aren't in the krew index, and some are named differently.  I created the krew branch to address the naming constraints and to provide only a subset of plugins that I thought would be useful to bring to krew (i.e., there's no need to add the ```kubectl switch``` plugin in master because there are already packages in krew that do the same thing).

PR #73, opened 2019-02-19
Add config-merge plugin

kairen:
Adds the config-merge plugin seen [here](https://github.com/kairen/kubectl-config-merge#kubectl-config-merge). Its usage as below:

```
$ k config-merge -h
Merge two or more kubeconfig files

Usage:
  config-merge [kubeconfig] [flags]

Examples:

	# Merge your kubeconfig and save it as "merge-config" in the current directory
	kubectl config-merge kubeconfig1 kubeconfig2 ...

	# Merge your kubeconfig and save it as "config" in "test" directory
	kubectl config-merge kubeconfig1 kubeconfig2 ... --path test/config

	# Merge your kubeconfig with $HOME/.kube/config as "$HOME/.kube/config"
	kubectl config-merge kubeconfig1 ... --home

	# To view merged kubeconfig result
	kubectl config-merge kubeconfig1 kubeconfig2 ... --view


Flags:
      --backup          If true, backup $HOME/.kube/config file to $HOME/.kube/config.bk (default true)
  -h, --help            help for config-merge
      --home            If true, merge with $HOME/.kube/config to $HOME/.kube/config
  -o, --output string   Merged kubeconfig output type(json or yaml) (default "yaml")
      --overwrite       If true, force merge kubeconfig
      --path string     Merged kubeconfig output name and path (default "merge-config")
      --version         Show config-merge version
      --view            View merged kubeconfig result 
```

-----

**Checklist for plugin developers:**

- [x] Read the [Plugin Naming Guide](https://github.com/GoogleContainerTools/krew/tree/master/docs/NAMING_GUIDE.md) (for new plugins)
- [x] Verify the installation from URL or a local archive works (`kubectl krew install --manifest=[...] --archive=[...]`)

PR #122, opened 2019-03-31
Add new plugin "konfig"

corneliusweig:
This is a small wrapper script around the `kubectl config view` command. It is intended as a way spread knowledge about this command with sane defaults. Users who know their way around `kubectl config view` do not need this.

Home: https://github.com/corneliusweig/konfig-merge
Initial release: https://github.com/corneliusweig/konfig-merge/releases/tag/v0.1.0

-----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
     )
 
   # TODO(ahmetb) pin this to a version via a variable
-  - env GOBIN=$HOME/bin go get github.com/GoogleContainerTools/krew/cmd/validate-krew-manifest
+  - env GOBIN=$HOME/bin go get sigs.k8s.io/krew/cmd/validate-krew-manifest
 
 install: true # skip go get ./...
 
--- a/plugins/konfig.yaml
+++ b/plugins/konfig.yaml
@@ -1,0 +1,51 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: konfig
+spec:
+  version: "v0.2.0"
+  platforms:
+    - uri: https://github.com/corneliusweig/konfig/releases/download/v0.2.0/bundle.tar.gz
+      sha256: 3ea1be03bc204aa4b50ce8547d82a5fe36f94b529b06977780c8ccab9d411977
+      bin: konfig-krew
+      files:
+        - from: ./konfig-krew
+          to: "."
+      selector:
+        matchLabels:
+          os: linux
+    - uri: https://github.com/corneliusweig/konfig/releases/download/v0.2.0/bundle.tar.gz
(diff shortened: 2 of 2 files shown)


corneliusweig:
This is similar to #73, however this implementation relies on the standard `kubectl config view` command to do the merging.

ahmetb:
Thanks for your new plugin contribution.

This basically doesn't provide any new functionality, it just makes the functionality existing in `kubectl config` command a bit more discoverable.

As a matter of fact, I wrote a blog post about these kubectl features because they aren't very discoverable and/or not widely known: https://medium.com/@ahmetb/mastering-kubeconfig-4e447aa32c75

So far we've been cautious about accepting such plugins and been putting them on `hold` until we come up with an acceptance among as the community. Accepting a plugin then removing it later right now has consequences, so we've been a little more careful lately.

@juanvallejo what are your thoughts on such plugins that provide a light shim over kubectl that don’t add new functionality but make existing functionality more discoverable?

corneliusweig:
@ahmetb Yeah, that's absolutely fine. I kind of had the same stance against the config-merge in go #73. However, I realized that I had written such a kubectl wrapper for myself, because I rarely needed config-merge, but when I did, I had forgotten how to do it. Therefore I think it could be helpful for others too.
Anyway, I appreciate that you want to keep the list of plugins short :)

Btw, I took your nice article as reference ;)

kairen:
@corneliusweig I have closed my PR because of this is better than #73.

corneliusweig:
@kairen Well, it seems as if this doesn't go into the index anyway. But if you have any idea how to improve this tool, we can collaborate on this. :bowing_man:

ahmetb:
We might take this in if users are finding it useful.
BTW isn't this kind off odd: `kubectl config-merge --extract`, the name says "merge", but the flag says "extract", so technically plugin doesn't always do "merge"?

corneliusweig:
Well.. this _is_ a bit odd. Other names I can come up with:
- `config-tool`: sounds awfully generic though
- `config-mixer` / `config-blender`: I kind of like that :).

I also plan to extend the tool to extract a kubeconfig for several contexts from any number of input kubeconfig files. That means option 2 may be more suited after all. WDYT?

I'll be away for a few days, so don't expect updates soon.

ahmetb:
🤔 Sounds like a bit of swiss-army knife of kubeconfig utilities really. 

I wonder if this could be designed like

- x merge
- x split (or export)
- x import

Where x is an original name like kubectl configutil, kubectl konfig, kubectl cfg or something like that.

corneliusweig:
@ahmetb I finally found time to revise the tool. It's now called just "konfig" and has the subcommands as suggested. Do you think this is now appropriate for the index?

ahmetb:
Looks great, thanks.
/lgtm
/approve

corneliusweig:
@ahmetb apparently, somebody needs to remove the "hold" label if it should be merged.

PR #126, opened 2019-04-01
Unify license headers

ahmetb:
The new copyright header is "# Copyright © 2019 The Kubernetes Authors" for
plugin manifests, although I'm not sure if manifest files are treated as code
in this case.

At any rate, I'm still figuring out how to actually test for presence and
correctness of the license headers as part of CI.

/assign @nikhita

### DIFF
--- a/plugins/krew.yaml
+++ b/plugins/krew.yaml
@@ -1,17 +1,3 @@
-# Copyright © 2018 Google Inc.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


ahmetb:
> This isn't really needed, but we can merge if you want to :) Manifest files don't really need the license header.

I think I wanted to reach some consistency about having license headers vs not (krew.yaml and one more plugin had it). Based on your recommendation I will remove this.

This saves us from using a hack/ script, which I was hesitant to write one as I was hoping there's a prow presubmit we could use. :P

PR #127, opened 2019-04-02
add remote sh

gabeduke:
-----


ahmetb:
Thanks for your contribution. I gather this plugin is pretty much an interactive chooser for `kubectl exec`, right? Does it have any other features?

If that's the case, I will recommend you to
- use "exec" instead of "remote shell" wording
- use a very clear description field, like the one I just explained above.

gabeduke:
@ahmetb I'm not sure if you meant the tool name itself or just the help language in the krew package info. I've updated the krew package info for now.

gabeduke:
@ahmetb the help language is updated as requested: see [6630bd6](https://github.com/kubernetes-sigs/krew-index/pull/127/commits/6630bd6c1b422e011b2a315f77f3924f90340f38)

ahmetb:
Sorry for the delay here.
What do you think about calling this interactive-exec, exec-interactive or iexec or something creative?

Rsh is a unix tool but it’s not relevant to this plugin, which can be misleading.

gabeduke:
@ahmetb changed to `iexec` as suggested

plugins/iexec.yaml
```diff
@@ -0,0 +1,42 @@
...
+    uri: https://github.com/gabeduke/kubectl-iexec/releases/download/0.1.14/kubectl-iexec_0.1.14_Darwin_x86_64.tar.gz
+    sha256: "e1c5524384fa5db09a84341ebd52bcfaa9d132fa341889c60ccb067335781d77"
+    files:
+    - from: "/kubectl-iexec"
+      to: "."
+    bin: "./kubectl-iexec"
+  shortDescription: Interactive selection tool for `kubectl exec`
+  homepage: https://github.com/gabeduke/kubectl-iexec
+  description: |
+    _          _               _   _        _
```
ahmetb: you might want to show your ascii art in `caveats` field, along with some examples perhaps.

right now I'm not very positive about `description` field having monospace art, as we might render them on readme, or an html site perhaps. `caveats` field is shown in the terminal after installing the plugin, so it's more appropriate. see other plugins as a reference.

if I were you, I'd remove ascii art altogether for the sake of simplicity. it doesn't add much to the experience.

plugins/iexec.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: iexec
+spec:
+  version: "0.1.9"
```
ahmetb: this is not matching the url below.
can you also start this string with `v`?

plugins/iexec.yaml
gabeduke: Removed as suggested. Fixed in the upstream tool help menu as well.

plugins/iexec.yaml
gabeduke: Fixed & updated to the latest version

PR #129, opened 2019-04-05
Add kubectl-config-cleanup plugin manifest

dbkegley:
-----

### DIFF
--- a/plugins/config-cleanup.yaml
+++ b/plugins/config-cleanup.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: config-cleanup
 spec:
-  version: "v0.3.0"
+  version: "v0.4.0"
   shortDescription: Automatically clean up your kubeconfig
   description: |
     This plugin will attempt to connect to each apiserver specified by a context entry in your kubeconfig.
@@ -11,8 +11,8 @@ spec:
     Otherwise, the entries are removed.
   homepage: https://github.com/b23llc/kubectl-config-cleanup
   platforms:
-  - uri: https://github.com/b23llc/kubectl-config-cleanup/releases/download/v0.3.0/kubectl-config-cleanup_0.3.0_Linux_x86_64.tar.gz
-    sha256: 8f934ecf33caaafbdaf1e949dac387241d326e35eddc5bedecb47ecf9bfc2f58
+  - uri: https://github.com/b23llc/kubectl-config-cleanup/releases/download/v0.4.0/kubectl-config-cleanup_0.4.0_Linux_x86_64.tar.gz
+    sha256: b33e747be6b99560a9e3078ee31e563af3b8c6c8a2c2ac9523db8c2d6b1aeaef
     bin: kubectl-config_cleanup
     files:
     - from: "./kubectl-config_cleanup"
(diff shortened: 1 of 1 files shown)


plugins/config-cleanup.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: config-cleanup
+spec:
+  version: "v0.2.0"
+  shortDescription: Automatically clean up your kubeconfig
```
ahmetb: Can you add a long description field?

Why is it automatic? How does it clean up? What parts does it cleanup?

Ideally these should be answered without installing the plugin.

dbkegley: I've updated the manifest with a longDescription

plugins/config-cleanup.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 spec:
   version: "v0.2.0"
   shortDescription: Automatically clean up your kubeconfig
+  longDescription: Every cloud provider has their own utilities for adding kubernetes cluster credentials to your kubeconfig but most don't offer the ability to clean it up once the cluster is deleted. For those of us who launch and delete multiple clusters per day, it is useful to have an automated way to clean up old kubeconfig entries. This plugin will attempt to connect to each cluster defined in a context, if the connection succeeds then the user, cluster, and context entries are maintained in the result. Otherwise, the entries are removed.
```
ahmetb: This should be “description”

Also please do line wrapping.

dbkegley: Updated field name, fixed line wrapping, and shortened the description

plugins/config-cleanup.yaml
```diff
@@ -0,0 +1,42 @@
+---
```
ahmetb: no need for this

plugins/config-cleanup.yaml
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: config-cleanup
+spec:
+  version: "v0.2.0"
+  shortDescription: Automatically clean up your kubeconfig
+  description: This plugin will attempt to connect to each apiserver specified by a context entry in your kubeconfig.
```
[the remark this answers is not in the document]
dbkegley: Fixed multi-line str and line length. Anything else I should add/remove in the desc?

ahmetb:
I think you also need to sign the Linux Foundation CLA as the bot has instructed above. Let me know if you already did.

ahmetb:
I just ran this plugin and it cleared up all my contexts:

```
kubectl config-cleanup
apiVersion: v1
clusters: []
contexts: []
current-context: ""
kind: Config
preferences: {}
users: []
```

The whole cmd took 0.1 secs. I don't think it's working correctly. I have 4 GKE clusters in my context. All of them are actually working fine.

dbkegley:
> I think you also need to sign the Linux Foundation CLA as the bot has instructed above. Let me know if you already did.

Still waiting on approval for the CLA, I'll update once it's approved.

> I don't think it's working correctly. I have 4 GKE clusters in my context. All of them are actually working fine.

I just tested on GKE and am seeing the same behavior, we can close this PR for now. I'll have to make some additional changes before this plugin is ready for general use. Thank you for the initial review. I'll reopen once the CLA is approved and I've tested the plugin more thoroughly with each of the auth-provider plugins.

ahmetb:
It would be good to keep it open so that we don't lose context on the review history.

ahmetb:
/hold
until CLA is signed, and until it works with auth providers like GKE.

dbkegley:
The newest version of the plugin should work with all auth provider plugins, please let me know if you run into any issues

ahmetb:
@dbkegley just tried, it has removed unreachable context, but I still see a bunch of dangling (unused) `users` and `clusters` entries, is this expected/by design?

ahmetb:
My bad, just saw `--users` and `--clusters`. But when I used it, I got error:

```
$ kubectl config-cleanup --users --clusters > kubeconfig

E0411 16:31:00.594794   59322 round_trippers.go:174] CancelRequest not implemented by *gcp.conditionalTransport
```

but `kubeconfig` was written to the file successfully.

dbkegley:
I've seen that one before as well, I believe it's related to this issue https://github.com/kubernetes/kubernetes/issues/73791

I've seen it from both the exec plugin and the gcp auth plugins

ahmetb:
Got it. The printed error is not a blocker. Next, please look at signing the CLA. 
/hold cancel

dbkegley:
I signed it

PR #135, opened 2019-04-13
Add debug plugin

aylei:
Hi there, thanks for reviewing my PR! Please don't hesitate to point out areas where I need to change.

Plugin [Home Page](https://github.com/aylei/kubectl-debug)

Signed-off-by: Aylei <rayingecho@gmail.com>

-----

### DIFF
--- a/plugins/debug-pod.yaml
+++ b/plugins/debug-pod.yaml
@@ -1,7 +1,7 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
-  name: debug
+  name: debug-pod
 spec:
   version: "v0.1.0"
   homepage: https://github.com/aylei/kubectl-debug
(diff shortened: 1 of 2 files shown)


ahmetb:
This name is too generic. Please choose something more unique and descriptive. Also there’s been a proposal to create an official debug command, which might be still going on: https://github.com/kubernetes/kubernetes/issues/45922

ahmetb:
I agree `debug-container` or `debug-pod` is better (debug-pod would be my preference). Sadly first-class kubectl commands can claim vauge/broad terminology like "debug" but I think we should be more careful.

Does this cause a problem with your documentation etc. Are you planning to go make the name change there as well?

aylei:
> Does this cause a problem with your documentation etc.
Not quite, I will add documents to explain the `krew` part about the slightly different name and it's a not a burden for me.

>  Are you planning to go make the name change there as well?
I prefer not, changing name can be troublesome for me and existing adopters.

aylei:
I realize that I have to update some codes to support both `kubectl debug-pod` and `kubectl debug` command suggestions, so I will close this PR now and open another PR in the next release of my project.

ahmetb:
Thanks! Expecting a new PR soon, happy to discuss what would be a more appropriate name.

PR #140, opened 2019-04-20
Add kubectl-grep

guessi:
### Description:

Search Kubernetes resources with keyword, and namespace supported

### Basic Usage:

From ...

    $ kubectl get -n namespace [resourceType] | grep 'keyword'

To ...

    $ kubectl grep [resourceType] [keyword]

Supported Resource Types:
- daemonset
- deployment
- hpa
- node
- pod

### Examples:

List all pods in all namespaces

    $ kubectl grep pods --all-namespaces

Shows all pods in namespace "star-lab" which contain the keyword "flash"

    $ kubectl grep pods -n star-lab flash
---

- [X] Read the [Plugin Naming Guide](https://github.com/GoogleContainerTools/krew/tree/master/docs/NAMING_GUIDE.md) (for new plugins)
- [X] Verify the installation from URL or a local archive works (`kubectl krew install --manifest=[...] --archive=[...]`)

### DIFF
--- a/plugins/grep.yaml
+++ b/plugins/grep.yaml
@@ -36,15 +36,57 @@ spec:
         arch: amd64
   version: v1.1.0
   homepage: https://github.com/guessi/kubectl-grep
-  caveats: This plugin requires an existing KUBECONFIG file, with a `current-context` field set.
+  caveats: |
+    This plugin requires an existing KUBECONFIG file, with a `current-context` field set.
+
+    Usage:
+
+      $ kubectl grep # output help messages
+      $ kubectl grep pods keyword
+      $ kubectl grep deployments keyword
+      $ kubectl grep pods keyword -n namespace
+      $ kubectl grep pods keyword --all-namespaces
+
+    More Info:
+    - https://github.com/guessi/kubectl-grep
+    Bug report:
+    - https://github.com/guessi/kubectl-grep/issues
+    Changelog:
+    - https://github.com/guessi/kubectl-grep/blob/master/CHANGELOG.md
   shortDescription: Search Kubernetes resources with keyword
   description: |
-    Search Kubernetes resources with keyword
+    Search Kubernetes resources with keyword, and namespace supported
 
     From ...
-    * kubectl get -n namespace [deployments|pods|...] | grep "keyword" # no Header output
+    $ kubectl get -n namespace [resourceType] | grep 'keyword'
 
     To ...
-    * kubectl grep -n namespace [deployments|pods|...] keyword
+    $ kubectl grep [resourceType] [keyword]
+
+    Supported Resource Types:
+    - daemonset
+    - deployment
+    - hpa
+    - node
+    - pod
+
+    For example, to list all pods in all namespaces,
+
+    $ kubectl grep pods --all-namespaces
+
+    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
+    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
+    marvel        spider-man-564d97dc9c-65mvw    1/1     Running    0         7d22h
+    dc            super-man-bb58c6784-26bmr      1/1     Running    0         1d3h
+
+    Or list all pods with specific keyword, under specific namespace,
+
+    $ kubectl grep pods -n star-lab flash
+
+    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
+    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
+
+    To support more type of resources, it is always welcome to open a pull-request
+    - https://github.com/guessi/kubectl-grep/pulls
 
     No more pipe, built-in grep :-)
--- a/plugins/grep.yaml
+++ b/plugins/grep.yaml
@@ -69,11 +69,9 @@ spec:
     Examples:
 
     List all pods in all namespaces
-
     $ kubectl grep pods --all-namespaces
 
-    Shows all pods in namespace "star-lab" which contain the keyword "flash"
-
+    List all pods in namespace "star-lab" which contain the keyword "flash"
     $ kubectl grep pods -n star-lab flash
 
     No more pipe, built-in grep :-)
--- a/plugins/grep.yaml
+++ b/plugins/grep.yaml
@@ -49,13 +49,9 @@ spec:
 
     More Info:
     - https://github.com/guessi/kubectl-grep
-    Bug report:
-    - https://github.com/guessi/kubectl-grep/issues
-    Changelog:
-    - https://github.com/guessi/kubectl-grep/blob/master/CHANGELOG.md
-  shortDescription: Search Kubernetes resources with keyword
+  shortDescription: Filter Kubernetes resources by matching their names
   description: |
-    Search Kubernetes resources with keyword, and namespace supported
+    Filter Kubernetes resources by matching their names
 
     From ...
     $ kubectl get -n namespace [resourceType] | grep 'keyword'
@@ -70,23 +66,14 @@ spec:
     - node
     - pod
 
-    For example, to list all pods in all namespaces,
+    Examples:
+
+    List all pods in all namespaces
 
     $ kubectl grep pods --all-namespaces
 
-    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
-    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
-    marvel        spider-man-564d97dc9c-65mvw    1/1     Running    0         7d22h
-    dc            super-man-bb58c6784-26bmr      1/1     Running    0         1d3h
-
-    Or list all pods with specific keyword, under specific namespace,
+    Shows all pods in namespace "star-lab" which contain the keyword "flash"
 
     $ kubectl grep pods -n star-lab flash
-
-    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
-    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
-
-    To support more type of resources, it is always welcome to open a pull-request
-    - https://github.com/guessi/kubectl-grep/pulls
 
     No more pipe, built-in grep :-)
--- a/plugins/grep.yaml
+++ b/plugins/grep.yaml
@@ -1,12 +1,12 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
-  name: search
+  name: grep
 spec:
   platforms:
-  - uri: https://github.com/guessi/kubectl-search/releases/download/v1.0.5/kubectl-search-Darwin-x86_64.tar.gz
-    sha256: feeb95a51cb2696757977f14da24d3d8a3a28a60b15ae5e15b87bba31a2c0fa1
-    bin: kubectl-search
+  - uri: https://github.com/guessi/kubectl-search/releases/download/v1.1.0/kubectl-grep-Darwin-x86_64.tar.gz
+    sha256: 2348dd6484903bfc394cfd9714e5bb66aa101c81e1f48227d819f726f89fbf96
+    bin: kubectl-grep
     files:
     - from: "*"
       to: "."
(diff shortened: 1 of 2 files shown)


corneliusweig:
@guessi Your doc sounds as if `kubectl grep` would be a more appropriate name. Am I mistaken?

guessi:
@corneliusweig It basically do the same job that `grep` does.

but `kubectl get ... | grep keyword` will remove the header, unless we execute `kubectl get ... | egrep '(^NAME|keyword)'`. Whereas `kubectl-search` will keep the format as `kubectl` does. So I think it's not just grepping the keyword.

ahmetb:
####  Name grab issues

The `search` name is a bit namegrab-ish. What if we have other plugins with the same name in the future that are potentially better?

#### Doesn't sound like a plugin we should have

This functionality actually exists in kubectl:

    kubectl get pods --filter=metadata.name=keyword

https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

In this case, we shouldn't probably convert something that could be a **very small** bash function to a plugin. You can probably just rewrite this as (pseudocode):

```sh
foo() {
    kubectl get args[-2] --filter=args[-1] args[3:*]
}
```

Similarly I think it can be done by `kubectl get pods | grep keyword` which utilizes grep, and good from an Unix philosophy perspective.

---

I currently vote to put a /hold on this, until we have better decision-making around whether ultra-small utilities like this should get in.

guessi:
@ahmetb Well... it's true, it can easily replaced by a small script, or any other kind of shell aliases, I love writing shell script for my daily jobs, and I do know it could be integrated into existed plugins, such as @corneliusweig 's great work: [ketall](https://github.com/corneliusweig/ketall).

---

My initial though was to provide another easy way to do `search` via a simple binary execution setup, and execute simple `kubectl search ...` right after the binary download, especially for the newbies.

Finding resources inside Kubernetes without caring:
- Where should I put the script?
- Is it compatible with my operation system?
- Is it compatible with existed aliases?
- It's not work under my shell environment, but I don't know why?
- Oops, there's no sample for the PowerShell, it's useless for me

Furthermore, everyone got his own dotfiles, but not all user familiar with these scripting jobs.

Take my colleague as example, pure Ops guy, or pure Dev guys, who run application on Kubernetes, but they don't write shell scripts for work. They just wanna **search** something.

~Yeah... lazy developers, but it's true, lol~

As for the **name** of the tool, just wanna made it easy to remember for all users, as I mention before, not all user familiar running their code on Unix-like system, e.g, `FindStr` for the PowerShell. So, I still insist **search** is much more better than **grep** for the tool's name.

That's why I create this tool, and named it as `kubectl-search`.

---

It's open discussion, please feel free to leave your opinion :-)

_* By the way, if there's bug, please do let met know, I will try to fix it ASAP._

ahmetb:
OK I’ll re-evaluate this and actually use the tool hands-on. It's unlikely I'll get back to it before KubeCon EU ends, so this may delay to end of May.

ahmetb:
/cc @corneliusweig 
wdyt? I feel like the plugin has matured enough since, and seems to be offering a complete experience. (Just that the name is still a bit generic, and doesn't use `cli-runtime` for parsing flags etc, which isn't mandatory).

corneliusweig:
Well this plugin has quite a narrow scope, but so do other plugins in the index. So that's fine I would say.

For the name, I also think that `search` or `find` is simply too generic, even if it makes for nicely readable command lines. Other names I can come up with: `kubectl grep`, `kubectl filter`, `kubectl find-matching`. Or you could opt for an acronym based name like `kubectl qsearch` (quick search), or something similar.

guessi:
@ahmetb @corneliusweig
- I've try to integrate with `cli-runtime`, but command options are vary from different type of resources, so I decide not to use `cli-runtime`, e.g., not all resources providing `-o wide`. maybe I'm trying the wrong way?
- As for the name of the plugin, I'm trying to `keep it short and easy to remember` initially
- Is it possible to create command alias for the `krew` such as `kubectl grep ...` would be aliased to `kubectl search ...` (note, I'm talking about built-in krew command alias, not shell aliases)

corneliusweig:
So the plugin system in `kubectl` works by looking for executables with naming pattern `kubectl-plugin_name`, which would translate to a subcommand `kubectl plugin-name`. The proper way to add another alias would therefore be to add another link to the binary in `$KREW_ROOT/store/search/<shasum>/kubectl-search` or so. But keep in mind that the `krew` home will most likely change in upcoming versions.
Bottom line: there is no easy way to achieve what you are trying to do.

ahmetb:
Common words like `kubectl search` are pretty dangerous because if/when kubectl natively implements a subcommand with this name, your plugin will not longer work since the builtin command will take priority.

In my opinion, the following names are all ok: `grep`, `search-resource`, `grep-resource`, `fd` (obscure but ok).

guessi:
@ahmetb @corneliusweig
Okay... I will rename it later, if rename is necessary, I'd prefer `kubectl-grep`, is that okay for you?

as WSL become more and more popular, `grep` might be _just okay_ for Windows users...

ahmetb:
Sounds good.

guessi:
working on that

guessi:
@ahmetb @corneliusweig
Sorry for the long wait, busy day 🤕 

Now, it's renamed as "grep", please review the changes, thanks

plugins/grep.yaml
```diff
@@ -0,0 +1,50 @@
...
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: v1.1.0
+  homepage: https://github.com/guessi/kubectl-grep
+  caveats: This plugin requires an existing KUBECONFIG file, with a `current-context` field set.
```
corneliusweig: `caveats` is shown post-install. You can also add some information about basic usage and how to get help.

guessi:
Update 👍

plugins/grep.yaml
guessi: Thanks & Updated

plugins/grep.yaml
```diff
@@ -0,0 +1,50 @@
...
+  shortDescription: Search Kubernetes resources with keyword
+  description: |
+    Search Kubernetes resources with keyword
```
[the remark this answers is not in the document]
guessi: Thanks & Updated

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      $ kubectl grep # output help messages
+      $ kubectl grep pods keyword
+      $ kubectl grep deployments keyword
+      $ kubectl grep pods keyword -n namespace
+      $ kubectl grep pods keyword --all-namespaces
+
+    More Info:
+    - https://github.com/guessi/kubectl-grep
+    Bug report:
```
corneliusweig: I'd not include links to bug reports or changelog. If somebody is interested, she will find it from the github link above.

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - https://github.com/guessi/kubectl-grep/issues
+    Changelog:
+    - https://github.com/guessi/kubectl-grep/blob/master/CHANGELOG.md
+  shortDescription: Search Kubernetes resources with keyword
+  description: |
+    Search Kubernetes resources with keyword, and namespace supported
```
corneliusweig:
```suggestion
    Filter Kubernetes resources by matching their names
```
That would also say how the plugin actually works: it only prints those resources where the keyword matches in the resource name.

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
+    marvel        spider-man-564d97dc9c-65mvw    1/1     Running    0         7d22h
+    dc            super-man-bb58c6784-26bmr      1/1     Running    0         1d3h
+
+    Or list all pods with specific keyword, under specific namespace,
+
+    $ kubectl grep pods -n star-lab flash
+
+    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
+    star-lab      the-flash-1563321660-bm225     1/1     Running    0         2d13h
```
corneliusweig: Instead of showing the output and letting the user derive what has happened, you could say something like:

> $ kubectl grep pods -n star-lab flash
Shows all pods in namespace `star-lab` which contain the keyword "flash".

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+    To support more type of resources, it is always welcome to open a pull-request
```
corneliusweig: Again, this is unnecessary cruft IMHO. If somebody wants to contribute, she will find her way from the github page.

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+      $ kubectl grep pods keyword -n namespace
+      $ kubectl grep pods keyword --all-namespaces
+
+    More Info:
+    - https://github.com/guessi/kubectl-grep
+    Bug report:
+    - https://github.com/guessi/kubectl-grep/issues
+    Changelog:
+    - https://github.com/guessi/kubectl-grep/blob/master/CHANGELOG.md
+  shortDescription: Search Kubernetes resources with keyword
```
corneliusweig:
```suggestion
  shortDescription: Filter Kubernetes resources by matching their names
```

plugins/grep.yaml
guessi: removed

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+    Supported Resource Types:
+    - daemonset
+    - deployment
+    - hpa
+    - node
+    - pod
+
+    For example, to list all pods in all namespaces,
+
+    $ kubectl grep pods --all-namespaces
```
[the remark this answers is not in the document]
guessi: it required to change the code of `kubectl-grep`

plugins/grep.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+    NAMESPACE     NAME                           READY   STATUS     RESTART   AGE
```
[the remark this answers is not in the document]
guessi: removed

plugins/grep.yaml
guessi: much better, thanks

plugins/grep.yaml
guessi: removed

plugins/grep.yaml
guessi: adopted, thanks

plugins/grep.yaml
guessi: adopted, thanks

guessi:
@corneliusweig 
Thanks for your precious time for reviewing these changes,
it's always welcome to have a strict review before getting merged.

I've adopt the advises except for the change of `--all-namespaces` to `-A` ,
since I haven't integrated `cli-runtime`, so it is not yet support `-A`.

_btw, I will try to integrate while I'm free, just not yet implemented._

plugins/grep.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - deployment
+    - hpa
+    - node
+    - pod
+
+    Examples:
+
+    List all pods in all namespaces
+
+    $ kubectl grep pods --all-namespaces
```
corneliusweig: Oh I didn't mean the flag `--all-namespaces` vs `-A`. What I meant to say was that users would most likely not use a plugin to do a job which is perfectly solved by plain `kubectl`.

But I think it's fine now. Don't bother.

plugins/grep.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - daemonset
+    - deployment
+    - hpa
+    - node
+    - pod
+
+    Examples:
+
+    List all pods in all namespaces
+
```
corneliusweig: Here it would greatly improve readability, you the blank line between the explanation and command would be gone, i.e.

> List all pods in all namespaces
$ kubectl grep pods --all-namespaces

vs.

> List all pods in all namespaces
>
> $ kubectl grep pods --all-namespaces

plugins/grep.yaml
```diff
@@ -0,0 +1,79 @@
...
+    $ kubectl grep pods --all-namespaces
+
+    Shows all pods in namespace "star-lab" which contain the keyword "flash"
+
```
corneliusweig: Same goes for this blank line.

corneliusweig:
@ahmetb do you want to have a final look before merging?

plugins/grep.yaml
guessi: BTW, I've added support for shorthand `-A` yesterday.

plugins/grep.yaml
guessi: Blank lines removed

plugins/grep.yaml
guessi: Blank lines removed

plugins/grep.yaml
```diff
@@ -0,0 +1,77 @@
...
+      $ kubectl grep pods keyword
+      $ kubectl grep deployments keyword
+      $ kubectl grep pods keyword -n namespace
+      $ kubectl grep pods keyword --all-namespaces
+
+    More Info:
+    - https://github.com/guessi/kubectl-grep
+  shortDescription: Filter Kubernetes resources by matching their names
+  description: |
+    Filter Kubernetes resources by matching their names
```
ahmetb: @corneliusweig in the future we should not have plugin authors have such long descriptions/caveats listing detailed USAGE

Ideally `description` explains what a plugin does, maybe in max 2-3 sentences.
`caveats` details any post-processing steps (and optionally points to the command to get more help).

I'll accept this but we should ideally clean up those long descriptions

PR #144, opened 2019-05-01
Add sort-manifests plugin

superbrothers:
This PR adds a new plugin manifest "sort-by-kind".

When installing manifests, they should be sorted in a proper order by Kind. For example, Namespace object must be in the first place when installing them.

"sort-by-kind" plugin sorts manfest files in a proper order by Kind, which is implementd by using tiller.SortByKind() in Kubernetes Helm.

```
$ ls ./manifests
deployment.yaml  ingress.yaml  namespace.yaml  service.yaml
$ kubectl sort-by-kind ./manifests | kubectl apply -f -
```

-----

### DIFF
--- a/plugins/sort-manifests.yaml
+++ b/plugins/sort-manifests.yaml
@@ -1,0 +1,36 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: sort-manifests
+spec:
+  version: v0.1.0
+  shortDescription: Sort manfest files in a proper order by Kind
+  description: |
+    When installing manifests, they should be sorted in a proper order by Kind.
+    For example, Namespace object must be in the first place when installing
+    them.
+
+    ksort sorts manifest files in a proper order by Kind, which is implementd by
+    using tiller.SortByKind() in Kubernetes Helm.
+  homepage: https://github.com/superbrothers/ksort
+  platforms:
+  - uri: https://github.com/superbrothers/ksort/releases/download/v0.1.0/ksort-darwin-amd64.zip
+    sha256: 4d7ab60c6e2aadcdc968e7bf9d383015cd8ff6ff6eca727b2475f5b0b336628d
+    bin: ksort
+    files:
+    - from: ksort
+      to: .
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  - uri: https://github.com/superbrothers/ksort/releases/download/v0.1.0/ksort-linux-amd64.zip
+    sha256: 52413ad1f8e1a92871afe376c888b8aa3ec534651fc4d6c90efb3c2fbeec03a7
+    bin: ksort
+    files:
+    - from: ksort
+      to: .
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64


corneliusweig:
I find the name not very intuitive. Based on what it does, I find `sort-for-install` or something alike more descriptive.

corneliusweig:
Thinking further, this version only supports install order. Why not also support
- Uninstall order
- alphabetically by resource name
- alphabetically by resource kind
- and maybe more?
Then the command name could be even more expressive, like `sort-manifests` with subcommands `for-apply`, `for-delete` `by-name`, `by-kind`,...

ahmetb:
I don't use Helm, but it sounds to me like alphabetical order (based on filename) isn't important to people (as `kubectl apply` would already do it).

Similarly, uninstall order isn't as important, you can delete a namespace and then a subresource in it, and `kubectl delete` would already attempt to delete all, at the end it would succeed. But maybe I'm missing something.

Thanks for reviewing @corneliusweig! Let me know if you have any interest in being a co-maintainer of index repo.

superbrothers:
> I find sort-for-install or something alike more descriptive.

This plugin command also supports uninstall order:

```
$ kubectl sort-by-kind --delete manifests/ | kubectl delete -f -
```

ahmetb:
I think a cleaner design would be:

```
kubectl sort-manifests --by=[kind,name,...] [--for-delete or --uninstall]
```

`sort-by-kind` is too specific. For a "plugin name", `sort-manifests`, `manifest-sort`, or even `sort` is better.

superbrothers:
Isn't `sort` too generic name? At first I intended to be `sort`, but I thought it was too general and I chose `sort-by-kind`. If there are no problem, I would like to choice `sort`.

ahmetb:
Yes “sort” is a bit too generic, although I can’t think of anything else to be sorted with kubectl. I recomment “sort-manifests”

superbrothers:
Thank you for suggestion. I renamed this plugin to "sort-manifests" at f3a3d95.

corneliusweig:
> Let me know if you have any interest in being a co-maintainer of index repo.

@ahmetb I'd be happy to help.

I think the plugin name `sort-manifests` is good.

@superbrothers Kustomize implements a very similar sort condition in `sigs.k8s.io/kustomize/pkg/resmap/idslice.go`. Depending on this for sorting  instead of `tiller` could reduce the file size of your binary.
Btw, any particular reason why you didn't provide a windows build in your manifest?

superbrothers:
@corneliusweig

> Kustomize implements a very similar sort condition in sigs.k8s.io/kustomize/pkg/resmap/idslice.go.

This looks good. I'll try to use it in the future.

> Btw, any particular reason why you didn't provide a windows build in your manifest?

Because I don't have a Windows machine, so I can't support it.

plugins/sort-manifests.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: sort-manifests
+spec:
+  version: v0.1.0
+  shortDescription: Sort manfest files in a proper order by Kind
```
ahmetb: manfest -> manifest

ahmetb:
Looks good to me, one minor nit.
/lgtm

superbrothers:
I squashed commits.

ahmetb:
🤔 the manifest validator is failing for some reason.

superbrothers:
> The command "env GOBIN=$HOME/bin go get github.com/GoogleContainerTools/krew/cmd/validate-krew-manifest" failed and exited with 2 during .

@ahmetb Could you retry the CI build? :thinking:

ahmetb:
#145 should fix it.
/hold cancel

PR #150, opened 2019-05-16
plugins: add tail (github.com/boz/kail)

boz:
see boz/kail#34
-----

### DIFF
--- a/plugins/tail.yml
+++ b/plugins/tail.yml
@@ -1,0 +1,58 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: tail
+spec:
+  version: "v0.10.1"
+  platforms:
+  - uri: https://github.com/boz/kail/releases/download/v0.10.1/kail_0.10.1_darwin_amd64.tar.gz
+    sha256: c0539d100c0c72ac88340976df7c3dd96febfcced7cf6cc5272b61a5ea514506
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    bin: kail
+    files:
+    - from: "kail"
+      to: "."
+  - uri: https://github.com/boz/kail/releases/download/v0.10.1/kail_0.10.1_linux_amd64.tar.gz
+    sha256: 253ab06570bc3a7afc6ab163b0ac9808b1ce257d2cbcd87c09859511bcbfc138
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    bin: kail
+    files:
+    - from: "kail"
+      to: "."
+  homepage: https://github.com/boz/kail
+  shortDescription: Stream logs from multiple pods and containers using simple, dynamic source selection.
+  description: -|
+    Kail https://github.com/boz/kail - "Just show me the logs"
+
+    Stream logs from all matched containers of all matched pods.  Match pods by service,
+    replicaset, deployment, and others.  Adjusts to a changing cluster - pods are
+    added and removed from logging as they fall in or out of the selection.
+  caveats: |
+    Documentation:
+
+      See https://github.com/boz/kail or
+
+      $ kubectl tail --help
+
+    Usage:
+
+      # match all pods
+      $ kubectl tail
+
+      # match pods in the 'frontend' namespace
+      $ kubectl tail --ns staging
+
+      # match pods belonging to a replicaset named 'workers' in any namespace.
+      $ kubectl tail --rs workers
+
+      # match pods belonging to the replicaset named 'workers' only in the 'staging' namespace
+      $ kubectl tail --rs staging/workers
+
+      # match pods belonging to both the service "frontend" and the deployment "webapp"
+      $ kubectl tail --svc frontend --deploy webapp


ahmetb:
/assign

looks like `-h` is failing for me on macOS

```
kubectl tail -h
usage: kubectl-tail [<flags>] <command> [<args> ...]

Tail for kubernetes pods

Flags:
  -h, --help                  Show context-sensitive help (also try --help-long and --help-man).
      --ignore=SELECTOR ...   ignore selector
  -l, --label=SELECTOR ...    label
  -p, --pod=NAME ...          pod
  -n, --ns=NAME ...           namespace
      --ignore-ns=NAME ...    ignore namespace
      --svc=NAME ...          service
      --rc=NAME ...           replication controller
      --rs=NAME ...           replica set
      --ds=NAME ...           daemonset
  -d, --deploy=NAME ...       deployment
      --node=NAME ...         node
      --ing=NAME ...          ingress
      --context=CONTEXT-NAME  kubernetes context
      --current-ns            use namespace from current context
  -c, --containers=NAME ...   containers
      --dry-run               print matching pods and exit
      --log-file=/dev/stderr  log file output
      --log-level=error       log level
      --since=DURATION        Display logs generated since given duration, like 5s, 2m, 1.5h or 2h45m. Defaults to 1s.panic: template: usage:30:53: executing "usage" at <
>: write /dev/stderr: resource temporarily unavailable

goroutine 1 [running]:
github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin%2ev2.(*Application).writeUsage(0xc420276000, 0xc42010b680, 0x0, 0x0)
	/home/travis/gopath/src/github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin.v2/app.go:235 +0x11b
github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin%2ev2.(*Application).maybeHelp(0xc420276000, 0xc42010b680)
	/home/travis/gopath/src/github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin.v2/app.go:249 +0xd0
github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin%2ev2.(*Application).Parse(0xc420276000, 0xc4200be010, 0x1, 0x1, 0xc420276000, 0x1d2e41b, 0x7, 0x1d3a53b)
	/home/travis/gopath/src/github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin.v2/app.go:213 +0x18a
github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin%2ev2.Parse(0x1d2e41b, 0x7)
	/home/travis/gopath/src/github.com/boz/kail/vendor/gopkg.in/alecthomas/kingpin.v2/global.go:37 +0x77
main.main()
	/home/travis/gopath/src/github.com/boz/kail/cmd/kail/main.go:97 +0x121
```

boz:
Thanks @ahmetb.  Fixed in v0.9.1.

/assign @ahmetb

plugins/tail.yml
```diff
@@ -0,0 +1,28 @@
...
+  version: "v0.9.1"
+  platforms:
+  - uri: https://github.com/boz/kail/releases/download/v0.9.1/kail_0.9.1_darwin_amd64.tar.gz
+    sha256: 3824de171a73c773649401f2bdccbd138f0022333c69e39e4f194e8ebd08290b
+    bin: kail
+    files:
+    - from: "kail"
+      to: "."
+    selector:
+      matchExpressions:
```
corneliusweig: I think you should also specify the supported architecture of the binary. This is better done with the `matchLabels` shorthand:
```yaml
selector:
  matchLabels:
    os: darwin
    arch: amd64
```

plugins/tail.yml
```diff
@@ -0,0 +1,28 @@
...
+      - {key: os, operator: In, values: [darwin]}
+  - uri: https://github.com/boz/kail/releases/download/v0.9.1/kail_0.9.1_linux_amd64.tar.gz
+    sha256: 6932b86f59c4b2cc73eb7aad889d3479566dfb4034d89a2d5f426722cdc5b780
+    bin: kail
+    files:
+    - from: "kail"
+      to: "."
+    selector:
+      matchExpressions:
```
corneliusweig: please also specify the architecture.

plugins/tail.yml
```diff
@@ -0,0 +1,28 @@
...
+      - {key: os, operator: In, values: [linux]}
+  homepage: https://github.com/boz/kail
+  shortDescription: Kubernetes log viewer
```
corneliusweig: To me this sounds a bit too generic. I can show logs with `kubectl logs`, so how is this better?

Looking at your homepage, it seems that this is the unique selling point:
> Match pods by service, replicaset, deployment, and others. Adjusts to a changing cluster ...

Can you condense that further, so that this fits in one line?

plugins/tail.yml
```diff
@@ -0,0 +1,28 @@
...
+  description: Show logs from matched pods.  See homepage or `-h` for usage.
```
corneliusweig: The description will be shown when people want to know more before installing, e.g. `kubectl krew info tail` (btw, the homepage is a new addition and is not yet shown with the latest krew release).

You do yourself a favor, if you explain a bit more what are the typical use-cases for your tool and how to invoke it.

boz:
Thanks for the feedback @corneliusweig.  I'll address these things when I have a chance.

> `E0517 11:42:07.325821   20111 streamwatcher.go:109] Unable to decode an event from the watch stream: context canceled
`
> Is that expected behavior?

It's from `glog` somewhere inside `k8s.io/{client-go,apimachinery,etc...}`.  I [tried to suppress](https://github.com/boz/kail/blob/master/cmd/kail/main.go#L167-L173) these but apparently they're still showing up.

I don't see them when I run `kail` by itself.  Is `kubectl` adding `glog` configuration?  Does your environment have `glog` settings?

Anyways... it's somewhat expected because a context is indeed cancelled to shut things down, but of course ideally the message wouldn't be shown :man_shrugging:

corneliusweig:
@boz I haven't checked where exactly the error is logged. But if you have it under control, do you already do a check like
```
if err ! = nil && ctx.Err() == context.Cancelled { return nil } 
```
(beware, typed on phone)

ahmetb:
If you’re seeing some weird behavior when process ran as a plugin vs standalone, this is totally worth filing a bug to kubectl.
Plugins are executed with execve(2) syscall, which would replace the kubectl process with your executable, so it shouldn't behave differently than running your executable standalone.

plugins/tail.yml
ahmetb: yeah if this block is just for darwin, `matchLabels:` is a better syntax.

plugins/tail.yml
ahmetb: If `-h` seems to be documenting the plugin well, we probably don’t need to explain usage string here.

boz:
> @boz I haven't checked where exactly the error is logged. But if you have it under control, do you already do a check like
> [snip]

I don't use `glog`.  I'm... not a fan ;)

> [snip]
> Plugins are executed with execve(2) syscall, which would replace the kubectl process with your executable, so it shouldn't behave differently than running your executable standalone.

I don't know of any other way of executing new programs.  It could still be that `kubectl` is adding environment variables that govern `glog` behavior.

It could be a lot of things, tho - including recent changes to `kail`.  I'll look into it when I have a chance.  I'm also [in the middle of bumping the k8s lib dependencies](https://github.com/boz/kcache/commit/a9ac92a30d8c1849909ae6a3042a92c9f2613e9a), which might resolve the issue.

@corneliusweig, @ahmetb: thanks for the help!

ahmetb:
OK no rush. Let’s hold until the issue is fixed.
/hold

boz:
Looks like `glog` has been removed from k8s.io libs :confetti_ball: :confetti_ball: :confetti_ball: 

I've updated the various manifest fields and bumped to a version of `kail` that uses up-to-date k8s.io libs.   Let me know what you guys think.

corneliusweig:
Looks like some examples in the `caveats` section are duplicated. I think three are fine anyway.

boz:
bump @ahmetb

ahmetb:
Sorry just got back from kubecon!

The only problem I'm seeing so far is a discrepancy with the potential style guide we'll be adding (#152): kail uses `--ns` flag whereas the recommended Kubernetes guideline is to use -n/--namespace (via git.k8s.io/kubernetes/cli-runtime).

But I think we can make an exception since the tool was existing before and has a broad user base that is not easy to break. @corneliusweig any input?

boz:
Right on.

FWIW `-n NAMESPACE` works and `--namespace NAMESPACE` can be added.  Note that the default is *all* namespaces sans `kube-system`; this can be overridden with `--current-ns` or `KAIL_CURRENT_NS=true`.

If it's a huge problem, adding `cmd/kubectl-tail` next to `cmd/kail` that conforms to some standard set of flags/behaviors is doable but something I'd like to avoid if at all possible.

ahmetb:
SG. I guess you can add —namespace later. 
/lgtm
/approve

We don’t strictly enforce the flags etc for now. So I think we can have this.

corneliusweig:
Sorry, for being so slow to respond.

I think it would be a great idea to also add `--namespace`. Having `--ns` in addition won't hurt anyone, but conforming to the k8s standard `-n/--namespace` is in general a good idea. I opened https://github.com/boz/kail/issues/38 so it is tracked.

PR #152, opened 2019-05-19
Krew plugin criteria

corneliusweig:
Hi @ahmetb @juanvallejo,

as you all know, krew-index currently has no criteria to apply for accepting or rejecting plugins. This creates friction for both contributors and maintainers:
- plugin contributors, because they have no idea whether their idea might be accepted.
- krew-index maintainers, because they need to do subjective case-by-case decisions whether a plugin actually "fits".

So I put up this draft for plugin criteria to get the discussion going. It is an opinionated crude first draft and probably misses a lot of things. But maybe it helps nevertheless. Let me know what you think.

cc @garethr because you wanted to tune in

### DIFF
--- a/doc/plugin-criteria.md
+++ b/doc/plugin-criteria.md
@@ -125,9 +125,10 @@ In order of decreasing usage frequency:
 
 Usability
 ---
-Further required acceptance criteria:
+Plugins should be as user-friendly as possible.
+Please follow the following guidelines:
 
-1. Standard look & feel: Adheres to standard kubectl best practices and in particular uses consistent (abbreviated) flag names. For example: bad `--ns`, good `--namespace` + `-n`.
-2. A plugin needs to be documented in such a way, that a proficient `kubectl` user immediately understands its purpose and how to use it.
+1. Standard look & feel: The plugin should adhere to standard kubectl best practices and have consistent (abbreviated) flag names. For example: bad `--ns`, good `--namespace` + `-n`.
+2. A plugin should be documented in such a way, that a proficient `kubectl` user immediately understands its purpose and how to use it.
 3. If the plugin has dependencies, try to limit those to fundamental UNIX tools (like bash, sed, grep, awk) or popular tools (like jq, yq) as this makes it easier to get started.
 4. Make sure the plugin fails with a user-friendly error when a required dependency is missing.
--- a/doc/plugin-criteria.md
+++ b/doc/plugin-criteria.md
@@ -129,6 +129,7 @@ Plugins should be as user-friendly as possible.
 Please follow the following guidelines:
 
 1. Standard look & feel: The plugin should adhere to standard kubectl best practices and have consistent (abbreviated) flag names. For example: bad `--ns`, good `--namespace` + `-n`.
+    - when writing a plugin in go, consider using the official `k8s.io/cli-runtime/pkg/genericclioptions`
 2. A plugin should be documented in such a way, that a proficient `kubectl` user immediately understands its purpose and how to use it.
 3. If the plugin has dependencies, try to limit those to fundamental UNIX tools (like bash, sed, grep, awk) or popular tools (like jq, yq) as this makes it easier to get started.
 4. Make sure the plugin fails with a user-friendly error when a required dependency is missing.
--- a/doc/plugin-criteria.md
+++ b/doc/plugin-criteria.md
@@ -37,9 +37,9 @@ This task should be classified in one of the categories:
           Mere command wrappers are not acceptable, as those are better coveredby shell aliases.
           However, convenience and productivity improvements are welcome. Examples:
            <ul>
-             <li>`konfig` is a convencience script to aid with merging of kubeconfig files, similar to `kubectl config view`.</li>
-             <li>`change-ns` is a utility easily switch between namespaces, similar to `kubectl config set-context --curent -n`</li>
-             <li>`view-secret` retrieves and decodes secrets, similar to `kubectl get secret -o jsonpath=... | base64 -d`</li>
+             <li><code>konfig</code> is a convencience script to aid with merging of kubeconfig files, similar to <code>kubectl config view</code>.</li>
+             <li><code>change-ns</code> is a utility easily switch between namespaces, similar to <code>kubectl config set-context --curent -n</code></li>
+             <li><code>view-secret</code> retrieves and decodes secrets, similar to <code>kubectl get secret -o jsonpath=... | base64 -d</code></li>
            </ul>
         </td>
     </tr>
@@ -53,17 +53,17 @@ This task should be classified in one of the categories:
           In general welcome unless it overlaps significantly with other plugins or has a bad usability.
           Examples:
           <ul>
-            <li>`get-all` plugin is a replacement for `kubectl get all` command which actually fetches only a subset of API resource types.</li>
-            <li>`access-matrix` gives an overview of authorities, whereas `kubectl auth can-i` only applies to single resource/verb pairs</li>
-            <li>`tail` plugin allows tailing logs from pods from a label selector, a namespace or all namespaces at once, whereas `kubectl logs` command has more limited functionality.</li>
-            <li>`mtail` plugin allows tailing logs from pods from a label selector, similar to `kubectl logs -l` but labels log lines.</li>
+            <li><code>get-all</code> plugin is a replacement for <code>kubectl get all</code> command which actually fetches only a subset of API resource types.</li>
+            <li><code>access-matrix</code> gives an overview of authorities, whereas <code>kubectl auth can-i</code> only applies to single resource/verb pairs</li>
+            <li><code>tail</code> plugin allows tailing logs from pods from a label selector, a namespace or all namespaces at once, whereas <code>kubectl logs</code> command has more limited functionality.</li>
+            <li><code>mtail</code> plugin allows tailing logs from pods from a label selector, similar to <code>kubectl logs -l</code> but labels log lines.</li>
           </ul>
         </td>
     </tr>
     <tr>
         <td>3.</td>
         <td>The task addresses a completely different problem than any of the kubectl subcommands, but is clearly focused on k8s.</td>
-        <td>Gray area. If the task fits well into `kubectl`, it may be accepted (for example `kubectl debug-pod`).</td>
+        <td>Gray area. If the task fits well into <code>kubectl</code>, it may be accepted (for example <code>kubectl debug-pod</code>).</td>
     </tr>
     <tr>
         <td>4.</td>
@@ -100,14 +100,14 @@ In order of decreasing usage frequency:
     <tbody>
     <tr>
         <td>1.</td>
-        <td>Recurring task on the same cluster. For example: `mtail`</td>
+        <td>Recurring task on the same cluster. For example: <code>mtail</code></td>
         <td>Favors acceptance as plugin.</td>
     </tr>
     <tr>
         <td>2.</td>
         <td>
           Recurring task on different cluster instances, but one-time task on the same cluster instance.
-          For example: `init-tiller`
+          For example: <code>init-tiller</code>
         </td>
         <td>Gray area. Usually this is not acceptable.</td>
     </tr>
--- a/doc/plugin-criteria.md
+++ b/doc/plugin-criteria.md
@@ -21,25 +21,90 @@ Purpose
 
 This task should be classified in one of the categories:
 
-| #  | Classification                                                                                                                                                                                                                 | Recommendation                                                                                                                                                                                            |
-|----|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| 1. |  The task is already solved by kubectl.                                                                                                                                                                                        |  Mere command wrappers are not acceptable, as those are better covered by shell aliases.  However, convenience and productivity improvements are welcome (examples: `konfig`, `change-ns`, `view-secret`) |
-| 2. | The task provides a different solution to a problem which is also solved by kubectl. However, the task would be impossible or very hard to reproduce with kubectl alone. Examples: `access-matrix`, `get-all`, `mtail`, `tail` |  In general welcome unless it overlaps significantly with other plugins or has a bad usability.                                                                                                           |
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for taking charge here. We definitely need this document, so this is gonna be a long PR.

Styling-wise: I feel like if you can put the Guidances right next to the situations (maybe markdown table, maybe just a sub-bullet point) it would be easier to read and leave review-comments.

corneliusweig:
Yeah, that looks so much nicer. Have you discussed about the plugin criteria during KubeCon?

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
+Krew plugin criteria
+===
+
+Rationale
+---
+
+Krew is a `kubectl` plugin manager.
+
+A plugin provides a technical solution to at least one _task_ concerning a problem around kubernetes.
```
garethr: The first paragraph (above) is talking about Krew. This paragraph jumps to talking about "A plugin", by which I take it to mean "A kubectl plugin". The description is then very opinionated.

I think it's fine to have strong opinions about what goes in the Krew index. I don't think it's OK to try and generalise that to the open plugin system that kubectl has. Anyone can create any plugin for kubectl that they want, simply by giving it a specific name and putting it on the path. Because of that the Krew index will always be a subset of available Kubectl plugins, which I think is fine.

I'm not sure if you want to define a "Krew plugin", or whether this should read something like "A plugin +in the Krew index+" or similar.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+| 2. | The task provides a different solution to a problem which is also solved by kubectl. However, the task would be impossible or very hard to reproduce with kubectl alone. Examples: `access-matrix`, `get-all`, `mtail`, `tail` |  In general welcome unless it overlaps significantly with other plugins or has a bad usability.                                                                                                           |
+| 3. | The task addresses a completely different problem than any of the kubectl subcommands, but is clearly focused on k8s.                                                                                                          |  Gray area. If the task fits well into `kubectl`, it may be accepted (for example `kubectl debug`).                                                                                                       |
```
garethr: The example of `kubectl debug` would appear to be against the [naming guidelines](https://github.com/kubernetes-sigs/krew/blob/master/docs/NAMING_GUIDE.md), in particular "Be Specific". It feels like the Login example provided there.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+| 4. | The task is very dissimilar to common kubectl tasks and is not primarily focused on k8s. However, it can be useful in the k8s ecosystem.                                                                                       | Not acceptable.                                                                                                                                                                                           |
+| 5. | The task is very dissimilar to common kubectl tasks and is not focused on k8s. Its general usefulness in the k8s ecosystem is questionable.                                                                                    | Not acceptable.                                                                                                                                                                                           |
```
garethr: Being arbiters if utility is bound to cause some friction. Usefulness is in the eye of the beholder. Homebrew hacks around this by saying packages have to be maintained by users, not by the original author. The gist being if a user is happy to maintain a package them by definition it's useful. Not saying Krew should follow the same pattern, but having guidelines which say "we decide based on what we think" might cause friction later.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+
+| #  | Classification                                                                                                            | Recommendation                             |
+|----|---------------------------------------------------------------------------------------------------------------------------|--------------------------------------------|
+| 1. | Recurring task on the same cluster. For example: `mtail`                                                                  | Favors acceptance as plugin.               |
+| 2. | Recurring task on different cluster instances, but one-time task on the same cluster instance. For example: `init-tiller` | Gray area. Usually this is not acceptable. |
+| 3. | One-time task per machine.                                                                                                | In general not acceptable as plugin.       |
+
+Usability
+---
+Furter required acceptance criteria:
```
garethr: Typo, Furter rather than Further

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Standard look & feel: Adheres to standard kubectl best practices and in particular uses consistent (abbreviated) flag names. For example: bad `--ns`, good `--namespace` + `-n`.
```
garethr: Would be good to link to a style guide here to avoid friction and nitpicking.

doc/plugin-criteria.md
corneliusweig: Yes, I fully agree. We should make clear that this is about krew plugins only. Can't do it right now, but you're welcome to suggest better wording.

doc/plugin-criteria.md
corneliusweig: Good point. @ahmetb do you know if such a document exists?

ahmetb:
Let's not merge all criteria into one document for now, it's harder to litigate all points in one PR.

doc/plugin-criteria.md
ahmetb: Yeah we should probably write a separate "Style Guide for Plugins" document. We could easily fill it up.

ahmetb:
I am thinking we can check this into krew main repo instead, since this repo ships to every krew client machine, and we already have a docs/ in the main repo.

doc/plugin-criteria.md
ahmetb: I recommend the following wording:

> The centralized Krew plugin index (`krew-index` repository) holds plugins that are applicable to a broad set of kubectl users.
>
> A plugin in Krew plugin index must add new functionality or enhance the existing functionality in kubectl, or capsulate a common operator/developer workflow for higher productivity.

We don't want to exclude people being creative and trying interesting things.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+- Purpose
+- Category
+- Usability
+
+Purpose
+---
+
+This task should be classified in one of the categories:
+
+| #  | Classification                                                                                                                                                                                                                 | Recommendation                                                                                                                                                                                            |
```
ahmetb: nit: (hope you won't hate me for saying this) I think if we actually do this part as an HTML table, we can also do 80-char word wrap and maintain this more easily:

```html
<table>
<thead>
<tr>
<td>column</td>
<td>column</td>
</tr>
</thead>
<tbody>
<tr>
<td>column</td>
<td>column</td>
</tr>
</tbody>
</table>
```

doc/plugin-criteria.md
```diff
@@ -0,0 +1,51 @@
...
+| 1. |  The task is already solved by kubectl.                                                                                                                                                                                        |  Mere command wrappers are not acceptable, as those are better covered by shell aliases.  However, convenience and productivity improvements are welcome (examples: `konfig`, `change-ns`, `view-secret`) |
+| 2. | The task provides a different solution to a problem which is also solved by kubectl. However, the task would be impossible or very hard to reproduce with kubectl alone. Examples: `access-matrix`, `get-all`, `mtail`, `tail` |  In general welcome unless it overlaps significantly with other plugins or has a bad usability.                                                                                                           |
```
ahmetb: I think we can give detailed examples. Take this sentence:

> For example:
> - `tail` plugin allows tailing logs from pods from a label selector, a namespace or all namespaces at once, whereas `kubectl logs` command has more limited functionality.
> - `get-all` plugin is a replacement for `kubectl get all` command which actually fetches only a subset of API resource types.

doc/plugin-criteria.md
ahmetb: +1 we rejected several plugins named `debug`.

A common task they do is: `kubectl cp` a busybox binary, `kubectl exec` to install busybox symlinks, then `kubectl exec -- sh` to drop user to a shell inside the container. Although I think we accepted one named `pod-debug`.

corneliusweig: Oops. I had https://github.com/kubernetes-sigs/krew-index/pull/135 in mind, but forgot that the agreement was to change the name.
Changed to `kubectl debug-pod`.

doc/plugin-criteria.md
corneliusweig: I agree that "usefulness" is rather subjective. But I think the krew index does need some form of filtering. Otherwise, where should the line be drawn? If there is no line (like for brew), krew would be come a mere package manager and not a kubectl package manager.
Maybe you have better ideas of what could be reasonable criteria?

corneliusweig:
> Let's not merge all criteria into one document for now, it's harder to litigate all points in one PR.

Are you referring to the idea to merge this with the naming guide?

doc/plugin-criteria.md
corneliusweig: No, I think this is absolutely necessary. Will do.

doc/plugin-criteria.md
corneliusweig: Yeah, I like the approach to argue from the user-basis as opposed to `kubectl`-centric.

ahmetb:
Thanks for doing the html migration, looks like we need to use full html i.e. `<code>` once we go html tables.

I want to capture an additional criteria about plugins being more broadly usable to most Kubernetes users, but this is hard to measure and evaluate.

- For example, `oidc-login` plugin is a nice plugin because there are many clusters using OIDC for authentication. <sup>[citation needed]</sup> So it's nice to have.
- On the other hand, if we had a plugin like `rancher-dashboard` that launches Rancher’s UI (and submitted by Rancher), then we're looking at a plugin applicable to a small % of the users <sup>[citation needed]</sup> –––and we can't measure whether people using OIDC for authentication VS clusters managed by Rancher are more.

One of them feels right, where the other one doesn’t, and this is hard to measure.

corneliusweig:
In the end, what you suggest always boils down to: how large is the userbase for setup XY. But even if we come up with an agreement what should be the acceptable number of users, how would we measure that in practice?

Hence, I'm more with @garethr here who brought up the brew philosophy: if somebody maintains it, it's fine. (Of course it still needs to match the other to-be-determined criteria).

What I _could_ imagine to work is a whitelist of well-known k8s projects for which plugins may be submitted. For example: knative, istio, linkerd, tekton, jenkins-x, rook.io, and so on. Of course such a list needs to be curated and be kept up to date. By default, all incubating CNCF projects should be on that list too.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,129 @@
...
+        <td>The task addresses a completely different problem than any of the kubectl subcommands, but is clearly focused on k8s.</td>
+        <td>Gray area. If the task fits well into <code>kubectl</code>, it may be accepted (for example <code>kubectl debug-pod</code>).</td>
+    </tr>
+    <tr>
+        <td>4.</td>
+        <td>
+          The task is very dissimilar to common kubectl tasks and is not primarily focused on k8s.
+          However, it can be useful in the k8s ecosystem.
+        </td>
+        <td>Not acceptable.</td>
```
ahmetb: +add:

> ...Consider distributing as a separate tool instead of making it a kubectl plugin.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,129 @@
...
+            <li><code>get-all</code> plugin is a replacement for <code>kubectl get all</code> command which actually fetches only a subset of API resource types.</li>
+            <li><code>access-matrix</code> gives an overview of authorities, whereas <code>kubectl auth can-i</code> only applies to single resource/verb pairs</li>
+            <li><code>tail</code> plugin allows tailing logs from pods from a label selector, a namespace or all namespaces at once, whereas <code>kubectl logs</code> command has more limited functionality.</li>
+            <li><code>mtail</code> plugin allows tailing logs from pods from a label selector, similar to <code>kubectl logs -l</code> but labels log lines.</li>
+          </ul>
+        </td>
+    </tr>
+    <tr>
+        <td>3.</td>
+        <td>The task addresses a completely different problem than any of the kubectl subcommands, but is clearly focused on k8s.</td>
```
ahmetb: I think an indicator of this is: 

> Most users will use this tool as a kubectl plugin.

If this is the case, then it's probably more suitable. e.g. in case of `helm` vs `kubectl helm`, there's no need to make it a plugin, because it doesn't add anything to kubectl. 

Worth adding.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,129 @@
...
+        <td>3.</td>
+        <td>One-time task per machine.</td>
+        <td>In general not acceptable as plugin.</td>
+    </tr>
+    </tbody>
+</table>
+
+Usability
+---
+Further required acceptance criteria:
```
ahmetb: I think we should relax this part a bit, and maybe make it optional to not to limit creativity too much.

doc/plugin-criteria.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Standard look & feel: Adheres to standard kubectl best practices and in particular uses consistent (abbreviated) flag names. For example: bad `--ns`, good `--namespace` + `-n`.
+2. A plugin needs to be documented in such a way, that a proficient `kubectl` user immediately understands its purpose and how to use it.
+3. Plugins may not depend on additional non-standard third-party tools which are not part of the plugin installation. Acceptable: jq, standard unix CLI tools (awk, sed, grep).
```
ahmetb: We should probably delete this item. It's hard to quantify while enforcing how many tools exactly is too many.

So maybe we can just say:

> - If your plugins are going to have dependencies, try to limit them to fundamental UNIX tools (like `bash`, `sed`, `grep`, `awk`) or popular tools (like `jq`, `yq`) as this makes it easier to get started.
> - Make sure your plugin fails with a user-friendly error when a required dependency is missing.

doc/plugin-criteria.md
corneliusweig: I've added that to the recommendation column. Was that what you intended?

doc/plugin-criteria.md
corneliusweig: Do you meant to not talk about "required" criteria?

ahmetb: I meant we should just remove the language that makes it "required". We can still list best practices, or develop a separate style guide.

corneliusweig: ok, makes sense.

corneliusweig:
@ahmetb I've added another recommendation to consider `genericclioptions` for plugins in go. Can you take another look?

corneliusweig:
@ahmetb @ferhatelmas WDYT, should we give this another try or should I close this?

ahmetb:
/remove-lifecycle stale
/lifecycle frozen

Let’s keep it open. Looks like we are gaining better confidence over time. We can take it in at some point.

PR #158, opened 2019-06-06
Add virt.yaml for v0.17.2

dhiller:
Create krew plugin for virtctl as part of the kubevirt project.
`virtctl` is the binary for executing advanced virtual machine
instance operations, see kubevirt.io for project details.

-----

- [x] Read the [Plugin Naming Guide](https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md) (for new plugins)

### DIFF
--- a/plugins/virt.yaml
+++ b/plugins/virt.yaml
@@ -13,8 +13,8 @@ spec:
       sha256: "bbf3b4f15b64ffe4c9196b53d045a92bc8113d24cfde471dded3c8c505aef663"
       files:
         - from: "/virtctl/virtctl-v0.17.2-linux-amd64"
-          to: "."
-      bin: "./virtctl-v0.17.2-linux-amd64"
+          to: "virtctl"
+      bin: "virtctl"
     - selector:
         matchLabels:
           os: windows
@@ -23,8 +23,8 @@ spec:
       sha256: "828876ae8abc50c7965f7b9eaa1ff3f6d3d9b0dad25c79e6a7d49dd6a1543087"
       files:
         - from: "/virtctl/virtctl-v0.17.2-windows-amd64.exe"
-          to: "."
-      bin: "./virtctl-v0.17.2-windows-amd64.exe"
+          to: "virtctl.exe"
+      bin: "virtctl.exe"
     - selector:
         matchLabels:
           os: darwin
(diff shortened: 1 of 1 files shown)


plugins/virt.yaml
```diff
@@ -0,0 +1,47 @@
...
+      sha256: "ffdc599d4581a98e230ffab30cde39a20db3a695d300d726f72bbdebd9a57361"
+      files:
+        - from: "/virtctl/virtctl-v0.17.2-darwin-amd64"
+          to: "."
+      bin: "./virtctl-v0.17.2-darwin-amd64"
+  shortDescription: controls virtual machine related operations.
+  homepage: https://kubevirt.io
+  caveats: |
+    Usage:
+      kubectl virt
```
ahmetb: This sounds redundant.

plugins/virt.yaml
```diff
@@ -0,0 +1,47 @@
...
+      bin: "./virtctl-v0.17.2-linux-amd64"
+    - selector:
+        matchLabels:
+          os: windows
+          arch: amd64
+      uri: "https://github.com/dhiller/kubectl-virt-plugin/releases/download/v0.17.2/virtctl-v0.17.2-windows-amd64.exe.tar.gz"
+      sha256: "828876ae8abc50c7965f7b9eaa1ff3f6d3d9b0dad25c79e6a7d49dd6a1543087"
+      files:
+        - from: "/virtctl/virtctl-v0.17.2-windows-amd64.exe"
+          to: "."
```
ahmetb: When you have version names in your binary, consider renaming here so you don’t update the bin field every time on a version change.

ahmetb:
I'm mostly cool with this.
/cc @corneliusweig 
wdyt?

Also a piece of feedback, when you run `kubectl virt` the help message shows `virtctl` instead of `kubectl virt` (you can look at os.args[0] to detect whether running as a plugin or not, the basename would be kubectl-virt[.exe]).

We also have this problem in a few other plugins, but if you can fix sometime later it would look even better.

plugins/virt.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - from: "/virtctl/virtctl-v0.17.2-darwin-amd64"
+          to: "."
+      bin: "./virtctl-v0.17.2-darwin-amd64"
+  shortDescription: controls virtual machine related operations.
+  homepage: https://kubevirt.io
+  caveats: |
+    Usage:
+      kubectl virt
+
+    Documentation:
```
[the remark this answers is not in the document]
corneliusweig: I didn't find examples there. However, this seems to be the best starting point for `virtctl`.
Nonetheless, you could recommend to run `kubectl virt help`.

plugins/virt.yaml
corneliusweig: You could mention here, that this plugin is a wrapper to make `virtctl` available as `kubectl virt`.

plugins/virt.yaml
```diff
@@ -0,0 +1,47 @@
...
+      https://kubevirt.io/user-guide/docs/latest/administration/intro.html#client-side-virtctl-deployment
+  description: |
+    virt plugin controls virtual machine related operations on your kubernetes cluster.
```
corneliusweig: Here, you should also mention that this plugin is a wrapper to make `virtctl` available as `kubectl virt`.

Besides, the description is what sells your plugin. Interested users will see the description when running `krew info virt`. Therefore you should provide all meaningful info about it, so that users understand why they would want it.

plugins/virt.yaml
corneliusweig: +1 for renaming. This should work:
```yaml
- from: "/virtctl/virtctl-v0.17.2-windows-amd64.exe"
  to: "virtctl.exe"
  bin: "virtctl.exe"
```

plugins/virt.yaml
```diff
@@ -0,0 +1,47 @@
...
+          os: darwin
+          arch: amd64
+      uri: "https://github.com/dhiller/kubectl-virt-plugin/releases/download/v0.17.2/virtctl-v0.17.2-darwin-amd64.tar.gz"
+      sha256: "ffdc599d4581a98e230ffab30cde39a20db3a695d300d726f72bbdebd9a57361"
+      files:
+        - from: "/virtctl/virtctl-v0.17.2-darwin-amd64"
+          to: "."
+      bin: "./virtctl-v0.17.2-darwin-amd64"
+  shortDescription: controls virtual machine related operations.
+  homepage: https://kubevirt.io
```
corneliusweig: I think it's a good idea to not link to your wrapper repo, but the official kubevirt docs instead :+1:

plugins/virt.yaml
dhiller: Good point. Will edit the caveat section to include advice on how to get started and that this is a wrapper for virtctl.

plugins/virt.yaml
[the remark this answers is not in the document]
dhiller: Yeah, actually that's all we have. We need to update the documentation there, I guess.

plugins/virt.yaml
dhiller: Interesting. I tried to check what that looks like by typing 

    $ kubectl krew install --manifest virt.yaml \
        --archive=virtctl-v0.17.2-linux-amd64.tar.gz
    $ kubectl krew info virt

And got this result: 

    F0607 09:48:58.882109   11512 root.go:52] plugin "virt" not found

Does that relate to the plugin being installed only locally and not yet present in the krew-index or is this a bug?

dhiller: > And got this result:
> 
> ```
> F0607 09:48:58.882109   11512 root.go:52] plugin "virt" not found
> ```
> 
> Does that relate to the plugin being installed only locally and not yet present in the krew-index or is this a bug?

It seems to be present though:

    $ kubectl krew list
    PLUGIN VERSION
    krew   dc2f2e1ec8a0acb6f3e23580d4a8b38c44823e948c40342e13ff6e8e12edb15a
    virt   bbf3b4f15b64ffe4c9196b53d045a92bc8113d24cfde471dded3c8c505aef663

ahmetb: Yes it's because it's not in the index. We have a PR that saves the installed plugin receipt, it would eventually solve this problem.

PR #160, opened 2019-06-11
virt: Adjust wording

fabiand:
This change is adjusting the wording in order to make the plugin discoverable
and alignt he language.

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>

### DIFF
--- a/plugins/virt.yaml
+++ b/plugins/virt.yaml
@@ -35,7 +35,7 @@ spec:
         - from: "/virtctl/virtctl-v0.17.2-darwin-amd64"
           to: "virtctl"
       bin: "virtctl"
-  shortDescription: Control KubeVirt Virtual Machines
+  shortDescription: Control KubeVirt virtual machines using virtctl
   homepage: https://kubevirt.io
   caveats: |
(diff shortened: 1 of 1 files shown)


plugins/virt.yaml
```diff
@@ -35,11 +35,11 @@ spec:
         - from: "/virtctl/virtctl-v0.17.2-darwin-amd64"
           to: "virtctl"
       bin: "virtctl"
-  shortDescription: controls virtual machine related operations.
+  shortDescription: Control KubeVirt Virtual Machines
```
corneliusweig: I understand the camel-case for "KubeVirt", but "virtual machines" is just plain English and can be lower-case, right? Also, I think the short description could hint that this is a wrapper around `virtctl`
```suggestion
  shortDescription: Control KubeVirt virtual machines using virtctl
```

fabiand: Both good points, I'll address them. Thanks.

fabiand: (Added a new commit as it suites my flow better)

fabiand:
Thanks!

PR #162, opened 2019-06-13
Add prune-unused plugin

etiennetremel:
Add kubectl-prune plugin, unfortunately I wasn't able to install it using,

I have been trying to install it manually but for some reason I'm getting the following `not executable` warning. When I look at the file locally installed it is correctly executable, might work on your machine... otherwise will fix... hint on how to fix it are welcome :)

```
$ wget https://github.com/thecloudnatives/krew-prune/archive/v0.1.0.zip
$ kubectl krew install --manifest=prune.yaml --archive=/tmp/v0.1.0.zip

$ kubectl plugin list
The following compatible plugins are available:

/Users/bob/.krew/bin/kubectl-krew
/Users/bob/.krew/bin/kubectl-prune
  - warning: /Users/bob/.krew/bin/kubectl-prune identified as a kubectl plugin, but it is not executable

error: one plugin warning was found
```
-----

- [*] Read the [Plugin Naming Guide](https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md) (for new plugins)
- [ ] Verify the installation from URL or a local archive works (`kubectl krew install --manifest=[...] --archive=[...]`)


plugins/prune.yaml
```diff
@@ -0,0 +1,27 @@
...
+  version: v0.1.0
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/v0.1.0.zip
+    sha256: 563d17508b37f56d668c2942922568cbaa014557651e69aedfcb8afda5a5f32c
+    files:
+    - from: kubectl-plugins-*/prune/kubectl-prune.sh
+      to: kubectl-prune
```
corneliusweig: Somehow, `krew` creates a folder here instead of renaming the file. @ahmetb is that a known issue / desired behavior?

You can work around as follows:
```yaml
files:
- from: kubectl-plugins-*/prune/kubectl-prune.sh
  to: "."
bin: ./kubectl-prune.sh
```

plugins/prune.yaml
```diff
@@ -0,0 +1,27 @@
...
+    bin: ./kubectl-prune
+  shortDescription: Prune secret/configmap resources from a given namespace
```
corneliusweig:
```suggestion
  shortDescription: Prune unused secret/configmap resources from a given namespace
```

ahmetb:
Thanks for your contribution.

Ideally for new plugins accepting namespace as an input, we want them to support `-n`/`--namespace` as non-positional arguments. Is it possible that you can implement this?

etiennetremel:
@ahmetb I added the namespace flag capabilities.
@corneliusweig I improved the script to take in consideration all the resources you mentioned. Also splitted the prune command into to: prune-configmaps and prune-secrets. Let me know if there is anything else

plugins/prune-configmaps.yaml
```diff
@@ -0,0 +1,31 @@
...
+  description: |
+    Delete configmaps that are not being used from a given namespace.
+    It checks mounted volumes, env and envFrom from all resources.
+  caveats: |
+    Usage:
+        kubectl prune-configmaps [options]
+
+    Options:
+        -n, --namespace='': If present, the namespace scope for this CLI request
+        -h, --help='': Deplay this help
```
corneliusweig:
```suggestion
        -h, --help='': Display this help
```

plugins/prune-secrets.yaml
```diff
@@ -0,0 +1,32 @@
...
+    Delete secrets that are not being used from a given namespace.
+    It checks mounted volumes, env, envFrom and imagePullSecrets from all
+    resources.
+  caveats: |
+    Usage:
+        kubectl prune-secrets [options]
+
+    Options:
+        -n, --namespace='': If present, the namespace scope for this CLI request
+        -h, --help='': Deplay this help
```
corneliusweig:
```suggestion
        -h, --help='': Display this help
```

etiennetremel:
I added the Daemonset resource but before moving forward, let's align on the naming convention. Thanks for catching that @corneliusweig .
I'm not a huge fan of using prune-app-config as naming convention for that plugin. The app-config part doesn't mean anything related to Kubernetes resource to me. At first I used prune which as you mentioned is too generic. What do you thing about `kubectl prune-resources secrets -n default` or `kubectl rm-unused secrets -n default`, other ideas?

corneliusweig:
Hey @etiennetremel, given that there is already a plugin `rm-standalone-pods`, I think the most in-line name would be `rm-unused`. Given that this name is also not too specific, I propose the following:
Keep the scripts for `secrets` and `configmaps` separate, but call them `kubectl-rm_unused-secrets` and `kubectl-rm_unused-configmaps`, respectively. The kubectl plugin mechanism will call those scripts when executing `kubectl rm-unused secrets`. That way, future plugins can use the same base-name for other plugins (e.g. `kubectl rm-unused volumes` or so).
@ahmetb WDYT?

ahmetb:
Sounds good. My main concern was that in the future kubectl may introduce a command called prune very easily and we don’t want to be in that situation.

ahmetb:
Circling back here: What do you think about this sort of naming:

```
kubectl prune-unused configmap [...]
kubectl prune-unused secret [...]
```

this lets you distribute it as single plugin and still have both functionality. Naming-wise, I think `prune-unused` is better than `rm-unused`.

etiennetremel:
Great, `prune-unused` it is. I will make the change in the coming days. Thanks.

etiennetremel:
@corneliusweig, as you described, I'm a little confuse on how to have sub command working in this case.

My plugin directory structure look like this:
```
prune-unused
├── kubectl-prune_unused-configmaps
└── kubectl-prune_unused-secrets
```

Then I would assume something like this would be the configuration to map each script:
```
    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/master.zip
    sha256: 4c84c36658ec414e99f4347a6c011dbe775c69d140768e04c3eceb2615481498
    files:
    - from: kubectl-plugins-*/prune-unused/*
      to: "."
    bin: ./kubectl-prune_unused-*
```

Is it possible to map sub command to a script as `kubectl prune-unused secrets` pointing to kubectl-prune_unused-secrets?
Or should I create 2 different plugin?

ahmetb:
We don’t support 2 plugin executables in one plugin at the moment. 
It came up before and we decided it’s not a common case. 

I recommend you write a small program in bash that accepts secret and configmap as a subcommand, then that program can call other executables you wrote in the same installation directory based on the subcommand. 

Or you can just merge two executables into one. (:

etiennetremel:
Thanks @ahmetb  for the clarification. I merged both scripts into a single one.

PR #173, opened 2019-07-09
plugins: add ksd.yaml (ashleyschuett/kubernetes-secret-decode)

bergerx:
This would address ashleyschuett/kubernetes-secret-decode#10

-----

- [ ] Read the [Plugin Naming Guide](https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md) (for new plugins)

I read it but I'm not the plugin owner, so i'm not sure about picking a name myself, i used "ksd" since the upstream plugin has been using either that or "kubernetes-secret-decode", but I don't think either is suitable. Maybe just `secret-decode`?
@ashleyschuett can you check the doc and comment on this one?

### DIFF
--- a/plugins/ksd.yaml
+++ b/plugins/ksd.yaml
@@ -1,0 +1,37 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ksd
+spec:
+  version: v3.0.0
+  platforms:
+  - sha256: 599e16041aa84f8a57812a860a6cc87d144eaa5126af953269e28577ed716e1b
+    uri: https://github.com/ashleyschuett/kubernetes-secret-decode/releases/download/v3.0.0/kubernetes-secret-decode_3.0.0_Linux_x86_64.tar.gz
+    bin: kubectl-ksd
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  - sha256: f0259dc015fae29294e8969e52ada3a44e71052e0584d379917c5b4f9434c2be
+    uri: https://github.com/ashleyschuett/kubernetes-secret-decode/releases/download/v3.0.0/kubernetes-secret-decode_3.0.0_Darwin_x86_64.tar.gz
+    bin: kubectl-ksd
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  homepage: https://github.com/ashleyschuett/kubernetes-secret-decode
+  shortDescription: Shows kubernetes secret with values base64 decoded
+  caveats: |
+    This plugin expects arguments similar to `get secret -o yaml`.
+    E.g. `kubectl ksd get secret <secretname> -o yaml`
+  description: |
+    Be able to easily see the values of a secret.
+    YAML and JSON are both supported and detection of the input type is performed automatically.
+    Usage: kubectl ksd get secret my-secret -o yaml
+


ashleyschuett:
Thank you! This has been on my todo list for a year now :/ `secret-decode` would be a fine name with me. I do think the actual binary will need to be updated depending on the name change, due to the way I was parsing commands.

bergerx:
@ahmetb @corneliusweig do you think the `secret-decode` is a suitable name for this plugin? Actually i briefly checked other plugin name `decode-secret` could actually be a better name.

Since the name change requires an update in the upstream repo lets agree on a name and we can follow up with an update on upstream and then here?

ahmetb:
I think both `secret-decode` and `decode-secret` are fine (though `{verb}-{noun}` usually makes more sense). We can respect the original name.

Also worth noting that you can have a different name in `krew` than the upstream repo (after all what determines a krew plugin’s name once is the `name:` field). We have several plugins whose names differ from their original repos. (Though, if the tool hardcodes its own name in USAGE messages etc, it's a problem).

corneliusweig:
I'm also fine with both variants, although `decode-secret` reads more naturally and is therefore better IMHO.

@ashleyschuett if you need to make adaptions upstream, then you might want to consider a simplification of the arguments. My understanding is that the original plugin is called as
```bash
kubectl ksd get secret my-secret
```
I think this could be made simpler like
```bash
kubectl decode-secret my-secret   # nb: no "get" or "secret"
```

---

As a side-note: There is already a plugin `view-secret` with a quite similar intention that decodes and shows one key. In addition, there is another plugin `edit-secret` in the pipeline. The `decode-secret` plugin works slightly different, but there is considerable overlap.

bergerx:
I personally used to use `view-secret` and even also `edit-secret` but after finding this one i just switched even before realising that i switched. It was due to the fact that i don't have to remember the usage. Its exactly same with getting a secret's yaml, if you want to see values decoded, just add the plugin name right after "kubectl" and voila. This is a personal preference and also the reason i created this pr, i personally prefer a `--decode` in kubectl and thia is closest.

ahmetb:
So have we decided to rename this as `decode-secret`?

corneliusweig:
> So have we decided to rename this as decode-secret?

Not quite, that is up to @bergerx :). However, my impression was that this was his favorite. The referenced PR is mainly to make the required command line for the plugin simpler.

ahmetb:
I feel like the author of the plugin probably should also have a say on how their plugin is also distributed. So I'm not sure if we rely on the preference of the _packager_ of the _author_ here. I’m OK in both cases, we should move this forward.

bergerx:
@corneliusweig is right, that is my personal preference (so i'd vote for number 3 below).

Seems like there are several opinions about this plugin's name and usage, I think its fair to go on with the upstream author's choice. So, @ashleyschuett, would you mind commenting on this one with your preference, and we can follow up with an upstream PR if needed and unblock this?

Based on the discussions above seems like these are the previously discussed options:
1 - `kubectl ksd get secret my-awesome-secret -o yaml` not really an option due to krew's [Plugin Naming Guide](https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md), this PR currently has this setup to fire up these discussions
2 - `kubectl secret-decode get secret my-awesome-secret -o yaml`, needs a PR on the upstream before addressing the change here
3 - `kubectl decode-secret get secret my-awesome-secret -o yaml` needs a PR on the upstream before addressing the change here
4 - `kubectl decode-secret my-awesome-secret -o yaml` as suggested in https://github.com/ashleyschuett/kubernetes-secret-decode/pull/15, uses `decode-secret` with explicitly dropping the `get secret` positional parameters

ahmetb:
ping @ashleyschuett , it would be great to have your opinion here.

corneliusweig:
@ashleyschuett We cannot go forward without you. Can you have a look here?

corneliusweig:
It's been over a month that @ashleyschuett last commented on this, but we need her to merge some required changes before `ksd` can go into the krew index. I think it's time to think about forking the project unless we hear from her. Would anybody be willing to maintain a fork?

cc @bergerx @ashleyschuett

corneliusweig:
/close
Due to inactivity. Feel free to re-open.

PR #177, opened 2019-07-15
Add exec-cronjob plugin

etiennetremel:
Add exec-cronjob plugin which extract the job spec of a cron job and instantly execute it.

```
Execute a CronJob, by extracting the Job spec and apply it in the given
namespace.

Usage:
    kubectl exec-cronjob <name> [options]

Options:
    -n, --namespace='': If present, the namespace scope for this CLI request
    --dry-run: If true, only print the object that would be sent, without sending it.
    -h, --help: Display this help
```

### DIFF
--- a/plugins/exec-cronjob.yaml
+++ b/plugins/exec-cronjob.yaml
@@ -1,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: exec-cronjob
+spec:
+  version: v0.4.0
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/v0.4.0.zip
+    sha256: 688bceed5fdd54414fc70080dacb776025c9ab31fb7bda3fca166bba05976a38
+    files:
+    - from: kubectl-plugins-*/exec-cronjob/*
+      to: "."
+    bin: kubectl-exec_cronjob.sh
+  shortDescription: Execute a cron job
+  description: |
+    Execute a CronJob, by extracting the Job spec and apply it in the given
+    namespace
+  caveats: |
+    Usage:
+        kubectl exec-cronjob <name> [options]
+
+    Options:
+        -n, --namespace='': If present, the namespace scope for this CLI request
+        --dry-run: If true, only print the object that would be sent, without sending it.
+        -h, --help: Display this help
+
+    Read the documentation at:
+      https://github.com/thecloudnatives/kubectl-plugins
+  homepage: https://github.com/thecloudnatives/kubectl-plugins
--- a/plugins/exec-cronjob.yaml
+++ b/plugins/exec-cronjob.yaml
@@ -3,21 +3,21 @@ kind: Plugin
 metadata:
   name: exec-cronjob
 spec:
-  version: v0.4.0
+  version: v0.4.1
   platforms:
   - selector:
       matchExpressions:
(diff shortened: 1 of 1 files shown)


plugins/exec-cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/v0.4.0.zip
+    sha256: 688bceed5fdd54414fc70080dacb776025c9ab31fb7bda3fca166bba05976a38
+    files:
+    - from: kubectl-plugins-*/exec-cronjob/*
+      to: "."
+    bin: kubectl-exec_cronjob.sh
+  shortDescription: Execute a cron job
```
corneliusweig:
```suggestion
  shortDescription: Run a cron job immediately
```

plugins/exec-cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+        kubectl exec-cronjob <name> [options]
+
+    Options:
+        -n, --namespace='': If present, the namespace scope for this CLI request
+        --dry-run: If true, only print the object that would be sent, without sending it.
+        -h, --help: Display this help
+
+    Read the documentation at:
+      https://github.com/thecloudnatives/kubectl-plugins
+  homepage: https://github.com/thecloudnatives/kubectl-plugins
```
corneliusweig: You are using one repo for multiple plugins, which is completely fine. But is there an anchor link to the docs for this particular plugin?

plugins/exec-cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/v0.4.0.zip
+    sha256: 688bceed5fdd54414fc70080dacb776025c9ab31fb7bda3fca166bba05976a38
+    files:
+    - from: kubectl-plugins-*/exec-cronjob/*
+      to: "."
+    bin: kubectl-exec_cronjob.sh
+  shortDescription: Execute a cron job
+  description: |
+    Execute a CronJob, by extracting the Job spec and apply it in the given
```
corneliusweig:
```suggestion
    Run a CronJob immediately, by extracting the Job spec and creating a Job instance thereof.
```

ahmetb:
Regarding naming, when I saw `exec-cronjob` I immediately thought this runs a scheduled cronjob "now", so I'm actually positive about the current name.

etiennetremel:
Made adjustments following your requests.

plugins/exec-cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/thecloudnatives/kubectl-plugins/archive/v0.4.1.zip
+    sha256: fc5410ebe09295a75ea9a67dd10ba83b3c6e5043ea9f0cf98897e362c665003a
+    files:
+    - from: kubectl-plugins-*/exec-cronjob/*
+      to: "."
+    bin: kubectl-exec_cronjob.sh
+  shortDescription: Run a cron job immediately
+  description: |
+    Run a cron job immediately by extracting the Job spec and creating a Job
```
ahmetb: I think this can be just `Run a CronJob immediately as Job` as most people would understand what it does.

Similarly, let's preserve Kubernetes API casing like CronJob.

corneliusweig:
I'm fine with `exec-cronjob`. Thanks @etiennetremel for bearing through this back-and-forth.
/lgtm

PR #182, opened 2019-07-17
meta: Make PR template invisible

ahmetb:
Better on the eye?
/assign @corneliusweig


corneliusweig:
Yes, good idea. That can be quite pesky over time!
/lgtm

PR #186, opened 2019-07-24
access-matrix: add RBAC keyword

ahmetb:
Would be good to mention RBAC in the shortDescription, as people scanning
the list cannot tell which plugins are rbac-related.

/assign @corneliusweig


corneliusweig:
Yeah, makes sense. Thanks for taking care of this :)
/lgtm
/approve

PR #189, opened 2019-07-31
Add kudo plugin

jbarrick-mesosphere:

### DIFF
--- a/plugins/kudo.yaml
+++ b/plugins/kudo.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: kudo
 spec:
-  version: "0.4.0"
+  version: "v0.4.0"
 
   shortDescription: Declaratively build, install, and run operators using KUDO.
   homepage: https://kudo.dev/
--- a/plugins/kudo.yaml
+++ b/plugins/kudo.yaml
@@ -8,9 +8,22 @@ spec:
   shortDescription: Declaratively build, install, and run operators using KUDO.
   homepage: https://kudo.dev/
   description: |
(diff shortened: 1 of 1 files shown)


plugins/kudo.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kudo
+spec:
+  version: "v0.4.0"
+
+  shortDescription: Declaratively build, install, and run operators using KUDO.
+  homepage: https://kudo.dev/
+  description: |
```
corneliusweig: Minor nit:
You are using quite long lines, which will lead to strange line-wrapping on smaller terminals. Can you re-format to a width of 80 characters?

ahmetb: We can consider making this a lint error if necessary since we have a validator tool.

jbarrick-mesosphere: Updated :+1:

jbarrick-mesosphere:
> If you have any advice you would want to show post-install, you can add a `caveats` field. For example, you could point to the documentation for writing kudo manifests and recommend to run `kubectl kudo -h` or so.

Added!

plugins/kudo.yaml
```diff
@@ -0,0 +1,90 @@
...
+  caveats: |
+    Example usage:
+      Install kafka:
+        kubectl kudo install kafka
+      List installed operator instances:
+        kubectl kudo get instances
+
+    See the documentation for more information: https://kudo.dev/docs/
+
+    Requires the KUDO controller to be installed:
```
ahmetb: you might wanna move requirements to the top.

jbarrick-mesosphere: Done :+1:

PR #190, opened 2019-08-01
Bump oidc-login to v1.14.1

int128:
This bumps oidc-login to [v1.14.1](https://github.com/int128/kubelogin/releases/tag/v1.14.1).

### DIFF
--- a/plugins/oidc-login.yaml
+++ b/plugins/oidc-login.yaml
@@ -10,11 +10,11 @@ spec:
     (Credential Plugin Mode)
       kubectl executes oidc-login before calling the Kubernetes APIs.
       oidc-login automatically opens the browser and you can log in to the provider.
-      After authentication, kubectl gets the token from oidc-login and you can access to the cluster.
+      After authentication, kubectl gets the token from oidc-login and you can access the cluster.
     (Standalone Mode)
       Run `kubectl oidc-login`.
       It automatically opens the browser and you can log in to the provider.
-      After authentication, it writes the token to the kubeconfig and you can access to the cluster.
+      After authentication, it writes the token to the kubeconfig and you can access the cluster.
 
   caveats: |
     You need to setup the OIDC provider, Kubernetes API server, role binding and kubeconfig.
--- a/plugins/oidc-login.yaml
+++ b/plugins/oidc-login.yaml
@@ -7,14 +7,18 @@ spec:
   shortDescription: Log in to the OpenID Connect provider
   description: |
     This is a kubectl plugin for Kubernetes OpenID Connect (OIDC) authentication.
-    (Credential Plugin Mode)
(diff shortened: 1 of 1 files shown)


plugins/oidc-login.yaml
```diff
@@ -3,30 +3,20 @@ kind: Plugin
 metadata:
   name: oidc-login
 spec:
-  shortDescription: Login for OpenID Connect authentication
+  homepage: https://github.com/int128/kubelogin
+  shortDescription: kubectl integration for OpenID Connect authentication
   description: |
-    This plugin gets a token from the OIDC provider and writes it to the kubeconfig.
```
corneliusweig: I found the previous description much better, I have to say. Is there a particular reason why you were unhappy with it?

int128: Since v1.14.0, kubelogin can be run as a client-go credential plugin or a standalone command. So the previous description is not accurate. I will revise the description.

int128:
Would you check the new description?

plugins/oidc-login.yaml
```diff
@@ -3,30 +3,27 @@ kind: Plugin
...
-    Just run:
-      % kubectl oidc-login
-
-    It opens the browser and you can log in to the provider.
-    After authentication, it gets an ID token and refresh token and writes them to the kubeconfig.
+    This is a kubectl plugin for Kubernetes OpenID Connect (OIDC) authentication.
+    (Credential Plugin Mode)
+      kubectl executes oidc-login before calling the Kubernetes APIs.
+      oidc-login automatically opens the browser and you can log in to the provider.
+      After authentication, kubectl gets the token from oidc-login and you can access to the cluster.
```
corneliusweig: Can you also link to the documentation with anchor tags, like https://github.com/int128/kubelogin#credential-plugin-mode ?

plugins/oidc-login.yaml
```diff
@@ -3,30 +3,27 @@ kind: Plugin
...
+    (Standalone Mode)
+      Run `kubectl oidc-login`.
+      It automatically opens the browser and you can log in to the provider.
+      After authentication, it writes the token to the kubeconfig and you can access to the cluster.
```
corneliusweig: Here, it's also worth linking to https://github.com/int128/kubelogin#standalone-mode

plugins/oidc-login.yaml
corneliusweig:
```suggestion
      After authentication, kubectl gets the token from oidc-login and you can access the cluster.
```

plugins/oidc-login.yaml
corneliusweig:
```suggestion
      After authentication, it writes the token to the kubeconfig and you can access the cluster.
```

int128:
Thank you for the review! Let me know if you have more suggestions.

PR #191, opened 2019-08-02
add plugin kubectl-neat

itaysk:
- [x] Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md


plugins/kubectl-neat.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: neat
+spec:
+  description: "If you try to `kubectl get` resources you have created, they will
```
corneliusweig: Is there a particular reason for not using multi-line yaml strings like
```yaml
description: |
  first line
  second line

  fourth line
```

itaysk:
Hi, Thanks for the feedback, you touched a good point, here's some additional context:
I'm aware of `get --export` and have been using it in the past. Since it was deprecated from core I thought to replace it with a plugin, and indeed when I started this project it was called `kubectl-export` and the intention was to replace `get --export` with an out of tree, community driven alternative. But the motivation for the project was refined while I was working on it (which caused me to decide to change the name). 
There are 2 problems to be solved by "cleaning" kubernetes resource definitions: 1) making them reusable by removing unique internal system information. 2) making them readable by removing stuff I'm not interested in.  
`get --export` aimed to solve the first problem (reuse) but it's unnecessary today because I can `kubectl get -oyaml | kubectl create -f -` and it will just ignore the unique internal information or anything that stands in the way to create the new resource. So if that problem went away, and export was deprecated, I decided to still pursue the other motivation of readability.  
readability requires more intrusive actions than just import/export and my plugin takes some non-universal assumptions in its heuristics compared to the "safe" and canonical `--export`. Also, readability was a side affect of export and I think it stands out from the name `export`.
This is why I deliberately decided to move away from 'export' semantics in order to not cause any confusion or false expectations with users. I chose neat because it kind of "cleans up" or "tidy up" complicated yamls for you, hence "neat" (and also "neat" is a.k.a cool :)).

I hope this answers your comment regarding the name of the plugin, as well as the question about the relation to import/export.

Regarding support for different resource kinds, it should be able to handle all resources, however not to the same degree. for resources it can predict defaults it will do more than for CRDs for example. But even for CRDs it can do some minimal cleanup in the metadata part so I'd generalize and say it handle all resource types.

plugins/kubectl-neat.yaml
itaysk: yes, the reason it that the plugin yaml file is auto generated from a json template. Since it's generated there's no way to preserve yaml features such as text blocks, but the result is functionally similar

corneliusweig: I see. In that case, that's fine. I don't want to break your pipeline :)
But could you make sure that the line length stays at around 80 characters? Long lines produce bad formatting on small terminals, so we try to avoid that.

corneliusweig:
@itaysk Btw, you need to rename the manifest to `neat.yaml`, not `kubectl-neat.yaml`.

plugins/kubectl-neat.yaml
itaysk: sure, I'll do that

itaysk:
1. supporting `-n/--namespace`: it's already supported
2. dependencies: when I started it actually worked like you suggest, I added a caveat about dependencies, and assumed the tools are in PATH. In a recent commit I bundled them into the package to ease onboarding. I'm was thinking about supporting both options, but I'm still not sure how to implement this. Is it ok if for now I leave it this way and add a caveat explaining this?
3. manifest name: sure, I'll rename to `neat.yaml`

plugins/kubectl-neat.yaml
```diff
@@ -0,0 +1,32 @@
...
+    files:
+    - from: /dist/linux/*
+      to: .
+    selector:
+      matchLabels:
+        arch: amd64
+        os: linux
+    sha256: 33f7afc2b5562c6cf73bb0ac026f0e60bf328eb965467458a10962dd0d025f1b
+    uri: https://github.com/itaysk/kubectl-neat/releases/download/v0.1.0/kubectl-neat_linux.tar.gz
+  shortDescription: Clean up Kuberntes yaml and json output to make it readable.
```
ahmetb: typo

ahmetb: it's also typoed multiple times on your repository page.

ahmetb:
Similarly your filename is wrong.

plugins/kubectl-neat.yaml
itaysk: fixed, thanks

plugins/kubectl-neat.yaml
itaysk: I managed to preserve text block style. also shortened to 80 chars per line

itaysk:
made the requested changes, let me know if this looks better now.

itaysk:
looks like tests are failing to install due to sha mismatch.. This is what I'm testing which works both on MacOS and Ubuntu, am I missing anything?

```
~/dev
❯ curl -OL https://raw.githubusercontent.com/kubernetes-sigs/krew-index/8f44352e12a501b39a2cdaf479a3af00aa394be1/plugins/neat.yaml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1482  100  1482    0     0   3392      0 --:--:-- --:--:-- --:--:--  3391

~/dev
❯ kubectl krew install --manifest ./neat.yaml
Installing plugin: neat
CAVEATS:
\
 |  This plugin depends on the following binaries which are bundled into the plugin -
 |  `jq`, `yq` and `sponge`, as well as the `kube-defaulter` included utility.
 |  glibc runtime is currently assumed for dynamic linking.
/
Installed plugin: neat

~/dev
❯ kubectl neat svc kubernetes
apiVersion: v1
kind: Service
metadata:
  name: kubernetes
  labels:
    component: apiserver
    provider: kubernetes
spec:
  clusterIP: 10.96.0.1
  ports:
  - name: https
    port: 443
    targetPort: 8443
```

ahmetb:
Restarted but I highly suspect you got it wrong. 

You can use KREW_OS and KREW_ARCH env vars to customize which os/arch spec is used while testing. And add -v=10 to see verbose logs.

itaysk:
looks like it worked

ahmetb:
I think bundling other binaries that you don't own is a bad idea. Especially when they can be installed with OS package managers.

That's way too many dependencies for a single plugin. We can still accept it, but I'm not sure if the users will like it. It would definitely be more idiomatic if it was written natively in a language like Go or Python that doesn't require shipping 4 extra binaries (jq, yq, sponge, kube-defaulter).

Also I just tried installing the current version on macOS, and despite the bundled binaries, it did not work:

```
$ kubectl neat namespace -o yaml
/Users/ahmetb/.krew/store/neat/f8e17b2dce4604b08056d3a4162376081b6d740ec4a2d7ffb0a3c2841e7b32b1/sponge: line 2: /usr/local/Cellar/moreutils/0.63/libexec/bin/sponge: No such file or directory
/Users/ahmetb/.krew/store/neat/f8e17b2dce4604b08056d3a4162376081b6d740ec4a2d7ffb0a3c2841e7b32b1/sponge: line 2: exec: /usr/local/Cellar/moreutils/0.63/libexec/bin/sponge: cannot execute: No such file or directory
```

itaysk:
@ahmetb yeah I get your point, I actually wrote it in the readme[1] but I guess I thought I'd push this v0.1 to the wild while I keep working on v1.0 rewrite:

> [1] Perhaps this seems like overuse of bash and other dependencies, and it might be true. I'll probably rewrite it completely in Go once I get the chance.

Anyway - I've now completely removed the dependency bundling (meaning the tools are expected in PATH) and bumped to v0.1.1. Feel free to reject this PR if you think it's wrong. Hopefully I'll submit a go version in a few weeks.

plugins/neat.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name: neat
+spec:
+  caveats: |
+    This plugin depends on the following binaries -
+    `jq`, `yq` and `sponge` (from `moreutils`), which should be installed by the user
+  description: |
+    If you try to `kubectl get` resources you have just created,
+    they be unreadably verbose. `kubectl-neat` cleans up that redundant information.
+    Can be used as a `kubectl get` wrapper, or by piping into it.
+    For example `kubectl neat pod mypod -o yaml` or
```
corneliusweig: It improves readability by putting examples in their own line, like

> Examples:
    $ kubectl neat pod mypod -o yaml
    $ kubectl get pod mypod -o yaml | kubectl neat

plugins/neat.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: neat
+spec:
+  caveats: |
+    This plugin depends on the following binaries -
+    `jq`, `yq` and `sponge` (from `moreutils`), which should be installed by the user
```
corneliusweig: Here, I'd also recommend to put the tools into one line each, in order to increase visibility. For example:

```yaml
caveats: |
  This plugin depends on the following tool which need to be installed independently:
    * jq
    * yq
    * sponge (from moreutils)
```

And you could mention where to get help, like `kubectl neat -h`?

plugins/neat.yaml
```diff
@@ -0,0 +1,38 @@
...
+    files:
+    - from: /dist/linux/*
+      to: .
+    selector:
+      matchLabels:
+        arch: amd64
+        os: linux
+    sha256: 18066653dc535646f940e8ed09f1d68eccdd63e6640eba5d8160544eec6a4f4f
+    uri: https://github.com/itaysk/kubectl-neat/releases/download/v0.1.1/kubectl-neat_linux.tar.gz
+  shortDescription: Clean up Kubernetes yaml and json output to make it readable.
```
corneliusweig: This is just a suggestion, maybe you like the wording:

> Remove clutter from manifests

or

> Remove clutter from manifests for readability

plugins/neat.yaml
itaysk: Thanks, updated

plugins/neat.yaml
itaysk: Thanks, updated

plugins/neat.yaml
itaysk: I've broken the lines as you suggested.
My plugin doesn't have a built in help option, because it's by default wrapping kubectl get (so -h will show the get help currently). In the rewrite I will refactor the get wrapping into a get subcommand (kubectl neat get) and will also add help.

itaysk:
Thanks for approving! I've accepted most of you suggestions
Yes, I'll reduce (eliminate) dependencies in the v1.0 release

ahmetb:
Seems like the fixes are in.
/lgtm
/approve

PR #193, opened 2019-08-05
New plugin: eksporter

Kyrremann:
This plugin helps you export resources, and trims away the fields not necessary for applying the resource back into the cluster.

It looks similar to the new plugin #191, but are focused on point 1 mentioned in the comments.

For me using `kubectl get -oyaml | kubectl create -f -` or similar technique aren't what I want.

---

- [x]  Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md
- [x]  Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...]

### DIFF
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -34,4 +34,4 @@ spec:
       kubectl get -oyaml | kubectl create -f -
 
     Run the tool with:
-      kubectl exporter <resource> <resource_name>
+      kubectl eksporter <resource> <resource_name>
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -7,7 +7,7 @@ spec:
   platforms:
   - selector:
       matchExpressions:
-      - {key: os, operator: In, values: [darwin, linux]}
+      - {key: os, operator: In, values: [darwin, linux, windows]}
     uri: https://github.com/Kyrremann/kubeflow-eksporter/releases/download/v1.0.0/eksporter.tar.gz
     sha256: "1c13fac59c1c306db8498f394ee4cf1ba98eaa44c6f404fc2fad505533f00835"
     files:
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -7,7 +7,12 @@ spec:
   platforms:
   - selector:
       matchExpressions:
-      - {key: os, operator: In, values: [darwin, linux, windows]}
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+        - windows
     uri: https://github.com/Kyrremann/kubeflow-eksporter/releases/download/v1.0.0/eksporter.tar.gz
     sha256: "1c13fac59c1c306db8498f394ee4cf1ba98eaa44c6f404fc2fad505533f00835"
     files:
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -14,11 +14,19 @@ spec:
     - from: "/eksporter.rb"
       to: "."
     bin: "./eksporter.rb"
-  shortDescription: Eksport resources without generated fields
+  shortDescription: Export resources and removes a pre-defined set of fields for later import
   homepage: https://github.com/Kyrremann/kubeflow-eksporter
   caveats: |
     This plugin needs the following programs:
     * ruby
   description: |
-    This plugin helps you export resources, and trims away the fields
-    not necessary for applying the resource back into the cluster.
+    This plugin helps you export a resource, and removes a pre-defined set of
+    fields. This is useful for when you have resource you want to migrate from
+    one namespace or cluster, to another one, but need to edit/add some fields
+    before applying.
+
+    If you just need to move one resource, it's probably easier to just:
+      kubectl get -oyaml | kubectl create -f -
+
+    Run the tool with:
+      kubectl exporter <resource> <resource_name>
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -1,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: eksporter
+spec:
+  version: "v1.0.0"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/Kyrremann/kubeflow-eksporter/releases/download/v1.0.0/eksporter.tar.gz
+    sha256: "1c13fac59c1c306db8498f394ee4cf1ba98eaa44c6f404fc2fad505533f00835"
+    files:
+    - from: "/eksporter.rb"
+      to: "."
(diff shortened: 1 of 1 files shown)


itaysk:
Hi! I'm the author of kubectl-neat (that other plugin you mentioned). I'm curios about the motivation for this plugin, can you please elaborate?
From skimming your code, looks like you are just removing some metadata, similar to what I do in [phase1](https://github.com/itaysk/kubectl-neat/blob/master/src/construct.resource) of my plugin. My understanding is that this is not required for functional export/import, meaning `kubectl get | kubectl create` just works, but you mentioned you're not interested in that. On the other hand you say your motivation is not readability either, so what use case is it?
I'm asking because I thought about allowing users to configure the plugin, for example to only perform this phase1 cleaning, and not proceed with the other phases. That would be similar result to this plugin but I didn't find an appealing use case for that.

Kyrremann:
@itaysk it is mostly for the `get|create` flow, but I also (usually) edit the file in some way. Or using a working resource as a template for something else I'm doing. So by making a simple export-function to strip out the generated fields shortens my work.

Of course, if your plugin can do only phase 1, we don't really need this plugin. I also see that you remove a lot more fields then what I would have removed, but that would of course always be a problem.

itaysk:
Thanks for clarifying

plugins/eksporter.yaml
```diff
@@ -0,0 +1,24 @@
...
+    sha256: "1c13fac59c1c306db8498f394ee4cf1ba98eaa44c6f404fc2fad505533f00835"
+    files:
+    - from: "/eksporter.rb"
+      to: "."
+    bin: "./eksporter.rb"
+  shortDescription: Eksport resources without generated fields
+  homepage: https://github.com/Kyrremann/kubeflow-eksporter
+  caveats: |
+    This plugin needs the following programs:
+    * ruby
```
corneliusweig: Is the `yaml` library built-in in ruby, or does the user have to install some gem for this?

plugins/eksporter.yaml
```diff
@@ -0,0 +1,24 @@
...
+  description: |
+    This plugin helps you export resources, and trims away the fields
+    not necessary for applying the resource back into the cluster.
```
corneliusweig: So from the looks of your code it seems you are trying to solve the problem, where a resource is exported from one cluster/namespace and re-applied in another namespace. Is that correct? If so, you should explicitly mention this use-case here and in the `shortDescription`

plugins/eksporter.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/Kyrremann/kubeflow-eksporter/releases/download/v1.0.0/eksporter.tar.gz
+    sha256: "1c13fac59c1c306db8498f394ee4cf1ba98eaa44c6f404fc2fad505533f00835"
+    files:
+    - from: "/eksporter.rb"
+      to: "."
+    bin: "./eksporter.rb"
+  shortDescription: Eksport resources without generated fields
```
corneliusweig: I think you are giving a false promise here. From what I can tell, your plugin removes a pre-defined set of fields from the manifest. For example, `metadata.labels` and `metadata.namespace`. Both of those could have been set explicitly by the user. There is no way for you to differentiate between automatically set defaults or user-specified values in those cases.

So I think you need to revise your description :)

plugins/eksporter.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: eksporter
+spec:
+  version: "v1.0.0"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
```
corneliusweig: The plugin is a ruby script. Shouldn't a windows environment with a ruby installation not also qualify as a target OS?

plugins/eksporter.yaml
Kyrremann: Yaml is built-in. At least from `Ruby-1.9.1`.

plugins/eksporter.yaml
Kyrremann: I agree, and I'll borrow some text from what you wrote:

> Export resources and strippes away a pre-defined set of fields

plugins/eksporter.yaml
Kyrremann: Fixd, and will be in the next release after I've updated the descriptions and help text.

ahmetb:
I also don't have problems with competing plugins doing the same thing.
However as you might imagine, it's better for the community to have a single solution that's easy to install (with few to no dependencies) and works for most use cases.

plugins/eksporter.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: eksporter
+spec:
+  version: "v1.0.0"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux, windows]}
```
ahmetb: This line is a bit long, can you unfold {} into separate lines:

```
- key: os
  operator: In
  values:
  - darwin
  - linux
  - windows
```

I think this is the first plugin we ever have that supports all 3 OSes.

ahmetb:
/lgtm
/approve
/hold
for a small nit

plugins/eksporter.yaml
Kyrremann: Do we even need a selector when I target all 3 OSes?

I'll update the code anyway, as it looks better on separated lines.

ahmetb: That’s a good point.
I think this used to work like:

   matchLabels: {}

But we recently introduced tighter validation code I believe. We might relax it in the future. Well, at least for now it’s a one-off operation.

ahmetb: Indeed we did prevent this: https://github.com/kubernetes-sigs/krew/blob/8d148b5be03eb1a6611378ddd78d412cd3a9b5da/pkg/index/validation/validate.go#L162

ahmetb: so I think matchLabels:{} might be worming today but it will stop working in the next versions of krew.

plugins/eksporter.yaml
```diff
@@ -0,0 +1,37 @@
...
+    This plugin helps you export a resource, and removes a pre-defined set of
+    fields. This is useful for when you have resource you want to migrate from
+    one namespace or cluster, to another one, but need to edit/add some fields
+    before applying.
+
+    If you just need to move one resource, it's probably easier to just:
+      kubectl get -oyaml | kubectl create -f -
+
+    Run the tool with:
+      kubectl exporter <resource> <resource_name>
```
ahmetb: eksporter?

Kyrremann: Fixd

PR #196, opened 2019-08-06
Add plugin: support-bundle

divolgin:
This is a new plugin that generates a support bundle for a cluster.

Example:

kubectl troubleshoot https://troubleshoot.replicated.com

The generated support bundle can be manually inspected or uploaded to https://vendor.replicated.com for automated analysis.

### DIFF
--- a/plugins/support-bundle.yaml
+++ b/plugins/support-bundle.yaml
@@ -1,0 +1,72 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: support-bundle
+spec:
+  version: "v0.9.3"
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/replicatedhq/troubleshoot/releases/download/v0.9.3/support-bundle_0.9.3_linux_amd64-0.9.3.tar.gz
+    sha256: "557d22dc64723623f61a94e1faa854dbaa3bf3ccecc6f02d3d219a6c8ba878fa"
+    files:
+    - from: "/support-bundle"
+      to: "."
+    bin: "support-bundle"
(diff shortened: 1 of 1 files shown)


ahmetb:
I think the name “troubleshoot” is far too generic per Krew’s naming guidelines. 
- It’s sort of a name grab (what happens when others want to develop troubleshooting plugins? what should they call them?)
- name doesn’t sufficiently describe what the plugin does (troubleshoot what? Nodes? Volumes? Load balancers? Perf issues?)

divolgin:
I signed it

divolgin:
@ahmetb I see how that name is generic.  The intention is to collect information that can be used for troubleshooting the entire cluster: nodes, volumes, load balancers, any applications that may be running on it, etc....  All this information is saved in a support bundle file.

I'll rename this to support-bundle.  That's does seem more specific about what this plugin does: generates a support bundle from a cluster.

ahmetb:
Thanks, I think a name like `support-bundle` is definitely more appropriate and specific to make it clear that this is not as general-purpose as the original name reads.

divolgin:
@ahmetb I've changed the name to `support-bundle`

ahmetb:
Thanks.  I will accept the plugin but you have 3 visible issues:

1. the help message doesn't understand the new plugin name

```
kubectl support-bundle
Error: requires at least 1 arg(s), only received 0
Usage:
  troubleshoot [url] [flags]
  troubleshoot [command]
```

2. please consider adding `-n` flag in addition to namespace.

3. please consider importing auth plugins like `gcp`, `azure`, `oidc` etc in your code, otherwise this will not work with managed cloud clusters (e.g. GKE, AKS).

```
kubectl support-bundle https://troubleshoot.replicated.com
panic: No Auth Provider found for name "gcp"
```

PR #199, opened 2019-08-14
Update virt (KubeVirt virtctl plugin package) to v0.20.1

dhiller:


corneliusweig:
/lgtm
/approve

Would it be possible to keep the order of platforms next time? Simplifies the review :smiley:

dhiller:
> Would it be possible to keep the order of platforms next time? Simplifies the review

Yeah that's a good idea, thanks!

PR #202, opened 2019-08-15
Add plugin list page

ahmetb:
Introduces a lightweight plugins.md that can be linked in onboarding docs
at the "krew" repo as well as this "krew-index" repo.

Supersedes https://github.com/kubernetes-sigs/krew/pull/303.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ This is the Plugin Index Repository for [krew kubectl plugin manager][krew].
 
 [krew]: https://github.com/kubernetes-sigs/krew
 
-You can find a list of kubectl plugins available on krew in the
-[`./plugins` directory](./plugins).
+You can find a list of kubectl plugins available on krew [here](./plugins.md)
+or in the [`./plugins` directory](./plugins).
 
 ## Submitting new plugins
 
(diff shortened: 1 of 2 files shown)


corneliusweig:
I was thinking about how to include this in krew-index but avoid distributing to our users. One idea would be to use `git hash-object -w` and then reference the object in an annotated tag, which is pushed to github. Then only users pulling all tags would get the object. Junio Hamano does a similar trick with his public gpg key for signing tags in git/git. I still have to explore though, if github can show the markdown file, though.

On second thought: maybe an easier approach is to have a dedicated branch where the plugins.md lives. I'm quite sure this does work :)

ahmetb:
The file is really small, like a plugin manifest. So I'm not too worried about it. It's like having a plugin that updates occassionally.

corneliusweig:
One final thought before merging: as this file is not automatically updated, I think we should include a disclaimer on top that this may be outdated -- and say what revision this file refers to. WDYT?

ahmetb:
Yeah that's possible, we should say run `kubectl krew search` for an up-to-date list.  But that requires change in the tool, so we can do later.

PR #203, opened 2019-08-15
New Eksporter release - v1.1.0

Kyrremann:
Someone asked for piping, so it's been added.
See https://github.com/Kyrremann/kubeflow-eksporter/issues/1.

Only thing I'm on the fence for, is if I should add a line with piping in the manifest-documentation?

### DIFF
--- a/plugins/eksporter.yaml
+++ b/plugins/eksporter.yaml
@@ -35,3 +35,5 @@ spec:
 
     Run the tool with:
       kubectl eksporter <resource> <resource_name>
+    Or with piping:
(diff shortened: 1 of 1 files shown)


corneliusweig:
I don't think it's necessary as long as the built-in help mentions this, which it should. But the description is short enough that it could afford the example. So your call...

Kyrremann:
I added it in :)

Kyrremann:
Going to add one more change, didn't see it before now, that the repo is named `kubeflow-ekpsorter`, it should be `kubectl`...guessing I've been working to much with kubeflow lately.

PR #204, opened 2019-08-16
Bump outdated to 0.3.0

marccampbell:
Bump outdated plugin to v0.3.0 with support for specifying an alternate kubecontext as a flag

### DIFF
--- a/plugins/outdated.yaml
+++ b/plugins/outdated.yaml
@@ -3,14 +3,14 @@ kind: Plugin
 metadata:
   name: outdated
 spec:
-  version: "v0.3.0"
+  version: "v0.3.1"
   platforms:
   - selector:
       matchLabels:
         os: linux
         arch: amd64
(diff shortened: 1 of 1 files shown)


corneliusweig:
I see that you chose `kubecontext` as the flag name. Would it be possible to use the standard flag name for that, which is just `context`? This will improve the usability, because users are already accustomed to that flag name.

Have you also considered using k8s.io/cli-runtime/pkg/genericclioptions for flags? This provides a lot of standard flags, but I'm not sure if this is viable in your case.

marccampbell:
Thanks @corneliusweig.  Makes sense to me, I'll update

corneliusweig:
Thanks, I'll put this on hold for now 
/hold

marccampbell:
@corneliusweig Thanks for the link and pointer to the `genericclioptions` package. This worked great, and we now have the full pflag support added to this plugin.

corneliusweig:
Looks great!
/lgtm
/approve

PR #208, opened 2019-08-18
Add gopass plugin to krew plugin index

martinhoefling:
PLUGIN DEVELOPERS: @martinhoefling, with some hints from @corneliusweig 

Exposes 
- inserting secrets to gopass from a template
- diffing and applying a single secret stored in gopass 
- recursive diffing and applying secrets

via gopass subcommand.

### DIFF
--- a/plugins/gopass.yaml
+++ b/plugins/gopass.yaml
@@ -1,0 +1,22 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: gopass
+spec:
+  version: "v0.0.2"
+  platforms:
+    - selector:
+        matchExpressions:
+          - {key: os, operator: In, values: [darwin, linux]}
+      uri: https://github.com/gopasspw/kubectl-gopass/archive/v0.0.2.zip
+      sha256: "794bcbf82428e321f3301c443437259cd79ee13d31dbff3dc4ad226421a2a7a7"
+      files:
+        - from: "/kubectl-gopass-*/kubectl-gopass"
+          to: "."
+      bin: "./kubectl-gopass"
+  shortDescription: Imports secrets from gopass
+  homepage: https://github.com/gopasspw/kubectl-gopass
+  caveats: |
+    This plugin requires bash and gopass to be set up correctly
+  description: |
+    This plugin allows applying and diffing secrets from gopass to k8s


plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
...
+      uri: https://github.com/gopasspw/kubectl-gopass/archive/v0.0.3.zip
+      sha256: "ae4d2aa8ec061aa6916b2120fa3caf48d1cee328bab7cd263a03f8b1aeec22f8"
+      files:
+        - from: "/kubectl-gopass-*/kubectl-gopass"
+          to: "."
+      bin: "./kubectl-gopass"
+  shortDescription: Imports secrets from gopass
+  homepage: https://github.com/gopasspw/kubectl-gopass
+  caveats: |
+    This plugin requires bash and gopass to be set up correctly
```
ahmetb: “Correctly” is redundant.

plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
...
+  description: |
+    This plugin allows applying and diffing secrets from gopass to k8s
```
ahmetb: Kubernetes is better.

plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: gopass
+spec:
+  version: "v0.0.3"
+  platforms:
+    - selector:
+        matchExpressions:
+          - {key: os, operator: In, values: [darwin, linux]}
```
corneliusweig: The script is in pure bash. I think it would also make sense to allow installation on Windows and mention bash as a pre-requisite (which you already do).

plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: gopass
+spec:
+  version: "v0.0.3"
```
corneliusweig: There is a version mismatch between this manifest and what is reported by `kubectl gopass version` (see https://github.com/gopasspw/kubectl-gopass/blob/59315e46389540624cca8c6ecb45694ed1dc3e87/kubectl-gopass#L140). Have you considered to generate deployment artifacts with a script?

plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - selector:
+        matchExpressions:
+          - {key: os, operator: In, values: [darwin, linux]}
+      uri: https://github.com/gopasspw/kubectl-gopass/archive/v0.0.3.zip
+      sha256: "ae4d2aa8ec061aa6916b2120fa3caf48d1cee328bab7cd263a03f8b1aeec22f8"
+      files:
+        - from: "/kubectl-gopass-*/kubectl-gopass"
+          to: "."
+      bin: "./kubectl-gopass"
+  shortDescription: Imports secrets from gopass
```
corneliusweig:
```suggestion
  shortDescription: Imports Secrets from gopass
```
Making it a capital `S` signals that this is about kubernetes Secret resources. Or maybe  even

> Imports Secret resources from gopass

plugins/gopass.yaml
```diff
@@ -0,0 +1,22 @@
...
+  homepage: https://github.com/gopasspw/kubectl-gopass
+  caveats: |
+    This plugin requires bash and gopass to be set up correctly
+  description: |
```
corneliusweig: Most plugins also give some basic usage instructions and how to get help.

ahmetb: I think it's ok if --help/-h works.

plugins/gopass.yaml
martinhoefling: Actually, windows has bash support in newest windows 10 build. I've enabled it. Not sure about the ending though. I will probably revisit later on as gopass support for windows itself is an ongoing topic.

plugins/gopass.yaml
martinhoefling: Fixed, I added a small helper to create the tags and extracted the version as variable in the script.

plugins/gopass.yaml
martinhoefling: I can copy / paste the usage output here, but not sure what you will gain. I think the usage is quite obvious for gopass users right?

plugins/gopass.yaml
corneliusweig: I'm a bit puzzled now. The manifest has a windows section which also make the CI test fail. Was that intentional?

ahmetb:
Current manifest fails to run on Windows:

> failed to install plugin "gopass": can't create symbolic link, source binary ("/tmp/krew-test665291632/store/gopass/8bae04664693153555af921fb19b896aaccf39f7dec200e430816ff3fa47e1eb/kubectl-gopass.exe") cannot be found in extracted archive: stat /tmp/krew-test665291632/store/gopass/8bae04664693153555af921fb19b896aaccf39f7dec200e430816ff3fa47e1eb/kubectl-gopass.exe: no such file or directory

plugins/gopass.yaml
martinhoefling: No it shouldn't fail. I made another attempt to fix the build.

corneliusweig:
/approve
@ahmetb Do you want to take another look, or should I merge?

PR #210, opened 2019-08-21
Add custom-cols plugin

fredleger:
- Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md : 🆗 
- Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...] : 🆗 

This adds custom-cols a simple wrapper and template repository that allow easy use of custom columns format output.

ex:

```bash
kubectl custom-cols -o qos-class pods -l app=nginx
```


fredleger:
CLA signed :-)

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,30 @@
...
+  homepage: https://github.com/webofmars/kubectl-custom-cols
+  caveats: |
+    This plugin needs the following programs:
+    - env(1)
+    - bash
+    - realpath
+  description: |
+    This plugin allow you to pre-define some custom columns file templates that
+    you can then re-use with ease.
+    It's time saving to retrieve some ofen neeeeded infos like
```
corneliusweig:
```suggestion
    It's time saving to retrieve some often needed infos like 
```

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  version: "v0.0.2"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/webofmars/kubectl-custom-cols/archive/v0.0.2.tar.gz
+    sha256: "f01d1aa63c320989071baa88d7b0a8fefa1147e0da8b9edf32874def21c84736"
+    files:
+    - from: "./kubectl-custom-cols-0.0.2/kubectl-custom-cols"
```
corneliusweig: It would keep the noise down for future manifest revisions, if you could change your archive to use a static path (without the version string), like so:
```suggestion
    - from: "./kubectl-custom-cols"
```

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,30 @@
...
+      to: "."
+    - from: "./kubectl-custom-cols-0.0.2/templates/*.tpl"
+      to: "templates"
+    bin: "./kubectl-custom-cols"
+  shortDescription: Easy use of custom columns output formating.
+  homepage: https://github.com/webofmars/kubectl-custom-cols
+  caveats: |
+    This plugin needs the following programs:
+    - env(1)
```
corneliusweig: I think you can safely assume that `env` and `realpath` are present, so you don't need to list those. (They are both part of the posix spec)

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - bash
+    - realpath
+  description: |
+    This plugin allow you to pre-define some custom columns file templates that
```
corneliusweig:
```suggestion
    This plugin allows you to pre-define some custom columns file templates that
```

plugins/custom-cols.yaml
fredleger: Humm. i added these from my tests using GH generated release and installing with krew (only overriding the manifest path). It seems that GH adds this folder when building a tar.gz release. Previous was what you suggested. 

Unless i missed something ? 🤔

plugins/custom-cols.yaml
fredleger: Looks fair enough !

fredleger: removed the requirements

fredleger:
hi @corneliusweig 

Thanks for taking the time to test and review it.

You are totally right about the fact you will need to list the templates. My idea was to add it in the next version and for now list it in the documentation.

Actually i just put 4 templates i add in a folder before writing the plugin (was the motivation for it though) as placeholders but i'am not very happy with them and i would like community to PR the main repo to add more useful ones :-)

Do you think it's acceptable like this and add the listing / adding of templates in another version ?

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,36 @@
...
+spec:
+  version: "v0.0.2"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/webofmars/kubectl-custom-cols/archive/v0.0.2.tar.gz
+    sha256: "f01d1aa63c320989071baa88d7b0a8fefa1147e0da8b9edf32874def21c84736"
+    files:
+    - from: "./kubectl-custom-cols-0.0.2/kubectl-custom-cols"
```
ahmetb: FYI instead of `./kubectl-custom-cols-0.0.2/kubectl-custom-cols` you can do `./kubectl-custom-cols-*/kubectl-custom-cols` as well.

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,36 @@
...
+      to: "."
+    - from: "./kubectl-custom-cols-0.0.2/templates/*.tpl"
+      to: "templates"
+    bin: "./kubectl-custom-cols"
+  shortDescription: Easy use of custom columns output formating.
+  homepage: https://github.com/webofmars/kubectl-custom-cols
+  caveats: |
+    For now you can't list witch templates are available and adding some can only
```
ahmetb: off-topic: consider adding a path like ~/.kube-custom-cols that your tool recognizes, or an env variable.

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,36 @@
...
+    be done either by doing a PR upstream (see github repo) or by copying a file in 
+    krew store (urgllll).
+  description: |
+    This plugin allows you to pre-define some custom columns file templates that
```
ahmetb: It seems like this is a "kubectl get" replacement, but it doesn't say so anywhere (incl. the tool's own repository).

I'll leave it for you to decide but a better shortDescription could be:

> A "kubectl get" replacement with customizable column presets.

ahmetb:
Very nice plugin! Thanks for submitting it.

plugins/custom-cols.yaml
corneliusweig: The plugin docs mention this:

> You can add you own templates in $HOME/.kube/kube-custom-cols/templates folder.

Although I think this is not working at the moment. @fredleger what's the status there?

corneliusweig:
> My idea was to add it in the next version and for now list it in the documentation.

Well, I couldn't find it in the documentation. If you put it in the docs, it's fine for the first iteration. Still, the `-h`/`--help` flag is very standard and I highly recommend supporting it. This is usually the first thing that people look at after installing.

plugins/custom-cols.yaml
ahmetb: **off-topic:** I'd not add anything serious under ~/.kube, people delete it all the time as it contains kubeconfig that can usually be restored very easily.

we should not block on these stuff.

fredleger: If i understood correctly (and maybe i did not) how krew is working we can't copy anything outside the store/<id> folder right ?

Also i would really like to bundle the templates file with the plugin actually i do think that most of the interest in the plugin is to have a descent collection of predefined templates to use with it (and ones with real use case usage).

If it's possible to copy file outside the store folder my best choice will be `$HOME/.kube-custom-cols` or similar. I agree with @ahmetb about the possible deletion of `.kube` here.

plugins/custom-cols.yaml
fredleger: fixed with a `kubectl-custom-cols-*` pattern

plugins/custom-cols.yaml
fredleger: done

plugins/custom-cols.yaml
fredleger: Nice one thxs

plugins/custom-cols.yaml
corneliusweig: I think it's fine to bundle the column configuration with the plugin. You should only keep in mind that this is not a safe place to store user data, because these directories get wiped when a new version of the plugin is installed.

corneliusweig:
@fredleger I see you updated your help with a listing of available configurations. Keep in mind, that `find` is part of findutils and not present on all machines. Maybe you are better of with `ls` and `sed`.

fredleger:
> Still, the `-h`/`--help` flag is very standard and I highly recommend supporting it. This is usually the > first thing that people look for after installing.

`-h | --help` was implemented and list templates now. Could be better i guess but works fine.

plugins/custom-cols.yaml
fredleger: I tested a bit and it seems my assumption was correct.

My idea is the following now:
- keep the templates in the krew store folder for distributing the plugin with the community templates
- in next version add the ability to search the templates in a list of folders (basically the one bundled and the users overrides/adds in `$HOME/.kube-custom-cols`.

What do you think ?

corneliusweig:
@fredleger Thanks, that looks much better. Have you also seen my comment about the implementation https://github.com/kubernetes-sigs/krew-index/pull/210#issuecomment-524079405?

fredleger:
> @fredleger Thanks, that looks much better. Have you also seen my comment about the > implementation [#210 (comment)](https://github.com/kubernetes-sigs/krew-index/pull/210#issuecomment-524079405)?

Fixed ;-)

fredleger:
@corneliusweig help message is also more explicit with synopsis & examples commands as well

fredleger:
Hi guys any update on the PR ? Is there any point you are waiting in addition of the one already fixed ?

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: custom-cols
+spec:
+  version: "v0.0.2"
```
corneliusweig:
```suggestion
  version: "v0.0.3"
```
The version should be "v0.0.3". I'll merge with v0.0.2, but keep in mind to update this in the future, otherwise krew will not distribute the new version.

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/webofmars/kubectl-custom-cols/archive/v0.0.3.tar.gz
+    sha256: "f83832f5ee0a3e3b44984f29e9ea7ef5b380e23d53da6d52b4a7d62d925ad9f6"
+    files:
+    - from: "./kubectl-custom-cols-*/kubectl-custom-cols"
+      to: "."
+    - from: "./kubectl-custom-cols-*/templates/*.tpl"
+      to: "templates"
+    bin: "./kubectl-custom-cols"
+  shortDescription: Easy use of custom columns output formating.
```
corneliusweig: As @.ahmetb suggested, you could mention `kubectl get` replacement in the short description. I think this will increase the discoverability.

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
...
+  homepage: https://github.com/webofmars/kubectl-custom-cols
+  caveats: |
+    The list of the templates is for now limited (can be retrieved with -h option).
```
corneliusweig:
```suggestion
    The list of templates is currently limited and can be retrieved with the -h option.
```

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
...
+    Please feel free to submit any PR upstream (see github repo) to add more.
```
corneliusweig:
```suggestion
    Please feel free to submit new templates as upstream PRs (see github repo).
```

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
...
+  description: |
+    A "kubectl get" replacement with customizable column presets.
+
+    It's time saving to retrieve some often needed infos like
+    image names, containers names, versions, labels, etc...
```
corneliusweig:
```suggestion
    image names, container names, versions, labels, etc...
```

plugins/custom-cols.yaml
```diff
@@ -0,0 +1,34 @@
...
+    It currently support the following templates (but PRs are welcome upstream):
```
corneliusweig:
```suggestion
    It currently supports the following templates (but PRs are welcome upstream):
```

corneliusweig:
@fredleger ..and sorry for the delay. Next time, I'll respond faster.

plugins/custom-cols.yaml
fredleger: Right, i replaced it in the main repo but not here. FIXED.

plugins/custom-cols.yaml
fredleger: FIXED

corneliusweig:
I think the bot does not like 'oops'

plugins/custom-cols.yaml
fredleger: FIXED

plugins/custom-cols.yaml
corneliusweig: I think you missed this one. The current wording sounds strange, but you can address this in the next revision.

corneliusweig:
Once you fix the commit message, I can merge.

fredleger:
> I think the bot does not like 'oops'

OK it seems it was related to `/` char and `FIX:` word. Fixed and squashed all commits to make it look nicer.

fredleger:
@corneliusweig @ahmetb everything should be fixed now. Any comment / suggestion before merging ?
Thanks for the great work guys.

fredleger:
works great \o/ thanks @corneliusweig

PR #212, opened 2019-08-21
Migrate mtail to new repo and bump version to 1.1.0

grzesuav:
Moved to new repository and added new features, please check
https://gitlab.com/grzesuav/kubectl-mtail/-/releases for details.

Verified locally

### DIFF
--- a/plugins/mtail.yaml
+++ b/plugins/mtail.yaml
@@ -4,18 +4,18 @@ metadata:
   name: mtail
 spec:
   platforms:
-  - head: https://github.com/ahmetb/kubectl-extras/archive/master.zip
-    sha256: 4cf181eb79b1527c20ceff4dba072a8db897ef448e5b276e2d2d70b0a61e6282
-    uri: https://github.com/ahmetb/kubectl-extras/archive/3802e1cae8c037e73d0fec7bc9519c1bca9d25a1.zip
+  - head: https://gitlab.com/grzesuav/kubectl-mtail/-/archive/master/kubectl-mtail-master.zip
+    sha256: 592dca7f2fac476438fea5a5861f7f57c5329e82f76179061e1d689e105150d1
+    uri: https://gitlab.com/grzesuav/kubectl-mtail/uploads/e7e0277d55b21d3004e078b3f67133d0/mtail.tar.gz
     bin: mtail.sh
     files:
-    - from: "/*/mtail/*"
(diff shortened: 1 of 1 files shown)


grzesuav:
Signed CLA

corneliusweig:
So it seems you had an agreement to move `mtail` out of kubectl-extras to its own repository? I found this: https://github.com/ahmetb/kubectl-extras/pull/7#issuecomment-504292816

@ahmetb Do you still want to hand over control of this plugin?

@grzesuav Are you sure with your decision to host on gitlab instead of github? Using a different platform than github will reduce the plugin discoverability.

Apart from those questions, everything looks fine and dandy.

grzesuav:
> So it seems you had an agreement to move `mtail` out of kubectl-extras to its own repository? I found this: [ahmetb/kubectl-extras#7 (comment)](https://github.com/ahmetb/kubectl-extras/pull/7#issuecomment-504292816)
> @grzesuav Are you sure with your decision to host on gitlab instead of github? Using a different platform than github will reduce the plugin discoverability.
> 
Hi, I choose gitlab as it provides more complex solution (CI integration). I am not sure if I understand correctly argument with discoverability, you have in mind that it won't be present in github search and tags ?

ahmetb:
We don't enforce GitHub over GitLab.

In the future we might build tools (such as a tool to automatically pick up new releases and update manifests) and those would work only for GitHub (unless someone is willing to contribute GitLab change).

Right now 100% of the plugins we have are hosted on GitHub which makes it much easier to build such tooling but we definitely don't restrict as for where the source code resides.

ahmetb:
> @ahmetb Do you still want to hand over control of this plugin?

Yeah I am hoping @grzesuav can rewrite it in Go or something.

corneliusweig:
Should we feel the need to host all plugins on GitHub we can do a second repo migration in the future.

/retitle Migrate mtail to new repo and bump version to 1.1.0
/lgtm

grzesuav:
@ahmetb about automatic updates, as I mentioned in https://github.com/eldadru/ksniff/issues/32 , it should be possible to extend renovate to cover this case

ahmetb:
@grzesuav we're planning to create a tool to bump up plugin versions automatically  or manually (join the discussion at https://github.com/kubernetes-sigs/krew/issues/239)

PR #213, opened 2019-08-22
Add plugin: add-config

nasa9084:
- [x] Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md

This is a plugin to add a new kubeconfig into local kubeconfig. Some k8s cluster provider provides the configuration for kubernetes cluster as full-spec kubeconfig file. It is troublesome adding the config file into local existing kubeconfig or using KUBECONFIG environment variable if there has already kubeconfig.

### DIFF
--- a/plugins/add-config.yaml
+++ b/plugins/add-config.yaml
@@ -1,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: add-config
+spec:
+  version: "v0.1.0"
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/nasa9084/kubectl-add_config/releases/download/v0.1.0/kubectl-add_config-v0.1.0.tar.gz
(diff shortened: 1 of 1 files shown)


nasa9084:
fix and squashed ( the url to download archive is mis-configured)

ahmetb:
We already have a plugin called konfig that works like: kubectl konfig merge. 
Do you think this plugin is different enough?

nasa9084:
I'm not deeply sure about konfig though, it looks not to be able to load from stdin, isn't it?
I think konfig merge is designed for merge two (or more?) kubeconfigs, so it is used for existing files.
This plugin is designed for adding a new configuration for a new cluster, copied from cloud providers. So this is not needed to create the new config as file.

corneliusweig:
@nasa9084 The use-case you are referring to relies on `pbpaste`. But then you can achieve the same with `konfig`:

```bash
kubectl konfig merge --save <( pbpaste )
```

I'm not totally opposed to your plugin, but I think it does not add much. Btw, if you find the above command unwieldy, I'm open to PRs for `konfig` :smiley:.

nasa9084:
umm... ok, I understand and I close this PR.
Thank you for your review.

PR #214, opened 2019-08-22
Bump support-bundle and preflight to v0.9.6

divolgin:


corneliusweig:
/lgtm
/approve
Great to see such active development for these plugins!

PR #220, opened 2019-08-30
introduces kubectl-doctor plugin to krew-index

emirozer:
Hi There!

Built the index manifest and tested locally, things are looking OK.
Thanks for the consideration :)

### DIFF
--- a/plugins/doctor.yaml
+++ b/plugins/doctor.yaml
@@ -45,7 +45,7 @@ spec:
       matchLabels:
         os: windows
         arch: amd64
-  shortDescription: kubectl cluster triage plugin - scans your cluster and reports anomalies.
+  shortDescription: Scans your cluster and reports anomalies.
   homepage: https://github.com/emirozer/kubectl-doctor
   description: |
     This plugin is inspired by brew doctor.
--- a/plugins/doctor.yaml
+++ b/plugins/doctor.yaml
@@ -45,12 +45,14 @@ spec:
       matchLabels:
         os: windows
         arch: amd64
-  shortDescription: kubectl cluster triage plugin - (brew doctor equivalent)
+  shortDescription: kubectl cluster triage plugin - scans your cluster and reports anomalies.
   homepage: https://github.com/emirozer/kubectl-doctor
   description: |
     This plugin is inspired by brew doctor.
-     It will scan the active kube-context for anomalies or
+    It will scan the active kube-context for anomalies or
     useful action points that it can report back to you.
-    This plugin does not change any state or configuration,
-    it merely just scans and gathers information than reports back anomalies in yaml format.
-    Please check the repository for an example report.
+    This plugin does not change any state or configuration.
+
+    Please check the repository for an example report:
+
+    * https://github.com/emirozer/kubectl-doctor
--- a/plugins/doctor.yaml
+++ b/plugins/doctor.yaml
@@ -49,7 +49,7 @@ spec:
   homepage: https://github.com/emirozer/kubectl-doctor
   description: |
     This plugin is inspired by brew doctor.
-    It will scan your currently targeted k8s cluster to see if there are anomalies or
+     It will scan the active kube-context for anomalies or
     useful action points that it can report back to you.
     This plugin does not change any state or configuration,
     it merely just scans and gathers information than reports back anomalies in yaml format.
--- a/plugins/doctor.yaml
+++ b/plugins/doctor.yaml
@@ -56,3 +56,7 @@ spec:
     Please check the repository for an example report:
 
     * https://github.com/emirozer/kubectl-doctor
+  caveats: |
+    This plugin needs higher privileges on core API group.
+    Potentially a ClusterRole that can get cluster-scoped resources.
+    Such as nodes / all namespaces etc.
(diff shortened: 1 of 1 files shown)


ahmetb:
Build is failing because we don’t support freebsd. I recommend omiting that for the time being.

emirozer:
cheers, build fixed.

plugins/doctor.yaml
```diff
@@ -0,0 +1,56 @@
...
+        arch: amd64
+  shortDescription: kubectl cluster triage plugin - (brew doctor equivalent)
+  homepage: https://github.com/emirozer/kubectl-doctor
+  description: |
+    This plugin is inspired by brew doctor.
+    It will scan your currently targeted k8s cluster to see if there are anomalies or
+    useful action points that it can report back to you.
+    This plugin does not change any state or configuration,
+    it merely just scans and gathers information than reports back anomalies in yaml format.
+    Please check the repository for an example report.
```
corneliusweig: Would be nice to also provide a link with anchor to the documentation here.

plugins/doctor.yaml
```diff
@@ -0,0 +1,56 @@
...
+    sha256: 1b86784a9dc0673462df5b75bbf7e23772bfb11665feacb91bc8f7d2474711a1
+    bin: kubectl-doctor.exe
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: kubectl cluster triage plugin - (brew doctor equivalent)
```
corneliusweig: Keep in mind that the majority of users do not use `brew` and will therefore not be familiar with `brew doctor`.

plugins/doctor.yaml
```diff
@@ -0,0 +1,56 @@
...
+  homepage: https://github.com/emirozer/kubectl-doctor
+  description: |
+    This plugin is inspired by brew doctor.
+    It will scan your currently targeted k8s cluster to see if there are anomalies or
```
corneliusweig:
```suggestion
     It will scan the active kube-context for anomalies or
```

plugins/doctor.yaml
```diff
@@ -0,0 +1,56 @@
...
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: kubectl cluster triage plugin - (brew doctor equivalent)
+  homepage: https://github.com/emirozer/kubectl-doctor
+  description: |
```
corneliusweig: The description as some weird looking line breaks. It's not super critical, but it would improve readability if those are gone. Remember that line lengths should not exceed 80 characters for proper formatting on small terminals.

emirozer:
@corneliusweig thanks for the review! I will take a look into the points once i have sometime today or tomorrow

plugins/doctor.yaml
emirozer: I can remove that information in parenthesis if you wish. The reason i put it there is that if the user is aware of brew they will probably know what the plugin does immediately.

plugins/doctor.yaml
emirozer: Good call, it is bad formatting apologies!

plugins/doctor.yaml
emirozer: okay i think its better to remove it and leave it as the description of the kubectl-doctor repo.

emirozer:
Added caveats section as well. Ok so for now the only comment that is not looked into is the genericclioptions. I will take a look, the reason i did not use it is that for the first version of the plugin I only wanted to target a cluster and scan everything in it, extra options or limited scanning was for later releases in my mind.

plugins/doctor.yaml
```diff
@@ -0,0 +1,62 @@
...
+    sha256: 1b86784a9dc0673462df5b75bbf7e23772bfb11665feacb91bc8f7d2474711a1
+    bin: kubectl-doctor.exe
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: kubectl cluster triage plugin - scans your cluster and reports anomalies.
```
ahmetb: This is a bit too long.

> Scans your cluster and reports anomalies

is enough.

ahmetb:
/lgtm
I'll have Cornelius /approve.

plugins/doctor.yaml
```diff
@@ -0,0 +1,62 @@
...
+    This plugin is inspired by brew doctor.
+    It will scan the active kube-context for anomalies or
+    useful action points that it can report back to you.
+    This plugin does not change any state or configuration.
+
+    Please check the repository for an example report:
+
+    * https://github.com/emirozer/kubectl-doctor
+  caveats: |
+    This plugin needs higher privileges on core API group.
```
corneliusweig: I think this is good for now. Users for whom this plugin doesn't work will get a hint what's going wrong.

Ideally, you could try running your plugin with reduced privileges and improve the error messages. But that's is your job ;)

plugins/doctor.yaml
corneliusweig: @emirozer I'll wait with merging until you say yes/no here.

emirozer: OK, @ahmetb has a point, ill reduce that to a single sentence

emirozer: Done, I think we can merge this.

corneliusweig:
Great! Thanks for contributing this.
/approve
/lgtm

PR #231, opened 2019-09-03
Added config-file bash plugin

Modulus:

### DIFF
--- a/plugins/config-file.yaml
+++ b/plugins/config-file.yaml
@@ -3,13 +3,13 @@ kind: Plugin
 metadata:
   name: config-file
 spec:
-  version: "v0.1.1"       
+  version: "v0.1.2"       
   platforms:
   - selector:             
       matchExpressions:
       - {key: os, operator: In, values: [darwin, linux]}
-    uri: https://github.com/Modulus/kubectl-config-file/archive/v0.1.1.zip
-    sha256: "644ddfb07c89b567a47ae4786e151351cc46f5c86fad6f7537e83a9c3a347987"
+    uri: https://github.com/Modulus/kubectl-config-file/archive/v0.1.2.zip
+    sha256: "9ee9aa2ac6510157d3531017aa6245e015bdd09f22da9a328109024cb62f76bc"
     files:
     - from: "/kubectl-config-file*/*.sh"
       to: "."              
--- a/plugins/config-file.yaml
+++ b/plugins/config-file.yaml
@@ -1,28 +1,21 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
-  name: config-file               
+  name: config-file
 spec:
-  version: "v0.1.0"      
+  version: "v0.1.0"       
   platforms:
   - selector:             
       matchExpressions:
       - {key: os, operator: In, values: [darwin, linux]}
     uri: https://github.com/Modulus/kubectl-cfg/archive/0.0.6.zip
-    sha256: "41952b7105e2a9f5f1b3fec306bc59c8e972578f9b5d543defeb970c5b5876d5"
+    sha256: "325aa06035e07d6da42ee5c0e44b29b92990b3584589447dcf80b093114b0a43"
     files:
-    - from: "/kubectl-config-file*/*.sh" 
+    - from: "/kubectl-config-file*/*.sh"
       to: "."              
     bin: "./kubectl-config-file.sh"  
-  shortDescription: List config files, and links chosen config file to ~/.kube/config without messing up your files.
-  homepage: https://github.com/Modulus/kubectl-cfg 
-  caveats: |
-    This plugin needs the following programs:
-    * ls(1)
-    * ln(2)
-    * find(3)
-    * grep(4)
-    * cut(5)
-    * echo(6)
+  shortDescription: List available kubeconfigs and link them to ~/.kube/config.
+  homepage: https://github.com/Modulus/kubectl-config-file
   description: |
-    This plugin list all files in ~/.kube folder. Can also link chosen file as ~/.kube/config. So you can work with multiple cluster configs without messing with environment variables.
+    List available kubeconfigs and link them to ~/.kube/config. So you can work
+    with multiple cluster configs without messing with environment variables.
--- a/plugins/config-file.yaml
+++ b/plugins/config-file.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: config-file               
 spec:
-  version: "0.0.6"      
+  version: "v0.1.0"      
   platforms:
   - selector:             
       matchExpressions:
--- a/plugins/config-file.yaml
+++ b/plugins/config-file.yaml
@@ -3,13 +3,13 @@ kind: Plugin
 metadata:
   name: config-file
 spec:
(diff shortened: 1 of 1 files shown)


plugins/config-file.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: config-file               # plugin name must match your manifest file name (e.g. foo.yaml)
+spec:
+  version: "0.0.6"       # required, must be in semver format, prefixed with "v"
+  platforms:
+  # specify installation script for linux and darwin (macOS)
```
ahmetb: do you mind cleaning up these comments? 
they are useful in the docs, but they don't serve a purpose here.

Modulus: Have cleaned up the comments. Should I change the version tag to "v0.0.6"

ahmetb: Yeah, that would be good. I'll go ahead and update the developer guide.

ahmetb: Looks like developer guide is already up-to-date with `v0.0.1`.

plugins/config-file.yaml
```diff
@@ -0,0 +1,28 @@
...
+  caveats: |
+    This plugin needs the following programs:
+    * ls(1)
+    * ln(2)
+    * find(3)
+    * grep(4)
+    * cut(5)
+    * echo(6)
+  description: |
+    This plugin list all files in ~/.kube folder. Can also link chosen file as ~/.kube/config. So you can work with multiple cluster configs without messing with environment variables.
```
ahmetb: please line wrap this to 80 chars.
also off-topic, I think people `rm -rf ~/.kube` all the time, I recommend finding a better path you can recommend people to store their kubeconfigs.

plugins/config-file.yaml
```diff
@@ -0,0 +1,28 @@
...
+  shortDescription: List config files, and links chosen config file to ~/.kube/config without messing up your files.
+  homepage: https://github.com/Modulus/kubectl-cfg 
+  caveats: |
+    This plugin needs the following programs:
+    * ls(1)
+    * ln(2)
+    * find(3)
+    * grep(4)
+    * cut(5)
+    * echo(6)
```
ahmetb: what are these (..) numbers?

plugins/config-file.yaml
```diff
@@ -0,0 +1,28 @@
...
+    sha256: "41952b7105e2a9f5f1b3fec306bc59c8e972578f9b5d543defeb970c5b5876d5"
+    files:
+    - from: "/kubectl-config-file*/*.sh" 
+      to: "."              
+    bin: "./kubectl-config-file.sh"  
+  shortDescription: List config files, and links chosen config file to ~/.kube/config without messing up your files.
+  homepage: https://github.com/Modulus/kubectl-cfg 
+  caveats: |
+    This plugin needs the following programs:
+    * ls(1)
```
ahmetb: Since you're installing on darwin/linux only, I'd say you can skip listing all of these tools. They're all probably preinstalled?

However, make sure in your plugin you check if they exist first in the startup.

plugins/config-file.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - selector:             
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/Modulus/kubectl-cfg/archive/0.0.6.zip
+    sha256: "41952b7105e2a9f5f1b3fec306bc59c8e972578f9b5d543defeb970c5b5876d5"
+    files:
+    - from: "/kubectl-config-file*/*.sh" 
+      to: "."              
+    bin: "./kubectl-config-file.sh"  
+  shortDescription: List config files, and links chosen config file to ~/.kube/config without messing up your files.
```
ahmetb: better alternative:

List available kubeconfigs and link them to ~/.kube/config.

?

plugins/config-file.yaml
Modulus: I have updated this.

plugins/config-file.yaml
Modulus: I have wrapped this line to 80 chars.

plugins/config-file.yaml
Modulus: Have removed this list

plugins/config-file.yaml
Modulus: I have added a plugin check to the shell program

plugins/config-file.yaml
Modulus: I have also changed this. I like your alternative better :)

plugins/config-file.yaml
ahmetb: Can you push the fix?

plugins/config-file.yaml
corneliusweig: As for the `description` field, this one is still unchanged. Can you make sure to push the latest manifest version?

plugins/config-file.yaml
Modulus: Sorry. Done :)

plugins/config-file.yaml
```diff
@@ -0,0 +1,21 @@
...
+    uri: https://github.com/Modulus/kubectl-cfg/archive/0.0.6.zip
+    sha256: "325aa06035e07d6da42ee5c0e44b29b92990b3584589447dcf80b093114b0a43"
+    files:
+    - from: "/kubectl-config-file*/*.sh"
+      to: "."              
+    bin: "./kubectl-config-file.sh"  
+  shortDescription: List available kubeconfigs and link them to ~/.kube/config.
+  homepage: https://github.com/Modulus/kubectl-config-file
+  description: |
+    List available kubeconfigs and link them to ~/.kube/config. So you can work
```
corneliusweig: Just curious: What will your plugin do, if `~/.kube/config` already exists and is not a link?
(No need for action)

Modulus: It will create the link. Meaning you loose the content

Modulus: I have added a new release and added the correct hash and zip file for the current version

Modulus: I have Signed the CLA

Modulus:
If you think so, you may reject the pull requests. I won't argue with you over  this :)

ahmetb:
I understand your point, but trying to give feedback on your approach.
Maybe my understanding of what the plugin does or how it does it is flawed, but if you don't feel like explaining, I won't insist. :)

Since krew-index is a curated repository, we are trying to make sure plugins do what they say without causing major confusion to users. And we're trying to do this curation without limiting the plugin author's creativity. Maybe you're onto something here, and I'm missing it. Feel free to clarify it. Otherwise you're free to close this.

We'll be also adding support for non-centralized plugin indexes hopefully soon so you can directly distribute it from your repo without going through us.

Modulus:
I appreciate your feedback :) But this just started as a small project to learn how to create a krew plugin. I will close this pull request. 

I will rather create a new pull request later if I manage to create a plugin with a more clear purpose. 

I also think the non-centralized plugin index, would be the way to go with this plugin. Thank you for that information :)

Enjoy the rest of your weekend :)

PR #244, opened 2019-09-14
First version

javiermugueta:
Starts or stops all the compute nodes of an OKE cluster

### DIFF
--- a/plugins/okecmd.yaml
+++ b/plugins/okecmd.yaml
@@ -1,0 +1,25 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: okecmd
+spec:
+  version: "v1.0.1"
+  platforms:
+  - selector:
+      matchExpressions:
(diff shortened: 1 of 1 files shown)


ahmetb:
1. What’s OKE?
2. Do you work at OKE and allowed to use their trademark in a plugin name?
3. Plugin name is not descriptive of what it does. Did you read the plugin naming guide?

javiermugueta:
Hi,

1) OKE is Oracle Kubeenetes Engine, same king of acronym such as GKE
2) I work for Oracle, but I don't know if GKE is a trademark (perhaps it
is). I'll circle back to PM and engineering your question
3) I did read it, if fact the first name I choose was different. The
command purpose is to start or stop all the data plane compute nodes of an
specific k8s cluster. Other options are:
oke-cluster-start-stop
oke-cmd-start-stop
Do you think these are better? Any clue?

Regards

ahmetb:
As long as you have the authority to use the name it’s good.

Have you thought of releasing a monolithic plugin like “kubectl oke”?

PR #247, opened 2019-09-22
add modify-secret.yaml

rajatjindal:
Requesting to add the plugin modify-secret to the krew index. 

Happy to address if there are more changes needed.

Thanks
Rajat Jindal

### DIFF
--- a/plugins/modify-secret.yaml
+++ b/plugins/modify-secret.yaml
@@ -1,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: modify-secret
+spec:
+  version: v0.0.4
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-modify-secret/releases/download/v0.0.5/darwin-amd64-v0.0.5.tar.gz
+    sha256: ab078c03ac54f23c957150420fff5c7d6fbfda1e3f99e302a6974ccb6f3d93ac
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-modify-secret/releases/download/v0.0.5/linux-amd64-v0.0.5.tar.gz
+    sha256: b03d08973277ba0c14c032ac2b24c4f24225781dd38efeac7ff33048d8cbfcd2
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  shortDescription: modify secret with implicit base64 translations
+  caveats: |
+    Usage:
+      kubectl modify-secret secret-name -n kube-system
+    
+    Read the documentation at:
+      https://github.com/rajatjindal/kubectl-modify-secret
+  description: |
+    Usage:
+      kubectl modify-secret secret-name -n kube-system
+
+      This plugin fetches the given secret manifest from the cluster, decodes 
+      the payload, opens an editor to make changes, and applies the modified 
+      manifest when done.
+  homepage: https://github.com/rajatjindal/kubectl-modify-secret
+


plugins/modify-secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-modify-secret/releases/download/v0.0.3/linux-amd64-v0.0.3.tar.gz
+    sha256: 9450d2c544859c18ad881e805d9fd9c3ec5bad5f82f6d6fba187f2e29706f672
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  shortDescription: kubectl-modify-secrets allows user to directly modify the secret without worrying about base64 encoding/decoding
```
corneliusweig: The short description gets truncated after ~50 characters when running `krew search`. It improves the discoverability of your plugin if you come up with a more succinct description.

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+  caveats: |
+    Usage:
+      kubectl modify_secret secret-name -n kube-system --kubeconfig /path/to/kube/config
+    
+    Read the documentation at:
+      https://github.com/rajatjindal/kubectl-modify-secret
+    This plugin needs the following programs:
+    * vim
```
corneliusweig: Although you made the right coice :wink:, I think that this is a bit too opinionated. Do you really want to unleash the wrath of the emacs community on you? So could you also support the `EDITOR` env variable?

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+  description: |
+    kubectl-modify-secrets allows user to directly modify the secret without worrying about base64 encoding/decoding
```
corneliusweig: Instead of repeating the short description, you could state how your plugin works:

> This plugin fetches the given secret manifest from the cluster, decodes the payload, opens an editor to make changes, and applies the modified manifest when done.

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  shortDescription: kubectl-modify-secrets allows user to directly modify the secret without worrying about base64 encoding/decoding
+  caveats: |
+    Usage:
+      kubectl modify_secret secret-name -n kube-system --kubeconfig /path/to/kube/config
+    
+    Read the documentation at:
```
corneliusweig: It would be nice to have usage and documentation also in the `description` field.

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,40 @@
...
+    uri: https://github.com/rajatjindal/kubectl-modify-secret/releases/download/v0.0.3/linux-amd64-v0.0.3.tar.gz
+    sha256: 9450d2c544859c18ad881e805d9fd9c3ec5bad5f82f6d6fba187f2e29706f672
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  shortDescription: kubectl-modify-secrets allows user to directly modify the secret without worrying about base64 encoding/decoding
+  caveats: |
+    Usage:
+      kubectl modify_secret secret-name -n kube-system --kubeconfig /path/to/kube/config
```
corneliusweig:
```suggestion
      kubectl modify-secret secret-name -n kube-system --kubeconfig /path/to/kube/config
```
Krew only allows single-word plugin names. So your plugin will become `modify-secret` (and not `modify_secret`).

corneliusweig: I'm not sure if you are show-casing the `--kubeconfig` flag or if it's necessary, but I highly recommend to use the official `genericclioptions` to set these options.

plugins/modify-secret.yaml
corneliusweig: Besides, you can safely assume that _some_ editor is installed on the systems you are targeting. So you don't need to list that as dependency.

plugins/modify-secret.yaml
rajatjindal: it seems like it uses underscores now instead of dashes for some reason. 

https://github.com/kubernetes-sigs/krew/blob/35dcafba7cf3d8f94f1c310a6f6c3bb685a8f8ab/docs/DEVELOPER_GUIDE.md#specifying-plugin-executable

personally I would have preferred it to behave like kubectl plugin naming behavior, but i am sure there must be a good reason for not doing that. (the usage would have been kubectl modify secret .......)

I will make other suggested changes soon, and will update the PR. (all of them are great suggestions, thanks for taking out time for that).

plugins/modify-secret.yaml
rajatjindal: done. changed it to "modify secret with implicit base64 translations"

plugins/modify-secret.yaml
rajatjindal: done. :). I did wanted to change that as one of the next release. better soon then later.

plugins/modify-secret.yaml
rajatjindal: thanks. Used same line. short but descriptive enough.

rajatjindal:
> Hey, thanks for submitting this util to the krew index! This plugin already looks quite good, but let's discuss some details before merging.
> 
>     * I see that you added some common k8s options, like `--kubeconfig` and `--namespace`. Are you aware that there is `k8s.io/cli-runtime/pkg/genericclioptions` which provides a much more complete set of common k8s options? I highly recommend using it, because it improves the usability of your plugin.
> 
>     * You could also think about passing the `-ldflags "-w -s"` options to the linker to reduce the binary size.

I've made these changes

plugins/modify-secret.yaml
rajatjindal: done. moved the plugin to use `genericclioptions`. thanks for the suggestion

rajatjindal: I am a little confused by this. Updated the manifest to use 'modify-secret'

plugins/modify-secret.yaml
rajatjindal: added.

rajatjindal:
Hi @corneliusweig 

Thanks for the detailed review. I've addressed all the comments. Let me know if any more changes are required.

rajatjindal:
gentle reminder @corneliusweig

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-modify-secret/releases/download/v0.0.4/linux-amd64-v0.0.4.tar.gz
+    sha256: a0c9928d25deb587b7c97a6cbab654aae045fa77c08ef3b420062bb86cf95540
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-modify-secret
+  shortDescription: modify secret with implicit base64 translations
```
corneliusweig:
```suggestion
  shortDescription: Modify secret with implicit base64 translations
```

plugins/modify-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+    Usage:
+      kubectl modify-secret secret-name -n kube-system
+    
+    Read the documentation at:
+      https://github.com/rajatjindal/kubectl-modify-secret
+  description: |
+    Usage:
+      kubectl modify-secret secret-name -n kube-system
+
+      This plugin fetches the given secret manifest from the cluster, decodes the payload, opens an editor to make changes, and applies the modified manifest when done.
```
corneliusweig: Can you break this line so that it wraps at 80 characters? This will ensure that it prints nicely for people with small terminal width (80 chars seems to be a common lower limit).

rajatjindal:
> Yeay, from the manifest this is good to go.
> 
> I'm a bit puzzled though about the namespace handling. It looks like you are still using your own flag and not the one from `genericclioptions`. I haven't verified it, but it looks like it's not fully correct like that. Can you have another look there?
> 


its actually using genericclioptions, but its not considering the scenario if '-n namespace' arg is not provided. I will verify this more. thanks for pointing out. (wow !!)


> Also, I've been thinking about the user flow here. AFAICT you are exposing the full manifest to the user for edit. This opens the possibility that the user breaks something which he shouldn't (like change the `kind` or so). Have you considered exposing only the (decoded) data entries? This would change the workflow of your plugin, though..
> 

It does expose only data and not the complete manifest:  https://github.com/rajatjindal/kubectl-modify-secret/blob/master/pkg/cmd/modify_secret.go#L131

(it gets data, saves to a tempfile, and ask user to edit that file).


> @ahmetb Since I haven't heard any objections about naming, I'm assuming you are ok with `modify-secret`.

ahmetb:
Name is ok. 
I am on vacation no need to wait for me here.

corneliusweig:
@rajatjindal Ok that's great. I think this is good to go. Just let me know if you are done with the `namespace` investigation and I'll merge.

rajatjindal:
Hi @corneliusweig 

I believe I've fixed that namespace thing. that was a great code review. thank you for doing that.

Thanks
Rajat Jindal

corneliusweig:
@rajatjindal You're welcome! Thanks for bearing through this and for contributing this nice plugin.
/lgtm
/approve

PR #248, opened 2019-09-23
Add plugin: debug

verb:
This adds a new plugin for `kubectl alpha-debug-pod`, which attaches an ephemeral container to a running pod to be used for debugging.

This plugin relies on an alpha feature to be enabled in the cluster, and users should expect it to be unstable as the feature evolves. I wanted to emphasize this, so I chose a name with "alpha" in it. (Also I didn't want to steal @aylei's plugin name from #135.)

### DIFF
--- a/plugins/debug.yaml
+++ b/plugins/debug.yaml
@@ -1,0 +1,46 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: debug
+spec:
+  version: v0.1.1
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/verb/kubectl-debug/releases/download/v0.1.1/kubectl-debug_darwin_amd64.tar.gz
+    sha256: "6319e06af56ca9510d08638b976103bd67aa246ca055179f71808a0795ac759f"
+    bin: ./kubectl-debug
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/verb/kubectl-debug/releases/download/v0.1.1/kubectl-debug_linux_amd64.tar.gz
+    sha256: "10769e18cc53fcdc0344374c05fbcfffa158a0bd108c6380c5fa4f3b65449008"
+    bin: ./kubectl-debug
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/verb/kubectl-debug/releases/download/v0.1.1/kubectl-debug_windows_amd64.zip
+    sha256: "a4b79d7615f53b775e05e849260c7dbdb153e15676ca8c93aad1353ea9985b42"
+    bin: ./kubectl-debug.exe
+  shortDescription: Attach ephemeral debug container to running pod
+  homepage: https://github.com/verb/kubectl-debug
+  caveats: |
+    This plugin requires the alpha EphemeralContainers feature to be enabled in the cluster.
+    See https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ for
+    enabling features.
+  description: |
+    This plugin attaches an Ephemeral Container to a running pod for use as an interactive
+    debugging session. For example, to run busybox in the running pod "mypod":
+    
+      # kubectl debug mypod --attach
+    
+    See `kubectl debug --help` for additional examples and usage instructions.
+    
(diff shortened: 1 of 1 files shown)


verb:
/assign @ahmetb 

Ahmet has a bit of a context here and I'd appreciate feedback on the naming.

The build that failed is using an old release of krew with an unsupported version of Go. I'm guessing that something changed in krew selector, though. Should I try to get this working with the old version?

plugins/alpha-debug-pod.yaml
```diff
@@ -0,0 +1,38 @@
...
+    uri: https://github.com/verb/kubectl-debug/releases/download/v0.1.0/kubectl-debug_windows_amd64.zip
+    sha256: "fbf3405dad362f1d6f4e766efba4878e3988bb719710fb3701bf150a5e064a2b"
+    bin: ./kubectl-debug.exe
+  shortDescription: Attach ephemeral debug container to running pod
+  homepage: https://github.com/verb/kubectl-debug
+  caveats: |
+    This plugin requires the alpha EphemeralContainers feature to be enabled in the cluster.
+    See https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ for
+    enabling features.
+  description: |
```
corneliusweig: Can you add some basic usage information to the `description` section? Often, it's also included in `caveats` which is fine unless it becomes lengthy.

plugins/alpha-debug-pod.yaml
```diff
@@ -0,0 +1,38 @@
...
+    This plugin attaches an Ephemeral Container to a running pod for use as an interactive
```
corneliusweig: Could you also add a note how to find out about the plugin usage? Like `--help` flag or so.

corneliusweig:
Btw, if you push changes, the CI should pass now :)

verb:
> > I think given that it requires to enable a cluster-side feature gate, it is a good idea to prefix it with `alpha`.
> 
> If we call it `kubectl debug`, and it gets merged into kubectl, it will peacefully transition into using the builtin command due to shadowing of builtin subcommands.

I really don't know whether to expect whether `kubectl debug` will be met with resistance or whether it will undergo a name change.

> I also assume the feature gate can be detected by the plugin? If so, the alpha- prefix isn't really necessary? We can also make a note of it.

It can't, really. We can't query the feature gates of a cluster, so the plugin just tries to add a container and it may fail.

> On the other hand, releasing it `alpha-debug` and renaming it to `debug` later (because, say, KEP did not succeed) will NOT leave its users on a graceful transition path.

naming is hard. :(

If there isn't good support around plugin renaming then maybe we should choose a long-term name. The plugin will continue to work even if the functionality becomes native to kubectl

plugins/alpha-debug-pod.yaml
verb: I'd prefer to rely on the built-in help for usage information so that I don't have to keep it updated in two places. I'll update this with how to get the usage information.

corneliusweig: Yeah, that's fine. :+1:

plugins/alpha-debug-pod.yaml
```diff
@@ -0,0 +1,46 @@
...
+    debugging session. For example, to run busybox in the running pod "mypod":
+    
+      # kubectl alpha-pod-debug mypod --attach
+    
+    See `kubectl alpha-pod-debug --help` for additiona examples and usage instructions.
```
corneliusweig:
```suggestion
    See `kubectl alpha-pod-debug --help` for additional examples and usage instructions.
```

corneliusweig:
Hey @ahmetb. Can you take another look? I think @verb is waiting for us here.

ahmetb:
@corneliusweig I'm mostly waiting for it to be renamed. Feel free to ping me again. Per my/your comment above, `alpha-` prefix isn't really a good idea.

ahmetb:
Ping @verb you can rename the `name` field and the file to either `debug` or `debug-pod` –you're free to use either.

verb:
@ahmetb @corneliusweig renamed!

ahmetb:
/retitle Add plugin: debug

Note: We're making an exception for a generic name because this command has been in the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190212-ephemeral-containers.md) process for a while and has a potential path to graduation, which in case, the builtin would shadow the plugin name gracefully.

/lgtm
/approve

PR #254, opened 2019-09-30
Adds kubectl-fields plugin

rewanth1997:
Requesting to add `fields` plugin to krew index.

An extensive set of testcases have been added to pipeline for robustness testing of new binary. 
Link to source code: [https://github.com/rewanth1997/kubectl-fields](https://github.com/rewanth1997/kubectl-fields)

```console
wget https://raw.githubusercontent.com/rewanth1997/kubectl-fields/master/deploy/krew/plugin.yaml
kubectl krew install --manifest=./plugin.yaml
```

Repo link: https://github.com/rewanth1997/kubectl-fields

Please review the code. I'm happy to make the necessary changes if any.

Regards,
Rewanth

### DIFF
--- a/plugins/fields.yaml
+++ b/plugins/fields.yaml
@@ -4,8 +4,8 @@ metadata:
   name: fields
 spec:
   platforms:
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.0/kubectl-fields-Darwin-x86_64.tar.gz
-    sha256: 2e163e28075a69d4d1dca2b338ec889d967b38fb3e0ed2de1c89993b99ae2889
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Darwin-x86_64.tar.gz
+    sha256: 9fe7b746fa74ba3deedabf77ecb35e7e24b7fc0410dd96219dff4ece02543fdf
     bin: kubectl-fields
     files:
     - from: "*"
@@ -14,8 +14,8 @@ spec:
       matchLabels:
         os: darwin
         arch: amd64
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.0/kubectl-fields-Linux-x86_64.tar.gz
-    sha256: 951739ca1c73dac0729ec46828e677effa80b607efbbf4e79829f7a243126af7
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Linux-x86_64.tar.gz
+    sha256: 21ee6334120bae36ab4a4b23fed731423d81b21058a6df33e92b38110e0fbb2b
     bin: kubectl-fields
     files:
     - from: "*"
@@ -24,8 +24,8 @@ spec:
       matchLabels:
         os: linux
         arch: amd64
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.0/kubectl-fields-Windows-x86_64.tar.gz
-    sha256: 2785c6ccb4b1895c37423192f988dca4f3d4f8222cdce9fa0fdf60810e7c4eda
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Windows-x86_64.zip
+    sha256: 1eea1afb8b623934a240d9cc31923e9ab036524ae3d8a8349b50daced7e99cd7
     bin: kubectl-fields.exe
     files:
     - from: "*"
@@ -34,11 +34,11 @@ spec:
       matchLabels:
         os: windows
         arch: amd64
-  version: v1.0.0
+  version: v1.0.1
   homepage: https://github.com/rewanth1997/kubectl-fields
   caveats: |
     Usage:
-      $ kubectl fields [options] <RESOURCE> <PATTERN>
+      $ kubectl fields <RESOURCE> [options] <PATTERN> [<PATTERN 2> ...]
 
     Options:
       -h, --help          help for kubectl-fields
@@ -59,7 +59,7 @@ spec:
     containers.securityContext.capabilities
     initContainers.securityContext.capabilities
 
-    $ kubectl fields -i svc ip
+    $ kubectl fields svc -i ip
     spec.clusterIP
     spec.externalIPs
     spec.loadBalancerIP
--- a/plugins/fields.yaml
+++ b/plugins/fields.yaml
@@ -4,8 +4,8 @@ metadata:
   name: fields
 spec:
   platforms:
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Darwin-x86_64.tar.gz
-    sha256: 9fe7b746fa74ba3deedabf77ecb35e7e24b7fc0410dd96219dff4ece02543fdf
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.1.1-beta/kubectl-fields-Darwin-x86_64.tar.gz
+    sha256: ac82ee062e8bf1a9c6f6fab2e718e6d560c4ca1855616c2a67b0411792b575bc
     bin: kubectl-fields
     files:
     - from: "*"
@@ -14,8 +14,8 @@ spec:
       matchLabels:
         os: darwin
         arch: amd64
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Linux-x86_64.tar.gz
-    sha256: 21ee6334120bae36ab4a4b23fed731423d81b21058a6df33e92b38110e0fbb2b
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.1.1-beta/kubectl-fields-Linux-x86_64.tar.gz
+    sha256: 051140e9df85e7c3b982fc7b4199084441bbfc6328624f9a0dff07193058f135
     bin: kubectl-fields
     files:
     - from: "*"
@@ -24,8 +24,8 @@ spec:
       matchLabels:
         os: linux
         arch: amd64
-  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.1/kubectl-fields-Windows-x86_64.zip
-    sha256: 1eea1afb8b623934a240d9cc31923e9ab036524ae3d8a8349b50daced7e99cd7
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.1.1-beta/kubectl-fields-Windows-x86_64.zip
+    sha256: d188b7d2d1c567d7b074a3d8eb6c075aa8713e483074b839ff0da0300a8c710a
     bin: kubectl-fields.exe
     files:
     - from: "*"
--- a/plugins/fields.yaml
+++ b/plugins/fields.yaml
@@ -34,7 +34,7 @@ spec:
       matchLabels:
         os: windows
         arch: amd64
-  version: v1.0.1
+  version: v1.1.1-beta
   homepage: https://github.com/rewanth1997/kubectl-fields
   caveats: |
     Usage:
--- a/plugins/fields.yaml
+++ b/plugins/fields.yaml
@@ -38,15 +38,9 @@ spec:
   homepage: https://github.com/rewanth1997/kubectl-fields
   caveats: |
     Usage:
-      $ kubectl fields <RESOURCE> [options] <PATTERN>
+      kubectl fields <RESOURCE> [options] <PATTERN>
 
-    Options:
-      -I, --case-sensitive   Case sensitive pattern match
-      -h, --help             help for kubectl-fields
-          --no-color         Do not print colored output
-          --stdin            Expects input via pipes
-
-    Read the documentation:
+    Documentation:
       https://github.com/rewanth1997/kubectl-fields/tree/v1.2.0-beta
   shortDescription: Grep resources hierarchy by field name
   description: |
--- a/plugins/fields.yaml
+++ b/plugins/fields.yaml
@@ -1,0 +1,78 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: fields
+spec:
+  platforms:
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.0/kubectl-fields-Darwin-x86_64.tar.gz
+    sha256: 2e163e28075a69d4d1dca2b338ec889d967b38fb3e0ed2de1c89993b99ae2889
+    bin: kubectl-fields
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  - uri: https://github.com/rewanth1997/kubectl-fields/releases/download/v1.0.0/kubectl-fields-Linux-x86_64.tar.gz
+    sha256: 951739ca1c73dac0729ec46828e677effa80b607efbbf4e79829f7a243126af7
+    bin: kubectl-fields
+    files:
+    - from: "*"
+      to: "."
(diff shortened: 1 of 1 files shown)


plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+        arch: amd64
+  version: v1.0.0
+  homepage: https://github.com/rewanth1997/kubectl-fields
+  caveats: |
+    Usage:
+      $ kubectl fields [options] <RESOURCE> <PATTERN>
+
+    Options:
+      -h, --help          help for kubectl-fields
+      -i, --ignore-case   Ignore case distinction
```
luksa: What does this apply to? The search pattern or the resource? If the pattern, then I suggest you change the example below from `kubectl fields -i svc ip` to `kubectl fields svc -i ip` (it confused me when I first saw the example).

luksa: Ah, and ignore-case should probably be true by default, with the option to turn it off. I think it would make the UX better (after all, does a user searching for a field know how the field is capitalized?)

rewanth1997: `-i` is a non-positional parameter. It can be used anywhere. I think just changing the example in the manifest will resolve the issue.

`kubectl fields -i svc ip` == `kubectl fields svc -i ip` == `kubectl fields svc ip -i`

Coming to the second question, I observed if the default value of `ignore-case` is true it will return lots of results that might not be required. Also, I was following `grep` convention just to make it easy for the users. Please let us know what you think.

```
rewanth@ubuntu:/$ kubectl fields daemonset -i ver
apiVersion
metadata.initializers.result.apiVersion
metadata.initializers.result.metadata.resourceVersion
metadata.managedFields.apiVersion
metadata.ownerReferences.apiVersion
metadata.resourceVersion
spec.template.metadata.initializers.result.apiVersion
spec.template.metadata.initializers.result.metadata.resourceVersion
spec.template.metadata.managedFields.apiVersion
spec.template.metadata.ownerReferences.apiVersion
spec.template.metadata.resourceVersion
spec.template.spec.containers.env.valueFrom.fieldRef.apiVersion
spec.template.spec.dnsConfig.nameservers
spec.template.spec.initContainers.env.valueFrom.fieldRef.apiVersion
spec.template.spec.volumes.csi.driver
spec.template.spec.volumes.downwardAPI.items.fieldRef.apiVersion
spec.template.spec.volumes.flexVolume.driver
spec.template.spec.volumes.iscsi.chapAuthDiscovery
spec.template.spec.volumes.nfs.server
spec.template.spec.volumes.projected.sources.downwardAPI.items.fieldRef.apiVersion
```
```
rewanth@ubuntu:/$ kubectl fields daemonset ver
spec.template.spec.dnsConfig.nameservers
spec.template.spec.volumes.csi.driver
spec.template.spec.volumes.flexVolume.driver
spec.template.spec.volumes.iscsi.chapAuthDiscovery
spec.template.spec.volumes.nfs.server
```

rewanth1997:
Please review the updated plugin file according to new release. I will squash the commits after approval.

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+      --stdin         Expects input via pipes
+
+    Read the documentation:
+      https://github.com/rewanth1997/kubectl-fields/blob/master/README.md
+  shortDescription: Kubectl resources hierarchy parser
+  description: |
+    Kubectl resources hierarchy parser.
```
corneliusweig: I don't understand what is the parsing that this plugin does.

To me the examples look more like a variant of grep on `kubectl explain --recursive`.

---

Looking at the description from the project homepage, this seems to be true:

> Kubectl-fields is a cli tool to parse kubectl explain --recursive output and grep matching pattern in one-liner hierarchy format.

I think this is much more descriptive of what the plugin is actually doing. Can you modify the description here?

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+    initContainers.securityContext.capabilities
+
+    $ kubectl fields svc -i ip
+    spec.clusterIP
+    spec.externalIPs
+    spec.loadBalancerIP
+    spec.sessionAffinityConfig.clientIP
+    status.loadBalancer.ingress.ip
+
+    Additional kubectl-fields example (the hard way not recommended). Developed to run tests on pipeline
```
corneliusweig: In what regard is the `--stdin` mode different? As far as I can tell, the plugin usually does the `kubectl explain --recursive` internally and this option is a fallback.

So I think this is better left for `--help` of this plugin. It will just clutter the users' terminals and not add much for getting started with this plugin (remember that all users will see the tail of the description first, due to how terminals work).

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+    More on https://github.com/rewanth1997/kubectl-fields
+
+    Examples:
+
+    $ kubectl fields po.spec capa
+    containers.securityContext.capabilities
+    initContainers.securityContext.capabilities
+
+    $ kubectl fields svc -i ip
+    spec.clusterIP
```
corneliusweig: Even if it's not correct, can you limit the example output to three lines here? Otherwise you are giving away a lot of space for uninteresting output. For this example, it's more important to say what the `-i` option does, like

> Find resource field names case-insensitively:

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+      https://github.com/rewanth1997/kubectl-fields/blob/master/README.md
+  shortDescription: Kubectl resources hierarchy parser
+  description: |
+    Kubectl resources hierarchy parser.
+
+    More on https://github.com/rewanth1997/kubectl-fields
+
+    Examples:
+
+    $ kubectl fields po.spec capa
```
corneliusweig: For this example, it would be nice to have a short description of what this plugin does, like

> Find resource field names case-sensitively:

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+    containers.securityContext.capabilities
```
corneliusweig: Even though the plugin output won't work like this, you could think about highlighting the matching portion of the string in bold using shell escape sequences. This will help to understand how the plugin works.

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+    Usage:
+      $ kubectl fields <RESOURCE> [options] <PATTERN> [<PATTERN 2> ...]
+
+    Options:
+      -h, --help          help for kubectl-fields
+      -i, --ignore-case   Ignore case distinction
+      --stdin         Expects input via pipes
+
+    Read the documentation:
+      https://github.com/rewanth1997/kubectl-fields/blob/master/README.md
```
corneliusweig: You might want to point your documentation to the _exact_ release instead of master. They will differ when you release new versions. I.e. `https://github.com/rewanth1997/kubectl-fields/tree/v1.0.1`

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: v1.0.1
+  homepage: https://github.com/rewanth1997/kubectl-fields
+  caveats: |
+    Usage:
+      $ kubectl fields <RESOURCE> [options] <PATTERN> [<PATTERN 2> ...]
```
corneliusweig: Can you also mention how several patterns are combined? Are they AND'ed or OR'ed?

plugins/fields.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    Options:
+      -h, --help          help for kubectl-fields
+      -i, --ignore-case   Ignore case distinction
+      --stdin         Expects input via pipes
+
+    Read the documentation:
+      https://github.com/rewanth1997/kubectl-fields/blob/master/README.md
+  shortDescription: Kubectl resources hierarchy parser
```
corneliusweig:
```suggestion
  shortDescription: Grep resource schema by field name
```

luksa:
Just giving my two cents here: to novice users, a name like "grep-schema" would most likely just be confusing. For example, does a novice user know what "schema" refers to? A k8s maintainer might immediately know, but a k8s _user_ knows only that k8s has resources and those resources have fields/attributes.

So I think a better name would be something like "search-resource-fields" or shortened to "search-fields". 

Why "search" and not e.g. "find"?  -> `yum search`, `dnf search`, `apt-cache search`, ...
Why "fields" and not e.g. "attributes"? -> `kubectl explain` already calls them "fields".

plugins/fields.yaml
rewanth1997: Description updated

plugins/fields.yaml
rewanth1997: You are absolutely right. I will remove the additional examples. It won't be required for normal users.

plugins/fields.yaml
rewanth1997: Updated with new example

plugins/fields.yaml
rewanth1997: Description added

plugins/fields.yaml
rewanth1997: I added colors to this using https://github.com/fatih/color
Please review it

plugins/fields.yaml
rewanth1997: Updated the reference link

plugins/fields.yaml
rewanth1997: They are OR'ed same as `grep`. Adding more details make the help menu complex. So, I removed the optional `"PATTERN 2"` attribute and added an example in README. I hope that should work.

plugins/fields.yaml
rewanth1997: I made a modification to the suggested description. Please check the commit

rewanth1997:
The coloring scheme is no supported by windows. The plugin detects the OS during runtime and disables the coloring scheme automatically if the plugin is being used in windows. I hope this kind of check should be good enough.

windows

luksa:
@rewanth1997 btw, I still think you should default to case-insensitive searching. Imagine a new user running `kubectl fields pod NFS` and getting no results, then spend a lot of time figuring out what's wrong, only to eventually realize that he should have typed `nfs` in lowercase. 

The whole point of the plugin is to be able to find fields by their partial name. This implies that the user may not know the actual field name, so it's better for the plugin to possibly show unnecessary results than not showing any results at all.

rewanth1997:
This completely makes sense, @luksa. I will change the default search to case-insensitive matching in the next release.

plugins/fields.yaml
corneliusweig: Regarding colors on windows terminals, this might be of interest to you: https://github.com/Azure/draft/pull/928

plugins/fields.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    Options:
+      -h, --help          help for kubectl-fields
+      -i, --ignore-case   Ignore case distinction
+      --stdin         Expects input via pipes
+
+    Read the documentation:
+      https://github.com/rewanth1997/kubectl-fields/tree/v1.0.1
+  shortDescription: Grep resources hierarchy by field name
+  description: |
```
corneliusweig: Can you wrap the description at 80 characters, so it is nicely displayed on smaller terminals?

corneliusweig:
> to novice users, a name like "grep-schema" would most likely just be confusing. For example, does a novice user know what "schema" refers to? A k8s maintainer might immediately know, but a k8s user knows only that k8s has resources and those resources have fields/attributes.

Although this comment may apply to a general audience, this plugin says about [itself](https://github.com/rewanth1997/kubectl-fields#kubectl-resources-hierarchy-parsing-tool)

> This tool is exclusively built for CKA/CKAD applicants.

...and those kind of people _should_ already be familiar with resource schemas :D. Besides, schemas are a standard concept.

My main concern about `fields` is that it does not convey that this is not about a concrete resource of which I want to see some field _value_, but find some field-name in the resource schema.

---

Maybe @ahmetb can also share his thoughts about plugin naming here?

luksa:
> Although this comment may apply to a general audience, this plugin says about [itself](https://github.com/rewanth1997/kubectl-fields#kubectl-resources-hierarchy-parsing-tool)
> 
> > This tool is exclusively built for CKA/CKAD applicants.
> 
> ...and those kind of people _should_ already be familiar with resource schemas :D. Besides, schemas are a standard concept.

I actually disagree that this is only for exam applicants. I see all new users using this, if they are aware of this plugin. Same as with kubectl explain. 

> My main concern about `fields` is that it does not convey that this is not about a concrete resource of which I want to see some field _value_, but find some field-name in the resource schema.

I do agree that `fields` is not specific enough, but `search-fields` should be ok. It's clear what it does and it can't be mistaken for doing anything else.

plugins/fields.yaml
rewanth1997:
```
kubectl-fields parses specified kubectl resources to match given pattern(s).
It prints matched fields parental hierarchy in one-liner format.
```

I guess, this kind of formatting should work well

plugins/fields.yaml
rewanth1997: Perfect. Done.

rewanth1997:
@luksa : I defaulted the pattern matching to case-insensitive

@corneliusweig : I added color support to windows as well. This update will print colored output on all platforms.

The output of the new beta release

**Default match**
default-match

**Case sensitive match**
case-sensitive-match

**Combination of all**
combo

**Output on windows cmd**
windows-all

rewanth1997:
I believe we are left only with resolving the name of the plugin.
Please review the latest commit. I will make the necessary changes if any.

ahmetb:
I think the `fields` name is good enough. I don't foresee it being easily used elsewhere. When we say "fields", we usually refer to type members on Kubernetes API objects.

`search-fields` is definitely better, but it doesn't add much to usability/discoverability. I'll vote +1 for `fields`.

plugins/fields.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: v1.2.0-beta
+  homepage: https://github.com/rewanth1997/kubectl-fields
+  caveats: |
+    Usage:
```
ahmetb: we don't need to copy usage string here.
you'll almost certainly forget to keep it up-to-date.

plugins/fields.yaml
```diff
@@ -0,0 +1,70 @@
...
+    $ kubectl fields po.spec capa
+    containers.securityContext.capabilities
+    initContainers.securityContext.capabilities
+
+    Find resource field names case-insensitively:
+    $ kubectl fields svc -i affinity
+    spec.sessionAffinity
+    spec.sessionAffinityConfig`
+
+    No more greps, pipes and scrolling up and down to find hierarchial order for resources :-)
```
ahmetb: I think this last line doesn't add much.
It's pretty clear from above what it does.

Also perhaps move examples to `-h`/`--help` of your program instead.

rewanth1997: Adding more examples increases cluttering in the terminal as @corneliusweig mentioned earlier. Hence I added minimal examples. I added more examples as screenshots to the README [examples](https://github.com/rewanth1997/kubectl-fields#examples) section.

Please share the other apt examples to be added to the help menu in your opinion.

plugins/fields.yaml
rewanth1997: Wow, thanks. That will be really helpful for me. I miss this often.

rewanth1997:
@corneliusweig : Last line is removed as suggested
@ahmetb @luksa Please let me know if any further changes are required. If not, I will squash the commits and update the release from `v1.2.0-beta` to `v1.2.0`

corneliusweig:
Let's get this on the way. The next version can be released automatically if you only change the version. Thanks for enduring this lengthy review process!
/lgtm
/approve

rewanth1997:
Thanks, @corneliusweig @luksa @ahmetb for your valuable inputs on making this plugin better. I hope the community finds this useful.

corneliusweig:
I think I'm gonna use it :)

PR #259, opened 2019-10-01
Add plugin: kubeval

nlamirault:
This plugin allow for running `kubeval` , a tool for validating a Kubernetes YAML or JSON configuration file. 
It does so using schemas generated from the Kubernetes OpenAPI specification, and therefore can validate schemas for multiple versions of Kubernetes.

### DIFF
--- a/plugins/kubeval.yaml
+++ b/plugins/kubeval.yaml
@@ -1,12 +1,12 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼kind: Plugin
 ￼metadata:
-￼  name: validate-config
+￼  name: kubeval
 ￼spec:
 ￼  version: "0.14.0"
-￼  shortDescription: Validate your Kubernetes configuration files
+￼  shortDescription: Validate schema of resource files
   description: |
-    The plugin validate a Kubernetes YAML or JSON configuration file.
+    The plugin validates a Kubernetes YAML or JSON configuration file.
     It does so using schemas generated from the Kubernetes OpenAPI specification,
     and therefore can validate schemas for multiple versions of Kubernetes
   homepage: https://github.com/instrumenta/kubeval
@@ -17,6 +17,8 @@ apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼      files:
 ￼      - from: "./kubeval"
 ￼        to: "."
+      - from: LICENSE
+        to: .
 ￼      selector:
 ￼        matchLabels:
 ￼          os: linux
@@ -25,8 +27,10 @@ apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼      sha256: "05b9993e59a32b95bc5496dd98598e4a425643151efd108c0449405dd4a2a4c7"
 ￼      bin: kubeval
 ￼      files:
-￼        - from: "./kubeval"
-￼          to: "."
+￼      - from: "./kubeval"
+￼        to: "."
+￼      - from: LICENSE
+        to: .
 ￼      selector:
 ￼        matchLabels:
 ￼          os: darwin
@@ -35,8 +39,10 @@ apiVersion: krew.googlecontainertools.github.com/v1alpha2
       sha256: "2a844518981848a7d77cced9b51a05174ba9c17fc007a1c48cd2af0d3fb021d7"
 ￼      bin: kubeval.exe
 ￼      files:
-￼        - from: "./kubeval.exe"
-￼          to: "."
+￼      - from: "./kubeval.exe"
+￼        to: "."
+￼      - from: LICENSE
+        to: .
 ￼      selector:
 ￼        matchLabels:
 ￼          os: windows
--- a/plugins/kubeval.yaml
+++ b/plugins/kubeval.yaml
@@ -19,7 +19,7 @@ apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼          arch: amd64
 ￼    - uri: https://github.com/instrumenta/kubeval/releases/download/0.14.0/kubeval-darwin-amd64.tar.gz
 ￼      sha256: "05b9993e59a32b95bc5496dd98598e4a425643151efd108c0449405dd4a2a4c7"
-￼      bin: kubectl-kubeval.sh
+￼      bin: kubeval
 ￼      files:
 ￼        - from: "./kubeval"
 ￼          to: "."
@@ -29,9 +29,9 @@ apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼          arch: amd64
     - uri: https://github.com/instrumenta/kubeval/releases/download/0.14.0/kubeval-windows-amd64.zip
       sha256: "2a844518981848a7d77cced9b51a05174ba9c17fc007a1c48cd2af0d3fb021d7"
-￼      bin: kubectl-kubeval.sh
+￼      bin: kubeval.exe
 ￼      files:
-￼        - from: "./kubeval"
+￼        - from: "./kubeval.exe"
 ￼          to: "."
 ￼      selector:
 ￼        matchLabels:
--- a/plugins/kubeval.yaml
+++ b/plugins/kubeval.yaml
@@ -1,11 +1,15 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 ￼kind: Plugin
 ￼metadata:
-￼  name: validate-manifests
+￼  name: validate-config
 ￼spec:
 ￼  version: "0.14.0"
-￼  shortDescription: Validate your Kubernetes configuration files, supports multiple Kubernetes versions
-￼  homepage: https://github.com/instrumenta/kubeval
+￼  shortDescription: Validate your Kubernetes configuration files
+  description: |
+    The plugin validate a Kubernetes YAML or JSON configuration file.
+    It does so using schemas generated from the Kubernetes OpenAPI specification,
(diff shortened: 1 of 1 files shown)


plugins/kubeval.yaml
```diff
@@ -0,0 +1,39 @@
...
+￼      files:
+￼        - from: "./kubeval"
+￼          to: "."
+￼      selector:
+￼        matchLabels:
+￼          os: darwin
+￼          arch: amd64
+    - uri: https://github.com/instrumenta/kubeval/releases/download/0.14.0/kubeval-windows-amd64.zip
+      sha256: "2a844518981848a7d77cced9b51a05174ba9c17fc007a1c48cd2af0d3fb021d7"
+￼      bin: kubectl-kubeval.sh
```
ahmetb: are you sure the windows entrypoint is a .sh ?

plugins/kubeval.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+￼kind: Plugin
+￼metadata:
+￼  name: validate-manifests
+￼spec:
+￼  version: "0.14.0"
+￼  shortDescription: Validate your Kubernetes configuration files, supports multiple Kubernetes versions
```
ahmetb: > supports multiple Kubernetes versions

part is redundant for a "short" description, it's better to actually keep it short.

ahmetb: you can also use the `description` field if necessary.

ahmetb:
Normally we don't accept plugins with kube* prefix, but since this is a well-established tool it qualifies for an exception to preserve its name.

However, this also might be an opportunity for you to rename and distribute your tool as `kubectl validate-config`.

I also recommend making sure the help (-h/--help) message for `kubectl kubeval` shows invocation/examples as "kubectl kubeval" and not just "kubeval" when invoked as a plugin.

ahmetb:
@nlamirault what’s your take on the naming? Should we keep the existing name fi it's well-recognized, or do you want your users to be primarily invoking it like `kubectl validate-config`?

nlamirault:
we can ask the author of the plugin what he thinks ?
Ping @garethr

corneliusweig:
@garethr Have you seen the PR regarding `kubeval` in `krew-index`? We are discussing how the plugin should be named. Maybe you can share your thoughts here?

garethr:
Mmm, turning it into a general question about Krew index opinions again :)

`validate-config` would basically mean `kubeval` is _the_ way of validating config. If something else exists, or comes along later, it's locked out or hidden behind an upstream abstraction or has to adhear to the same interface.

For that reason, I'd suggest plugins carry the name of the underlying tool. That also makes support less surprising. If kubeval is hidden then users won't find it easy to file issues in the right place. But I'm of the view the Krew index should be less opinionated I think.

ahmetb:
Regarding **naming**:

- We've been preserving the original names of plugins that are already popular. We can do the same here. 
- `kubectl kubeval` sounds odd (and the stutter is something we try to avoid), and as Gareth said, `kubectl validate-config` hides the fact that it's actually kubeval.

Regarding **distributing as a plugin:**

kubeval [already has packaging](https://kubeval.instrumenta.dev/installation/) for Windows, Linux, macOS. Why does it make sense to add yet another distribution (krew)?

At this point, I don't think it even makes sense to stop distributing with existing methods and use Krew. That would leave existing users behind. Also, krew helps plugins which haven't created distributions for multiple OSes yet, but kubeval is already there.

Does kubeval benefit from being a krew plugin? Maybe the users can benefit, as they can declaratively install most of their kubernetes client-side tooling via krew. But overall, I'm not even sure if that's a huge benefit.

garethr:
> Does kubeval benefit from being a krew plugin? Maybe the users can benefit, as they can declaratively install most of their kubernetes client-side tooling via krew. But overall, I'm not even sure if that's a huge benefit.

From a users perspective, discovery and central management are useful in my view. That's normally the whole point of a package manager. See Homebrew as a good example.

I think the Krew project needs to make clear it's not intending to be a generic package manager for Kubectl plugins, rather it's a very opinionated distribution of Kubectl plugins.

I think that's fine, but I do think that leaves a gap in the market for a registry/tool for discovering/ managing any and all kubectl plugins. At least until/unless Krew adds the ability to have additional search paths (at which point I'd be happy to create the equivalent of a Homebrew tap) for this plugin definition.

corneliusweig:
So if I understand both of you correctly, you both agree (for maybe different reasons) to _not_ include `kubeval` as a plugin in `krew`, right?

garethr:
@corneliusweig I'd be happy to see it included under `kubeval`. I think it would aid discovery and, for some users, aid installation and management.

ahmetb:
@garethr Yeah we can release it as a plugin. Some cooperation required from your side would be:

- Make releases on Krew for new versions occassionally (it's on our radar to bump plugin versions automatically via github releases, we're not there yet).
- Make sure your `--help` message and other USAGE strings show `kubectl kubeval` when the program is invoked as a plugin (you can tell this via argv[0] ends with `kubectl-kubeval`).

At any rate, please review this PR for accuracy of description etc.

garethr:
@ahmetb sounds good. Issue on kubeval opened https://github.com/instrumenta/kubeval/issues/187

Not sure if you're using CODEOWNERS or similar? I'm seeing some multi-user repos do similar. Happy to be added if useful.

The description and other details look correct, thanks @nlamirault for submitting.

ahmetb:
I am not sure if Kubernetes/prow has support for CODEOWNERS so we aren’t doing that. Also I am not sure how it is relevant and what you have in mind to achieve with it.

garethr:
@ahmetb spurred by a conversation today regarding tekton https://github.com/tektoncd/catalog/pull/104/files. They are starting to use owners as a way of scaling shared content and routing issues.

ahmetb:
Yeah, we don't have a per-plugin directory structure to support that. We already have the standard OWNERS files and we know who the owners are. I hope I'm not missing something.

ahmetb:
@nlamirault do you have time to help https://github.com/instrumenta/kubeval/issues/187, then we can merge this as `kubectl kubeval`. 

/retitle Add plugin: kubeval

plugins/kubeval.yaml
```diff
@@ -0,0 +1,43 @@
...
+    It does so using schemas generated from the Kubernetes OpenAPI specification,
+    and therefore can validate schemas for multiple versions of Kubernetes
+  homepage: https://github.com/instrumenta/kubeval
+￼  platforms:
+￼    - uri: https://github.com/instrumenta/kubeval/releases/download/0.14.0/kubeval-linux-amd64.tar.gz
+￼      sha256: "8b1b8c63df9ee3206113a3352e07474ea071610cfacf40a64db74c56e163f1e2"
+￼      bin: kubeval
+￼      files:
+￼      - from: "./kubeval"
+￼        to: "."
```
corneliusweig: Please also install a `LICENSE` file

plugins/kubeval.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+￼kind: Plugin
+￼metadata:
+￼  name: validate-config
+￼spec:
+￼  version: "0.14.0"
+￼  shortDescription: Validate your Kubernetes configuration files
```
corneliusweig:
```suggestion
￼  shortDescription: Validate schema of resource files
```

plugins/kubeval.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+￼kind: Plugin
+￼metadata:
+￼  name: validate-config
```
corneliusweig:
```suggestion
￼  name: kubeval
```

ahmetb:
If nobody’s signing up to maintain this manifest, I suggest we don’t proceed. We equally need support from kubeval to look like a plugin (help msgs etc, and release bundle not breaking).

nlamirault:
@ahmetb @corneliusweig i just push a commit with your remarks

corneliusweig:
So to sum up @ahmetb's concerns:

1. Who will maintain this manifest? My current assumption is that @nlamirault wants to do this, correct?
2. Release bundle not breaking. This is also @nlamirault's task if he's going to maintain this manifest.
3. Help messages for plugin usage: because the plugin is called the same as the standalone, this should be fine for now. Or do you see potential for big confusion here?

Provided that @nlamirault agrees to maintain this manifest, I'm still supportive of merging.

plugins/kubeval.yaml
```diff
@@ -0,0 +1,49 @@
...
+￼metadata:
+￼  name: kubeval
+￼spec:
+￼  version: "0.14.0"
+￼  shortDescription: Validate schema of resource files
+  description: |
+    The plugin validates a Kubernetes YAML or JSON configuration file.
+    It does so using schemas generated from the Kubernetes OpenAPI specification,
+    and therefore can validate schemas for multiple versions of Kubernetes
+  homepage: https://github.com/instrumenta/kubeval
```
corneliusweig: The indentation of these two fields is wrong. Please fix this to get the CI pass.

corneliusweig:
@nlamirault Nothing has happened here for almost a month. If you don't plan to continue, could you please close the issue?

ahmetb:
/close
Kubeval is already popular outside krew and there doesn’t seem to be enough interest to pursue this. So I will close.

PR #267, opened 2019-10-07
Add kubectl-passman @ 0.0.1 to krew index

chrisns:

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -42,6 +42,7 @@ Name | Description | Stars
 [oidc-login](https://github.com/int128/kubelogin) | Log in to the OpenID Connect provider | ![GitHub stars](https://img.shields.io/github/stars/int128/kubelogin.svg?label=stars&logo=github)
 [open-svc](https://github.com/superbrothers/kubectl-open-svc-plugin) | Open the Kubernetes URL(s) for the specified service in your browser. | ![GitHub stars](https://img.shields.io/github/stars/superbrothers/kubectl-open-svc-plugin.svg?label=stars&logo=github)
 [outdated](https://github.com/replicatedhq/outdated) | Finds outdated container images running in a cluster | ![GitHub stars](https://img.shields.io/github/stars/replicatedhq/outdated.svg?label=stars&logo=github)
+[passman](https://github.com/chrisns/kubectl-passman) | Provides glue between common password managers and kubectl for user credentials | ![GitHub stars](https://img.shields.io/github/stars/chrisns/kubectl-passman.svg?label=stars&logo=github)
 [pod-logs](https://github.com/danisla/kubefunc) | Display a list of pods to get logs from | ![GitHub stars](https://img.shields.io/github/stars/danisla/kubefunc.svg?label=stars&logo=github)
 [pod-shell](https://github.com/danisla/kubefunc) | Display a list of pods to execute a shell in | ![GitHub stars](https://img.shields.io/github/stars/danisla/kubefunc.svg?label=stars&logo=github)
 [preflight](https://github.com/replicatedhq/troubleshoot) | Executes application preflight tests in a cluster | ![GitHub stars](https://img.shields.io/github/stars/replicatedhq/troubleshoot.svg?label=stars&logo=github)
--- a/plugins/passman.yaml
+++ b/plugins/passman.yaml
@@ -1,0 +1,277 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: passman
+spec:
+  version: "v0.0.1"
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/chrisns/kubectl-passman/releases/download/v0.0.1/kubectl-passman-darwin-amd64.zip
+      sha256: "1bb5e6b37303229ce4cb157fd4348757b9f9eed0f7db99ad7e054bb76630d29c"
+      bin: "./kubectl-passman"
+      files:
+        - from: kubectl-passman-darwin-amd64
+          to: kubectl-passman
+
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: 386
+      uri: https://github.com/chrisns/kubectl-passman/releases/download/v0.0.1/kubectl-passman-darwin-386.zip
+      sha256: "340286ac7bda27b3d58c4fdd17ff921633ca6fc09dcb77585b88e44370d7b45c"
+      bin: "./kubectl-passman"
+      files:
+        - from: kubectl-passman-darwin-386
+          to: kubectl-passman
+
+    - selector:
+        matchLabels:
+          os: freebsd
+          arch: 386
+      uri: https://github.com/chrisns/kubectl-passman/releases/download/v0.0.1/kubectl-passman-freebsd-386.zip
+      sha256: "09484063fc7dc69b291d8a318c73fb85c66eea741cc489cfbf85b93544594f7e"
+      bin: "./kubectl-passman"
+      files:
+        - from: kubectl-passman-freebsd-386
+          to: kubectl-passman
+
+    - selector:
+        matchLabels:
+          os: freebsd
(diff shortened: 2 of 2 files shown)


plugins/passman.yaml
```diff
@@ -0,0 +1,90 @@
...
+          os: windows
+          arch: 386
+      uri: https://github.com/chrisns/kubectl-passman/releases/download/v0.0.1/kubectl-passman-windows-386.zip
+      sha256: "5adaa33ab1ed6fbaa063bb56e937c7ec0d3e7c93e8a5a42985cc4c84bb912320"
+      bin: "./kubectl-passman.exe"
+      files:
+        - from: kubectl-passman-windows-386.exe
+          to: kubectl-passman.exe
+
+  shortDescription: kubectl plugin that aspires to provide the missing link/glue between common password managers and kubectl
```
ahmetb: this is too long to be a shortDescription.

consider something like: `Store kubeconfig secrets in keychains or password managers`

plugins/passman.yaml
```diff
@@ -0,0 +1,90 @@
...
+  homepage: https://github.com/chrisns/kubectl-passman
+  caveats: |
+    This plugin needs a usable keychain or password manager, see the docs https://github.com/chrisns/kubectl-passman
```
ahmetb: consider line wrapping at ~80 chars

plugins/passman.yaml
```diff
@@ -0,0 +1,90 @@
...
+  description: |
+    An easy way to store your kubernetes credentials in 1password or Mac OS Keychain (more password managers coming soon)
```
ahmetb: consider line wrapping at ~80 chars

chrisns:
oh wow, sorry didn't realise the spammy repercussions of me trying to add more metadata

ahmetb:
Thanks for your contribution.
I see you invested in release procedure with krew a bit. There's an automated Krew release action for github at https://github.com/rajatjindal/krew-plugin-release if that is interesting to you (it's very recently developed by a third party).

/lgtm
/assign @corneliusweig 
for a second look.

chrisns:
thanks @ahmetb for sighting that. my google foo needs some work!
Had looked around but didn't see that, will look to move to that or similar, my supply chain spider sense does get a little (a lot) paranoid about generating tokens and providing them to 3rd party tools though :)
Hopefully github acitons will at somepoint allow hashing the full package of remote actions ¯\_(ツ)_/¯

corneliusweig:
@chrisns I can understand your reservations about the token. I think the recommended approach is to have a dedicated GitHub account which has the necessary permissions for just this one repo.

plugins/passman.yaml
```diff
@@ -0,0 +1,92 @@
...
+          to: kubectl-passman.exe
+
+  shortDescription: Store kubeconfig credentials in keychains or password managers
+  homepage: https://github.com/chrisns/kubectl-passman
+  caveats: |
+    This plugin needs a usable keychain or password manager
+    See usage docs https://github.com/chrisns/kubectl-passman
+  description: |
+    An easy way to store your kubernetes credentials in 1password or Mac OS Keychain
+    (more password managers coming soon)
```
corneliusweig: Can you integrate https://github.com/gopasspw/gopass soon :wink: :eyes:

plugins.md
```diff
@@ -42,6 +42,7 @@ Name | Description | Stars
 [oidc-login](https://github.com/int128/kubelogin) | Log in to the OpenID Connect provider | ![GitHub stars](https://img.shields.io/github/stars/int128/kubelogin.svg?label=stars&logo=github)
 [open-svc](https://github.com/superbrothers/kubectl-open-svc-plugin) | Open the Kubernetes URL(s) for the specified service in your browser. | ![GitHub stars](https://img.shields.io/github/stars/superbrothers/kubectl-open-svc-plugin.svg?label=stars&logo=github)
 [outdated](https://github.com/replicatedhq/outdated) | Finds outdated container images running in a cluster | ![GitHub stars](https://img.shields.io/github/stars/replicatedhq/outdated.svg?label=stars&logo=github)
+[passman](https://github.com/chrisns/kubectl-passman) | Provides glue between common password managers and kubectl for user credentials | ![GitHub stars](https://img.shields.io/github/stars/chrisns/kubectl-passman.svg?label=stars&logo=github)
```
corneliusweig: The description here is different from your manifest `shortDescription`. Although this is not a problem, you should know that this page is auto-generated and uses the `shortDescription` for the second column. So this variant is only temporary. (And your short description is good, because it gets truncated at ~50 characters)

plugins/passman.yaml
chrisns: 👍 added https://github.com/chrisns/kubectl-passman/issues/25 PRs welcome if you get to it before I do 😄

plugins.md
chrisns: This is going to be in active development for a bit, wanted it on krew to get some exposure and feedback. will fix the table on my next PR, thanks!

chrisns:
@corneliusweig yeah, I guess I'd need to refresh myself on github's auth scopes, and what the consequential blast radius would be of the permissions required. e.g. could the creds be used to PR other repos, for spam or malware.

Seperate accounts is a weird smell, even if I might then be able to claim an extra hacktoberfest t-shirt 😜  

For this particular plugin I'm tending to be especially nervous/respectful since people end up trusting it with access to their password managers, which in turn have frustratingly big scopes (e.g. from best I can tell 1password cli gives you a session token and anything with that can retrieve anything without alert/warning/prompt/logs 😱 )

rajatjindal:
Hi Chris/Ahmet/Cornelius

thanks for considering/suggesting the krew-plugin-release for this. 

I understand the reservation about providing token to be able to submit PR on any repo. 

can you think of a way that will help people have trust on this plugin? 

e.g. I had setup automated docker builds to avoid manual push, code is open source.

I will be more than willing to make changes if there is anything that will make this better in terms of security/UX.

Thanks
Rajat Jindal

rajatjindal:
how about, instead of asking each user to provide access to token, I can have a krew-plugin-release-bot account, that can open PR from its own repo instead of forking in your user's scope?

I think that should mitigate the concern?

chrisns:
hi @rajatjindal
Not really any 'good' ideas, it's a universal supply chain problem especially with github actions that'll hopefully be addressed.

I guess if it became owned by the krew team that would provide some legitimacy; I'd personally still be nervous.
I've not tried it, but if the `uses` could refer to a SHAs then at least I'd know it is as good as any due diligence I did at the time I installed it, however my ability to know about updates inc security ones would be lost unless the plugin 'phoned home' to check.

Your idea of PR'ing from an account you provide sounds like a good one, you'd need to invent some way to avoid clashing though. e.g. someone else intentionally or even accidently clashing on name and you publishing over my plugin name.

The alternative could be revisiting how krew manages its package registry, whereby instead of providing a manifest, I provide a link to where a manifest can be retrieved, and krew poll it every n minutes to then provide a (validated) static cache to the cli to query.

chrisns:
if you did use your own token, you'd need to somehow do it in such a way that it couldn't be extracted, which is already sounding complex and gnarly

chrisns:
> e.g. I had setup automated docker builds to avoid manual push, code is open source.

yes, though the registry that it pulls from doesn't distinguish if it was a manually created and pushed with potentially anything hidden in there, or automatically which would be harder to poison caches to achieve the same thing.
It'd be nice if the auto builder signed the image, though sadly I've not yet seen any (free) ones do that.

rajatjindal:
I was thinking more like the github action pushes an event to a downstream service (using GITHUB_TOKEN which is provided by default to all the actions), and then that downstream service actually opens PR using its token. 

I need to think more, but I totally understand your concern, and appreciate the discussion. I will try to find sometime this weekend to do a POC around this.

corneliusweig:
@rajatjindal @chrisns In general the bot like the one written by @rajatjindal is a great idea. Maintaining that in krew could be an option but we need to think this through.

Let's move this discussion to a `krew` issue instead of this PR.

PR #285, opened 2019-10-21
add new plugin remove-in-allns

rajatjindal:
Hi Krew Team

I would like to add a new plugin 'remove-in-allns' to the index. This plugin helps deletes a given resource from all the namespaces.

Thanks
Rajat Jindal

### DIFF
--- a/plugins/remove-in-allns.yaml
+++ b/plugins/remove-in-allns.yaml
@@ -1,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: remove-in-allns
+spec:
+  version: v0.0.2
+  homepage: https://github.com/rajatjindal/kubectl-remove-in-allns
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-remove-in-allns/releases/download/v0.0.2/darwin-amd64-v0.0.2.tar.gz
+    sha256: faf99884619da7d6455f500eb2993790e75c7af84aa6e08a138846db45a6fa23
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-remove-in-allns
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-remove-in-allns/releases/download/v0.0.2/linux-amd64-v0.0.2.tar.gz
+    sha256: 630be54a643fbceb7acfb6caf8794679f1f2a04e3146d440bcde0c21ec37b1b7
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-remove-in-allns
+  shortDescription: Remove the requested resource from all the namespaces.
+  caveats: |
+    This is potentially disruptive operation. Not recommended for use in Production and use at your own risk.
+
+    supports following resources right now (PR welcome for supporting other resource types):
+
+    - configmaps
+    - secrets
+    - ingress
+    - deployments
+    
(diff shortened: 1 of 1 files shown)


plugins/remove-in-allns.yaml
```diff
@@ -0,0 +1,55 @@
...
+    sha256: 630be54a643fbceb7acfb6caf8794679f1f2a04e3146d440bcde0c21ec37b1b7
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-remove-in-allns
+  shortDescription: Remove the requested resource from all the namespaces.
+  caveats: |
+    This is potentially disruptive operation. Not recommended for use in Production and use at your own risk.
+
+    supports following resources right now (PR welcome for supporting other resource types):
```
ahmetb: given this caveat I'm questioning the usefulness of this.

it's often an anti-pattern to remove something from multiple namespaces given the namespaces can be a prod/staging/dev boundary.

plugins/remove-in-allns.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: remove-in-allns
```
ahmetb: this is not a great name IMO. doesn't quite explain what's going on.

ahmetb:
@rajatjindal thanks for all these new plugins. 

I will make a personal recommendation: Try spending more time on each plugin an improving its quality, instead of creating more plugins without user input.

It is very likely that there will be a time you will not be able to keep these plugins up-to-date, and they will be malfunctioning (if not damaging to users). At that point, we'll remove them.

That's why I will recommend you to curate each plugin for a longer time instead of producing a new plugin before the previous plugin is not tested by at least a few users and works for them properly.

rajatjindal:
> @rajatjindal thanks for all these new plugins.
> 
> I will make a personal recommendation: Try spending more time on each plugin an improving its quality, instead of creating more plugins without user input.
> 
> It is very likely that there will be a time you will not be able to keep these plugins up-to-date, and they will be malfunctioning (if not damaging to users). At that point, we'll remove them.
> 
> That's why I will recommend you to curate each plugin for a longer time instead of producing a new plugin before the previous plugin is not tested by at least a few users and works for them properly.

I appreciate your recommendation, and following this now. I've stopped working on new plugins (for now :) ) , and focusing on improving the quality of these existing plugins as well as some other personal projects.

I am absolutely looking forward for user feedback, and have updated the comments on the tickets where these plugins were originally requested. 

many thanks again for the recommendation.

Thanks
Rajat Jindal

plugins/remove-in-allns.yaml
rajatjindal: I had thought about these names:

- remove-resource-everywhere
- remove-resource-allnamespaces

what do you think about one of these. Also if you have anything specific in mind, would be great to hear that.

plugins/remove-in-allns.yaml
rajatjindal: I wrote this as an enduser requested on kubernetes repo. I can understand if we don't want to include this in krew plugin list.

corneliusweig:
To me, removing across all namespaces is an anti-pattern. I guess there will be users out there who have such a workflow, but it's not something I would endorse.

Maybe such plugins are better suited for private collections, which `krew` does not yet support though.

So let's put this plugin on hold for now and wait for more evidence via user feedback.

/hold

rajatjindal:
Hi Cornelius/Ahmet

Thanks for your feedback. I can understand why this is an anti-pattern. 

I will close the PR for now, and will keep the plugin on my github account, incase someone needs it.

Thanks
Rajat Jindal

PR #287, opened 2019-10-22
Update view-secrets references

elsesiy:
Hi,
as announced to @ahmetb [here](https://github.com/ahmetb/kubectl-extras/pull/12), I re-wrote the `view-secrets` plugin.

This PR changes the existing references to the new repo.

### DIFF
--- a/plugins/view-secret.yaml
+++ b/plugins/view-secret.yaml
@@ -3,22 +3,24 @@ kind: Plugin
 metadata:
   name: view-secret
 spec:
-  version: "v0.1.0"
+  version: "v0.2.0"
   homepage: https://github.com/elsesiy/kubectl-view-secret
   shortDescription: |
     Decode Kubernetes secrets
   description: |+2
-    Base64 decode all key/value pairs in a given secret.
+    Base64 decode by key or all key/value pairs in a given secret.
 
-    $ kubectl view-secret <secret>
-    $ kubectl view-secret <secret> --namespace foo
+    $ kubectl view-secret <secret> # print secret keys
+    $ kubectl view-secret <secret> <key> # decode secret by key
+    $ kubectl view-secret <secret> -a/--all # decode all secret contents
+    $ kubectl view-secret <secret> -n/--namespace foo # print keys for secret in different namespace
   platforms:
     - selector:
         matchLabels:
           os: darwin
           arch: amd64
-      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.1.0/kubectl-view-secret_0.1.0_Darwin_x86_64.tar.gz
-      sha256: "35d6922031ac89726880218cd5b25a682ddc3a23f877b937d750a8f92707e29c"
+      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.2.0/kubectl-view-secret_0.2.0_Darwin_x86_64.tar.gz
+      sha256: "46d5fa2dadbdf13d1a470263fe75a4c6e4b061cfd383ab4642e8786206334f3d"
       files:
         - from: kubectl-view-secret
           to: .
@@ -27,8 +29,8 @@ spec:
         matchLabels:
           os: linux
           arch: amd64
-      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.1.0/kubectl-view-secret_0.1.0_Linux_x86_64.tar.gz
-      sha256: "67f6f13b437aa7381423001743c30ee43a939a547e2b41f4abd6377d3139edf2"
+      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.2.0/kubectl-view-secret_0.2.0_Linux_x86_64.tar.gz
+      sha256: "ac24776dd1170b78280445cac7562d3d4f5c984cedb613daef8cb1b0b934f442"
       files:
         - from: kubectl-view-secret
           to: .
@@ -37,8 +39,8 @@ spec:
         matchLabels:
           os: windows
           arch: amd64
-      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.1.0/kubectl-view-secret_0.1.0_Windows_x86_64.tar.gz
-      sha256: "80201481c07747878cb449ad868adcef8b4c3fb75e91e455551cc74b6d868daa"
+      uri: https://github.com/elsesiy/kubectl-view-secret/releases/download/v0.2.0/kubectl-view-secret_0.2.0_Windows_x86_64.tar.gz
+      sha256: "bac09b27a02738892d90199a2a3659d7fed5b6f43ac6f8dcd06e2ab9415293fb"
       files:
         - from: kubectl-view-secret.exe
           to: .
--- a/plugins/view-secret.yaml
+++ b/plugins/view-secret.yaml
@@ -3,19 +3,43 @@ kind: Plugin
 metadata:
   name: view-secret
 spec:
+  version: "v0.1.0"
+  homepage: https://github.com/elsesiy/kubectl-view-secret
+  shortDescription: |
+    Decode Kubernetes secrets
+  description: |+2
+    Base64 decode all key/value pairs in a given secret.
+
+    $ kubectl view-secret <secret>
+    $ kubectl view-secret <secret> --namespace foo
   platforms:
-  - sha256: 208fde0b9f42ef71f79864b1ce594a70832c47dc5426e10ca73bf02e54d499d0
-    uri: https://github.com/ahmetb/kubectl-extras/archive/b16b7c43a28240ed280c03b20e6a2f232f992479.zip
-    bin: view-secret.sh
-    files:
-    - from: "/*/view-secret/*"
-      to: "."
-    selector:
-      matchExpressions:
-      - {key: os, operator: In, values: [darwin, linux]}
-  version: "v0.0.0"
-  homepage: https://github.com/ahmetb/kubectl-extras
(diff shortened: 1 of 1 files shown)


plugins.md
```diff
@@ -62,7 +62,7 @@ Name | Description | Stars
 [sudo](https://github.com/postfinance/kubectl-sudo) | Run Kubernetes commands impersonated as group system:masters | ![GitHub stars](https://img.shields.io/github/stars/postfinance/kubectl-sudo.svg?label=stars&logo=github)
 [support-bundle](https://github.com/replicatedhq/troubleshoot) | Creates support bundles for off-cluster analysis | ![GitHub stars](https://img.shields.io/github/stars/replicatedhq/troubleshoot.svg?label=stars&logo=github)
 [tail](https://github.com/boz/kail) | Stream logs from multiple pods and containers using simple, dynamic source selection. | ![GitHub stars](https://img.shields.io/github/stars/boz/kail.svg?label=stars&logo=github)
-[view-secret](https://github.com/ahmetb/kubectl-extras) | Decode secrets | ![GitHub stars](https://img.shields.io/github/stars/ahmetb/kubectl-extras.svg?label=stars&logo=github)
+[view-secret](https://github.com/elsesiy/kubectl-view-secret) | Decode secrets | ![GitHub stars](https://img.shields.io/github/stars/elsesiy/kubectl-view-secret.svg?label=stars&logo=github)
```
ahmetb: This file is auto generated. You don’t need to edit. Make sure your manifest has the homepage set correctly is all.

plugins/view-secret.yaml
```diff
@@ -3,19 +3,43 @@ kind: Plugin
 metadata:
   name: view-secret
 spec:
+  version: "v0.1.0"
+  homepage: https://github.com/elsesiy/kubectl-view-secret
+  shortDescription: |
+    Kubernetes CLI plugin to decode Kubernetes secrets
```
ahmetb: Everything is a plugin in this repo & on krew search output. :)

plugins/view-secret.yaml
```diff
@@ -3,19 +3,43 @@ kind: Plugin
...
+  description: |+2
+    Let's you conveniently base64 decode all key/value pairs in a given secret.
```
ahmetb: You can delete up to “base64...“

plugins/view-secret.yaml
```diff
@@ -3,19 +3,43 @@ kind: Plugin
...
+
+    $ kubectl view-secret <secret>
```
ahmetb: Out of curiosity: do you support decoding certain data.* keys through a cli arg?

ahmetb:

-----

This pull request **doesn't** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/view-secret.yaml is not a straightforward version bump: diff line unrecognized for version bumps: [+  homepage: https://github.com/elsesiy/kubectl-view-secret]

plugins.md
elsesiy: Okay, I reverted this change

plugins/view-secret.yaml
elsesiy: Done!

plugins/view-secret.yaml
elsesiy: Done :)

plugins/view-secret.yaml
elsesiy: Currently not, I thought of adding this and just today someone actually asked for it so I probably ship it in the next version 👍

ahmetb:
+1 for version/syntax parity. We don't know how many users exist out there, but if we can get the same syntax (or an error that shows the new syntax gracefully) we should wait on flipping the plugin manifest (this PR).

elsesiy:
@corneliusweig @ahmetb My bad, I didn't even notice this because I didn't spend too much time trying to use the current version. In general I'm happy to add this but this somewhat changes how I anticipated using this plugin. Even though there might be instances where specifying the key makes sense, I see it more as an optional and not a mandatory attribute as this changes the workflow again to:
1. `kubectl view-secret <sec>`
2. Copy key
3. `kubectl view-secret <key from 2.>`

I wanted to achieve simplicity by removing the friction of having to specifically providing either `namespace` or `secret:key`. Should I rename this plugin and submit it as a separate plugin?

ahmetb:
We should keep the name the same, since I won't maintain view-secret anymore.

Is it not at all possible to accommodate something like:
- `kubectl view-secret <name>`
- `kubectl view-secret <name.subkey>`

at the same time? -n/--namespace should still exist separately.

elsesiy:
I think that works. Another approach could be to use `kubectl view-secret <name> --all`
and then preserve the rest of the functionality as described above.
Which one do you prefer?

ahmetb:
I don't follow, what does --all do?

elsesiy:
`--all/-a` would output all `key:value` pairs from the provided secret whereas `kubectl view-secret <name>` would just print out the keys/follows the existing flow. But I do like your suggestion with `secret.key` a lot so I'll probably go ahead with this one

ahmetb:
Why do you need —all? Existing plugin already prints all keys in a secret without needing any argument. Please try the existing plugin first. 🙃

elsesiy:
Yes I've seen the current plugin, `--all` however would print `keys=values` and not just `keys`. Anyway as I said, I'll go with the `name.key` approach!

ahmetb:
Yeah in case that makes sense. —all can print all secrets in the ns in k=v format.

corneliusweig:
So just to clarify to avoid misunderstandings about the semantics of `--all`:
@elsesiy you are going to reproduce the behavior of `view-secrets` where it will print a list of keys in a given secret, when called as `kubectl view-secret my-secret`. When called as `kubectl view-secret my-secret my-key`, it will decode the content under `my-key` in `my-secret`. That's the behavior of the old implementation.
In addition, you'll support a flag `--all`, which is used as `kubectl view-secret my-secret --all`. This will print all keys and decoded values inside a given secret.

Is that correct?

elsesiy:
@corneliusweig yes that's exactly how I implemented it now to ensure backward-compatibility but also allow for easy decoding of all contents of a secret. I'll publish it as `v0.2.0` and update this PR soon!

corneliusweig:
@elsesiy If you are revisiting your plugin, also consider if you really need the "k8s.io/cli-runtime/pkg/genericclioptions" and "k8s.io/utils/pointer" dependencies. You are not really making use of those, because you are shelling out to kubectl and the flags don't get passed along. I think you can make your binaries smaller by removing those deps (8MB compressed is quite hefty for what you are doing).

elsesiy:
@corneliusweig Yeah I noticed that too and removed these dependencies too, thanks!!

ahmetb:
/hold
Please let us know when you want to move forward with the new plugin. I'm putting this on hold since it seems like you wanna make some changes.

ahmetb:

-----

This pull request **doesn't** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/view-secret.yaml is not a straightforward version bump: diff line unrecognized for version bumps: [+  homepage: https://github.com/elsesiy/kubectl-view-secret]

elsesiy:
@corneliusweig @ahmetb I just pushed the changes as discussed, let me know if you have any remarks :)

plugins/view-secret.yaml
```diff
@@ -3,19 +3,45 @@ kind: Plugin
...
   name: view-secret
 spec:
+  version: "v0.2.0"
+  homepage: https://github.com/elsesiy/kubectl-view-secret
+  shortDescription: |
+    Decode Kubernetes secrets
+  description: |+2
+    Base64 decode by key or all key/value pairs in a given secret.
+
+    $ kubectl view-secret <secret> # print secret keys
```
corneliusweig: nit: can you align the comments? That would read so much nicer...

ahmetb: > 2\. The output is still not quite the same.

I think this is ok.

elsesiy: Will do but I just noticed I'm also missing [this](https://github.com/ahmetb/kubectl-extras/blob/master/view-secret/view-secret.sh#L38) case here. Do you want me to add it as well?

ahmetb: I think it would be nice. Many Secrets will only have 1 key.

Do this:
1. print these communication messages to `stderr`, print parseable output (by scripts etc) to `stdout`
2. offer a `-q`/`--quiet` to turn off communication messages like ("I'm selecting the only key").

ahmetb:

-----

This pull request **doesn't** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/view-secret.yaml is not a straightforward version bump: diff line unrecognized for version bumps: [+  homepage: https://github.com/elsesiy/kubectl-view-secret]

plugins/view-secret.yaml
elsesiy: Done!

elsesiy:
@corneliusweig @ahmetb Alright, I changed the format to the original output format, log info messages to `stderr` and added the `-q/--quiet` flag. Please let me know if there's anything missing at this point!

corneliusweig:
@elsesiy great, this looks pretty good now! Thanks for bearing with us through this lengthy review round and for taking over maintenance of this plugin!

/approve
/lgtm

elsesiy:
@corneliusweig Thank you, it was a good experience and sorry for not paying too much attention to the backwards-compatibility from the beginning! Won't happen again :) I think the `plugins.md` needs to be refreshed again, right?

ahmetb:
I'll send a PR to re-generate that, you can do as well if you like.

PR #294, opened 2019-10-31
Add plugin view-allocations 0.5.0

davidB:
This plugin lists allocations (cpu, memory, gpu,... X requested, limit, allocatable,...).

[x] Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md
[x] Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...]

Thanks


davidB:
I signed it

plugins/view-allocations.yaml
```diff
@@ -0,0 +1,29 @@
...
+          os: linux
+          arch: amd64
+      uri: https://github.com/davidB/kubectl-view-allocations/releases/download/0.4.0/kubectl-view-allocations_0.4.0-x86_64-unknown-linux-gnu.tar.gz
+      sha256: "97466fc77bc985bffe72f9c09b5aa9d0f2af4b184d1a20f46cd7bdfae2048a9d"
+      files: [{ from: "*", to: "." }]
+      bin: "./kubectl-view-allocations" # path to the plugin executable after copying files above
+  shortDescription: Prints the environment variables.
+  homepage: https://github.com/davidB/kubectl-view-allocations # optional, url for the project homepage
+  description: |
+    This plugin lists allocations (cpu, memory, gpu,... X requested, limit, allocatable,...).
```
corneliusweig: I think it would be worth to briefly explain the difference between `kubectl top nodes` and your plugin.

Also, could you explain a little more what is the meaning of the columns in your output (should not be here but on the project page). It will help the adoption of your tool, if you elaborate on `Request`, `Limit` and how that connects to a given container resource request spec.

plugins/view-allocations.yaml
```diff
@@ -0,0 +1,29 @@
+#see https://github.com/kubernetes-sigs/krew/blob/master/docs/DEVELOPER_GUIDE.md
```
corneliusweig: Also this line shouldn't be there.

corneliusweig:
@davidB Also you should think about distributing statically linked binaries. I'm on a pretty normal Linux system and I get:
```
        linux-vdso.so.1 (0x00007fffb45bc000)
        libssl.so.1.1 => not found
        libcrypto.so.1.1 => not found
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f9dbf4e7000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f9dbf4dd000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f9dbf4ba000)
        libgcc_s.so.1 => /usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/libgcc_s.so.1 (0x00007f9dbf2a2000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f9dbf0d1000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f9dbef94000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f9dbfb25000)
```

plugins/view-allocations.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  version: "v0.4.0" # required, must be in semver format, prefixed with "v"
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/davidB/kubectl-view-allocations/releases/download/0.4.0/kubectl-view-allocations_0.4.0-x86_64-apple-darwin.tar.gz
+      sha256: "b37c19d2b372b4a120582d33c1299d827f0c6324fce6548cbff35cffec1f2da3"
+      files: [{ from: "*", to: "." }]
```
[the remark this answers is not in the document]
davidB: I got an error when I test krew without this line (I'll check, maybe I don't have the latest version of krew)

davidB:
Thanks for your useful the feedback, I'll fix them.

ahmetb:
Also please make sure you import the Rust equivalent of Go package:

```
import _ "k8s.io/client-go/plugin/pkg/client/auth"
```

so cloud k8s clusters work with this plugin. I assume you use api directly and not shell out to kubectl.

davidB:
@ahmetb Do you meet an issue ? I used the [kube-rs](https://crates.io/crates/kube) lib. The lib is young. About auth part, the lib works/tested with gcp, minikube and oidc token but with some limititations (eg: no update of the token stored into kubeconfig).

ahmetb:
```
thread 'main' panicked at 'failed to load kubeconfig: Error { inner:

Error loading kube config: Missing GOOGLE_APPLICATION_CREDENTIALS env }', src/libcore/result.rs:1084:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
[1]    95119 abort      kubectl view-allocations
```

Doesn't work well on GCP. :(
Ideally you shouldn't need $GOOGLE_APPLICATION_CREDENTIALS, kubectl doesn't. It just calls into `gcloud config config-helper --format=json` to get an access_token+expiration.

ahmetb:
I also don't think you should need static linking for sure, I'm ok with having dynamically linking plugin binaries, especially given the benefits of OS package manager updating those common libraries on the client machine frequently.

ahmetb:
/hold 

Until the auth issues are fixed. We can get the plugin in, but I don't think most users on cloud providers will be able to take advantage of it. I left a comment at https://github.com/clux/kube-rs/issues/84 on how client-go authenticates to GCP. Hope someone implements it.

davidB:
Please wait, before merge. I should release a new version next week(end) with more static-linked dependency, and maybe a workaround to manage expired token (until kube-rs include a cleaner solution, may I'll made a PR).
I'll update this PR after the release of my plugin.

davidB:
kubectl-view-allocations 0.5.0 was released, with fixes for reported issue

- openssl is staticly linked
- include a workaround to managed expired access-token (not tested of GKE)
- update README, description

feedbacks are welcome.

plugins/view-allocations.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: view-allocations
+spec:
+  version: "v0.5.0"
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/davidB/kubectl-view-allocations/releases/download/0.5.0/kubectl-view-allocations_0.5.0-x86_64-apple-darwin.tar.gz
```
corneliusweig: You should also install a license file. As you are defaulting to the copy-all rule, it will be sufficient to just include the `LICENSE` file in your plugin archive.

davidB:
Hi do not have access to any GKE cluster, but I guess my workaround should work, @ahmetb can give a quick try to this version please ?

@corneliusweig I include the LICENSE.txt for the next version (0.5.1), I'll release it this week after adding info to the README. Can you update the https://github.com/kubernetes-sigs/krew/blob/master/docs/DEVELOPER_GUIDE.md with the requirement about LICENSE file for coming developer ?

Thanks for your valuable feedback.

corneliusweig:
@davidB https://github.com/kubernetes-sigs/krew/pull/394 was merged yesterday. Would you see the need for more information around this?

Btw, you were just unlucky with regards to the license. We are in the process of making this a requirement for plugins.

davidB:
the file is named `LICENSE.txt` in my archives (and project) is it OK ?

corneliusweig:
> the file is named `LICENSE.txt` in my archives (and project) is it OK ?

Yes absolutely. Ensuring that the license is there is part of the review. As long as it is clear that this is a license file, you can pick whatever you like.

davidB:
I also update the README.md (description) on github.

corneliusweig:
Hi @davidB, login on GKE looks good now :+1:.

My recommendation would be to also include more help text in the command itself, i.e. when running `kubectl view-allocations --help`. Also, you can try to improve the `description` in the manifest. Currently, it's hard to understand. Let's re-iterate this with the next version.

/lgtm
/approve
/hold cancel

PR #296, opened 2019-11-01
Add github stars to KUDO plugin

runyontr:
Just cleaning up the plugin list page to have the github stars for KUDO


corneliusweig:
This file is auto-generated with `sigs.k8s.io/krew/cmd/generate-plugin-overview`. The tool uses the `homepage` field from the manifest to generate the github stars shield.

Thus, your manual edit would soon be overwritten when we re-generate the plugins list. I think it would be ok to add exceptions in this script, because this is tool is solely dedicated for this task. So would you care to open a PR in `krew` instead?

runyontr:
Closing in favor of https://github.com/kubernetes-sigs/krew/pull/363

PR #303, opened 2019-11-06
Bump Outdated to 0.3.2

marccampbell:


ahmetb:

-----

This pull request **doesn't** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/outdated.yaml is not a straightforward version bump: diff line unrecognized for version bumps: [-      or https://github.com/replicatedhq/outdated/blob/v0.3.1/doc/USAGE.md]

plugins/outdated.yaml
```diff
@@ -43,7 +43,7 @@ spec:
 
     For additional options:
       $ kubectl outdated --help
-      or https://github.com/replicatedhq/outdated/blob/v0.3.1/doc/USAGE.md
+      or https://github.com/replicatedhq/outdated/blob/v0.3.2/doc/USAGE.md
```
ahmetb: Can you consider not versioning this string?
If you do that, you’ll get these PRs merged automatically without human reviews in 5 minutes.

corneliusweig: I do have sympathy for that pattern, though. It ensures that the usage guide matches with the release. Otherwise one might find documentation for a version that is too new.

What about extending your bot to allow hunks in `description` and `shortDescription` if the change is just the version bump?

ahmetb: It won't be easy since these fields are multi-line, and we verify certain 1-line fields are updated.

We can remove all such field restrictions and check every line to see if replacing vOld&rarr;vNew + sha256: field is good enough.

But I think since this plugin is the only exception (currently), I'm not inclined to make that big change.

marccampbell: 👍  will change this.

marccampbell: @ahmetb updated to prevent this from changing in the _next_ release

PR #309, opened 2019-11-10
Add new plugins “ctx” and “ns”

ahmetb:
Adds kubectx and kubens (from https://github.com/ahmetb/kubectx/)
as two separate plugins named:

- ctx
- ns

### DIFF
--- a/plugins/ctx.yaml
+++ b/plugins/ctx.yaml
@@ -8,7 +8,7 @@ spec:
   version: v0.7.1
   description: |
     Also known as "kubectx", a utility to switch between context entries in
-    your kubeconfig file effectively.
+    your kubeconfig file efficiently.
   caveats: |
     If fzf is installed on your machine, you can interactively choose
     between the entries using the arrow keys, or by fuzzy searching
--- a/plugins/ctx.yaml
+++ b/plugins/ctx.yaml
@@ -4,7 +4,7 @@ metadata:
   name: ctx
 spec:
   homepage: https://github.com/ahmetb/kubectx
-  shortDescription: Switch between clusters in your kubeconfig
+  shortDescription: Switch between contexts in your kubeconfig
   version: v0.7.1
   description: |
     Also known as "kubectx", a utility to switch between context entries in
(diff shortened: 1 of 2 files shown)


plugins/ctx.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: ctx
+spec:
+  homepage: https://github.com/ahmetb/kubectx
+  shortDescription: Switch between clusters in your kubeconfig
+  version: v0.7.1
+  description: |
+    Also known as "kubectx", a utility to switch between context entries in
+    your kubeconfig file effectively.
+  caveats: |
+    If fzf is installed on your machine, you can interactively choose
```
corneliusweig: It would make sense to include a link to the fzf repo.

plugins/ctx.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ctx
+spec:
+  homepage: https://github.com/ahmetb/kubectx
+  shortDescription: Switch between clusters in your kubeconfig
```
corneliusweig: I would say
```suggestion
  shortDescription: Switch between contexts in your kubeconfig
```
because users may define different contexts for the same cluster.

---

Also, it would be nice to squeeze a reference to `kubectx` in there, so that existing `kubectx` users know that they get same tool here _plus_ automatic updates via krew. For example

```suggestion
  shortDescription: Switch between contexts (same as kubectx)
```

plugins/ctx.yaml
```diff
@@ -0,0 +1,31 @@
...
+  version: v0.7.1
+  description: |
+    Also known as "kubectx", a utility to switch between context entries in
+    your kubeconfig file effectively.
```
corneliusweig: extreme nit-picking:
```suggestion
    your kubeconfig file efficiently.
```

plugins/ctx.yaml
ahmetb: I think fzf is easily findable and makes the text longer, so we should leave it out. I'm similarly thinking all these plugins listing USAGE strings in  their caveats/description. Similarly not ideal.

PR #351, opened 2019-11-19
Kubectl-use plugin

kvaps:
Tiny plugin for easy switch between Kubernetes contexts and namespaces

### DIFF
--- a/plugins/use.yaml
+++ b/plugins/use.yaml
@@ -1,0 +1,39 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: use
+spec:
+  version: "v1.0.0"
+  homepage: https://github.com/kvaps/kubectl-use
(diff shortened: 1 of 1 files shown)


ahmetb:
This sounds a lot like `ctx`, `ns` and `change-ns` plugins that we already have. :) Due to their existing user base and popularity, I'm inclined to not having this in the centralized index, also due to the vague name "use" which goes against our naming recommendations.

corneliusweig:
I also think that it does not make sense to have this particular plugin in the index. As already said, there already are `ctx` and `ns` which are pretty widespread and also offer a nice interactive switcher.
Or where do you see a unique feature of `use` that the other options do not offer?

corneliusweig:
Hearing no unique selling points for this plugin over `ctx`/`ns`, let's stick with what we already have in krew-index. Thanks for submitting anyway!

/close

PR #354, opened 2019-11-21
Add v0.1.0 of kubectl-snap

jeefy:
Hiya!

This is the product of a fun conversation at Kubecon. :)

It's a simple plugin that we're saying is a chaos-engineering tool. I'll happily dive down any rabbit hole needed to push this through.

Tested it locally against kind and it was rather devastating.

### DIFF
--- a/plugins/snap.yaml
+++ b/plugins/snap.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: snap
+spec:
+  version: "v1.0.1"
+  homepage: https://github.com/honk-ci/kubectl-snap
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/honk-ci/kubectl-snap/archive/v1.0.1.tar.gz
+    sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
+    files:
+    - from: "*"
+      to: .
+    bin: kubectl-snap
+  shortDescription: A chaos test that restores balance to your cluster at the speed of a snap.
+  caveats: |
+    Documentation:
+
+      See https://github.com/honk-ci/kubectl-snap or
+
+      $ kubectl tail --help
+
+  description: |
+    This plugin will delete half of the pods in a namespace or the whole cluster. 
--- a/plugins/snap.yaml
+++ b/plugins/snap.yaml
@@ -17,7 +17,7 @@ spec:
       to: .
     - from: "LICENSE"
       to: .
-  shortDescription: Randomly delete half of the pods in a namespace or cluster for chaos testing
+  shortDescription: Delete half of the pods in a namespace or cluster
   caveats: |
     To reiterate: This will delete half of the pods within the designated space. 
 
--- a/plugins/snap.yaml
+++ b/plugins/snap.yaml
@@ -3,13 +3,13 @@ kind: Plugin
 metadata:
   name: snap
 spec:
-  version: "v1.0.1"
+  version: "v0.0.1"
   homepage: https://github.com/honk-ci/kubectl-snap
   platforms:
   - selector:
       matchExpressions:
       - {key: os, operator: In, values: [darwin, linux]}
-    uri: https://github.com/honk-ci/kubectl-snap/archive/v1.0.1.tar.gz
+    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1/kubectl-snap-0.1.tar.gz
     sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
     files:
     - from: "*"
--- a/plugins/snap.yaml
+++ b/plugins/snap.yaml
@@ -3,25 +3,23 @@ kind: Plugin
 metadata:
   name: snap
 spec:
-  version: "v0.0.1"
+  version: "v0.1.0"
   homepage: https://github.com/honk-ci/kubectl-snap
   platforms:
   - selector:
       matchExpressions:
       - {key: os, operator: In, values: [darwin, linux]}
-    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1/kubectl-snap-0.1.tar.gz
+    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1.0/kubectl-snap-0.1.0.tar.gz
     sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
-    files:
-    - from: "*"
-      to: .
     bin: kubectl-snap
-  shortDescription: A chaos test that restores balance to your cluster at the speed of a snap.
+  shortDescription: Randomly delete half of the pods in a namespace or cluster for chaos testing
   caveats: |
+    To reiterate: This will delete half of the pods within the designated space. Currently, if there are pods that have failed or completed, they may get deleted and you may lose debug/log data. 
     Documentation:
 
       See https://github.com/honk-ci/kubectl-snap or
 
-      $ kubectl tail --help
+      $ kubectl snap -h
 
   description: |
-    This plugin will delete half of the pods in a namespace or the whole cluster. 
+    This plugin will delete half of the pods in a namespace or the whole cluster.  Restore balance to your cluster at the speed of a snap.
--- a/plugins/snap.yaml
+++ b/plugins/snap.yaml
@@ -12,9 +12,18 @@ spec:
     uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1.0/kubectl-snap-0.1.0.tar.gz
     sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
     bin: kubectl-snap
+    files:
+    - from: "kubectl-snap"
+      to: .
+    - from: "LICENSE"
+      to: .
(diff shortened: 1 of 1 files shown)


plugins/snap.yaml
```diff
@@ -3,13 +3,13 @@ kind: Plugin
 metadata:
   name: snap
 spec:
-  version: "v1.0.1"
+  version: "v0.0.1"
```
ahmetb: pls consider v0.1.0 you can, per semver.org

plugins/snap.yaml
```diff
@@ -0,0 +1,27 @@
...
+  homepage: https://github.com/honk-ci/kubectl-snap
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1/kubectl-snap-0.1.tar.gz
+    sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
+    files:
+    - from: "*"
+      to: .
```
ahmetb: this is the _default_, so it's redundant.

plugins/snap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    bin: kubectl-snap
+  shortDescription: A chaos test that restores balance to your cluster at the speed of a snap.
```
ahmetb: Doesn't explain what the plugin that.
`description` field does a better job.

Let's just say `Randomly delete pods in namespace or cluster for chaos testing`

plugins/snap.yaml
```diff
@@ -0,0 +1,27 @@
...
+  caveats: |
+    Documentation:
+
+      See https://github.com/honk-ci/kubectl-snap or
+
+      $ kubectl tail --help
```
ahmetb: You don't need this whole caveats section I think. :)
Maybe express gotchas here.
And `tail` doesn't sound right :))

plugins/snap.yaml
jeefy: Fair fair ;) Updating.

plugins/snap.yaml
jeefy: lol ;) My copy-pasta game was pretty weak, thanks KubeCon

plugins/snap.yaml
```diff
@@ -3,25 +3,23 @@ kind: Plugin
...
       - {key: os, operator: In, values: [darwin, linux]}
-    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1/kubectl-snap-0.1.tar.gz
+    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1.0/kubectl-snap-0.1.0.tar.gz
     sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
-    files:
-    - from: "*"
-      to: .
     bin: kubectl-snap
-  shortDescription: A chaos test that restores balance to your cluster at the speed of a snap.
+  shortDescription: Randomly delete half of the pods in a namespace or cluster for chaos testing
```
corneliusweig: This description shows up in `krew search`, but gets truncated at ~50chars. Can you make your description more succinct?

plugins/snap.yaml
```diff
@@ -3,25 +3,23 @@ kind: Plugin
...
   caveats: |
+    To reiterate: This will delete half of the pods within the designated space. Currently, if there are pods that have failed or completed, they may get deleted and you may lose debug/log data.
```
corneliusweig: Can you clarify how the pods are chosen? For example, does it consider each deployment independently and delete half (rounding?) of the pods in each? Or does it take the set of all pods and randomly delete from those. In the latter case, it may happen that _all_ pods of some deployment get deleted.

In any case, it would be interesting to know if at least one pod of each deployment is guaranteed to survive (again, I'm thinking about rounding errors).

plugins/snap.yaml
```diff
@@ -3,25 +3,23 @@ kind: Plugin
...
     Documentation:
 
       See https://github.com/honk-ci/kubectl-snap or
 
-      $ kubectl tail --help
+      $ kubectl snap -h
 
   description: |
-    This plugin will delete half of the pods in a namespace or the whole cluster. 
+    This plugin will delete half of the pods in a namespace or the whole cluster.  Restore balance to your cluster at the speed of a snap.
```
corneliusweig: > Restore balance to your cluster at the speed of a snap.

I guess that's a Star Wars reference? Although I appreciate this, I'm a bit worried that this joke will no longer work in January when the hype is over. If you want to use flowery language, something like this may be more apt:

> Wreak havoc unto your cluster!

plugins/snap.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  version: "v0.1.0"
+  homepage: https://github.com/honk-ci/kubectl-snap
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/honk-ci/kubectl-snap/releases/download/0.1.0/kubectl-snap-0.1.0.tar.gz
+    sha256: "8df10f76e82f71b34c737d128b7b369e0daf641118577822872537e787a6755d"
+    bin: kubectl-snap
```
corneliusweig: You archive lists the following files:
```
.gitignore
LICENSE
README.md
hack/
hack/release.sh
kubectl-snap
```
Some of these files don't need to be installed on the users' machine. Can you add a `files` section to just install `LICENSE` and `kubectl-snap`?

plugins/snap.yaml
```diff
@@ -0,0 +1,25 @@
...
+  shortDescription: Randomly delete half of the pods in a namespace or cluster for chaos testing
+  caveats: |
+    To reiterate: This will delete half of the pods within the designated space. Currently, if there are pods that have failed or completed, they may get deleted and you may lose debug/log data.
```
corneliusweig: Can you wrap lines in `caveats` and `description` at 80 chars, so it becomes readable on smaller terminals?

plugins/snap.yaml
ahmetb: > Can you clarify how the pods are chosen? For example, does it consider each deployment independently and delete half (rounding?) of the pods in each? Or does it take the set of all pods and randomly delete from those. In the latter case, it may happen that _all_ pods of some deployment get deleted.

I think this is not too relevant to here.
If users are curious they can read the repo.

ahmetb:
> What do you think about `zap` instead? Or even `snap-zap`?

I think it's worse. :)

corneliusweig:
Hey @epk, I see that you do not agree with my comments. Can you elaborate how exactly you differ?

plugins/snap.yaml
corneliusweig: Well it does make a difference if everything is still working well if at least one pod for each deployment remains. If all pods in a deployment get deleted, this means at least some 503 errors for this service, right?

plugins/snap.yaml
jeefy: Avengers actually. A year after Infinity War came out and the Snap is still definitely a relevant thing. 

Short version: In the movie, the villain is attempting to restore balance to the galaxy by preventing over-population. To that end, he attempts to gather stones of power, add it to a super powerful gauntlet, and **snap** his fingers.

plugins/snap.yaml
jeefy: Absolutely! Didn't know that limitation.

plugins/snap.yaml
jeefy: That is a fair point, but I also don't think that merging this in should be blocked by that. The actual goal of the plugin is to delete, at random, half of all pods in a namespace (or all cluster)

I think a flag that would leave at least one pod in a deployment is a decent addition for a 0.2.0 release, even if it somewhat counters the spirit of the **snap**

plugins/snap.yaml
jeefy: Absolutely!

plugins/snap.yaml
jeefy: Will do.

plugins/snap.yaml
jeefy: Done!

plugins/snap.yaml
jeefy: I updated the description to I did update the description to explicitly call out the possibility. I won't resolve this conversation in case there's more to be said.

jeefy:
Addressed most comments. Happy to discuss further if needed. :)

ahmetb:
I think the deletion respects graceful terminations etc. So the delete operation takes a very long time, as opposed to happening in a snap of a finger.

If you could use the evict API (we have a plugin for that too) it could be more interesting. Because right now it will take ~3-4 mins for half of my pods to delete, but most deleted pods will come back in ~2-3 secs. So there's almost no chaos in the cluster with this at all. :)

It's up to you how you want to fix it. We won't enforce how the plugin works. :)

jeefy:
I'd opt for merging and iterating. This was me taking a conversation/joke
with some folks at Kubecon and turning it into a reality in a couple hours.
I plan on making it more useful and applicable in the near future.

plugins/snap.yaml
corneliusweig: :laughing: I see! Then that's a pretty cool name for this plugin!

plugins/snap.yaml
corneliusweig: That's cool. As long as users are warned, that's fine.

corneliusweig:
/lgtm
/approve
/hold cancel
Thanks for bearing through this discussion!

PR #364, opened 2019-11-28
add df-pv plugin

yashbhutwala:
Signed-off-by: Yash Bhutwala <ymb002@bucknell.edu>

Add a plugin to see `df` for Persistent Volumes, [details here](https://github.com/yashbhutwala/kubectl-df-pv/blob/v0.1.0/README.md)

### DIFF
--- a/plugins/df-pv.yml
+++ b/plugins/df-pv.yml
@@ -1,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: df-pv
+spec:
+  version: "v0.1.0"
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/yashbhutwala/kubectl-df-pv/releases/download/v0.1.0/df-pv_0.1.0_linux_amd64-0.1.0.tar.gz
+    sha256: "eff764fb1d2869599f131ede6af3b64db4c6126ad711ba03868c2d218da4fbb3"
+    files:
+    - from: "df-pv"
+      to: "."
+    - from: "LICENSE"
+      to: "."
+    bin: "df-pv"
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/yashbhutwala/kubectl-df-pv/releases/download/v0.1.0/df-pv_0.1.0_darwin_amd64-0.1.0.tar.gz
+    sha256: "2f85948bba1cc16a7f1383a57a3afa083010c97ce2678a2a893319aa75bf5bc6"
+    files:
+    - from: "df-pv"
+      to: "."
+    - from: "LICENSE"
+      to: "."
+    bin: "df-pv"
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/yashbhutwala/kubectl-df-pv/releases/download/v0.1.0/df-pv_0.1.0_windows_amd64-0.1.0.zip
+    sha256: "d27f9407505917282d09ce0e720d7ac14d6c5d940e5e6c68463203cad19565da"
+    files:
+    - from: "df-pv.exe"
+      to: "."
+    - from: "LICENSE"
+      to: "."
+    bin: "df-pv.exe"
+  shortDescription: df (disk free) for persistent volumes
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for your PR. 
Have you considered pv-df name? Just sounded more natural but I am ok with both.

yashbhutwala:
@ahmetb I don't mind either way, but since I've already named it `df-pv` all over the plugin repo, let's just stick with it 😄

plugins/df-pv.yml
```diff
@@ -0,0 +1,55 @@
...
+  caveats: |
+    Usage:
+      $ kubectl df-pv
+
+    For additional options:
+      $ kubectl df-pv --help
+      or https://github.com/yashbhutwala/kubectl-df-pv/blob/v0.1.0/doc/USAGE.md
+
+  description: |
+    This plugin emulates df for persistent volumes.  This utility is meant for `cluster-admin` like user; specifically, you need a service account with enough RBAC privileges to access `api/v1/nodes/` from the `api-server`.  The output is very similar to df on Unix.
```
corneliusweig: Can you wrap this at ~80 chars, so that it prints nicely on smaller terminals?

corneliusweig:
Weird thought: what about renaming the plugin to just `df`? Or is that too generic?

ahmetb:
Df is too generic. What if df-node comes up?

@yashbhutwala you don’t have to rename the repo. You can rename the repo later but you can’t rename a plugin once it’s accepted. So I recommend choosing wisely.

yashbhutwala:
@ahmetb good point, branding is important I guess 😄.  @corneliusweig interesting suggestion 😄 ... I wouldn't be opposed to making this a generic `df` plugin with flags like `--nodes`, `--pv`, `--all-volumes`, `--images`.  What do you guys think?

ahmetb:
> I wouldn't be opposed to making this a generic `df` plugin

We've seen in "kubectl debug" example that it's advisable you develop small plugins, which may be someday turned into an official command like "df" you suggested. But until then, you should write small plugins for each, and if you successfully get it all right, you can maybe propose a builtin command or a "df" plugin.

ahmetb:
Let us know what your last call wrt naming is. df-pv is ok with me, too. And we can have df-* other plugins later on as you described.

corneliusweig:
Good point. Then let's stick with more specific plugin names for now.

plugins/df-pv.yml
yashbhutwala: Done 😃

yashbhutwala:
@ahmetb let's go with `df-pv`, mainly cause I like the suggestion of making multiple `df-*` 🙃

plugins/df-pv.yml
```diff
@@ -0,0 +1,55 @@
...
+  shortDescription: df (disk free) for persistent volumes
+  homepage: https://github.com/yashbhutwala/kubectl-df-pv
+  caveats: |
+    Usage:
+      $ kubectl df-pv
+
+    For additional options:
+      $ kubectl df-pv --help
+      or https://github.com/yashbhutwala/kubectl-df-pv/blob/v0.1.0/doc/USAGE.md
+
```
ahmetb: can you eliminate this entire section?
it's not really listing any caveats.

plugins/df-pv.yml
```diff
@@ -0,0 +1,55 @@
...
+        arch: amd64
+    uri: https://github.com/yashbhutwala/kubectl-df-pv/releases/download/v0.1.0/df-pv_0.1.0_windows_amd64-0.1.0.zip
+    sha256: "d27f9407505917282d09ce0e720d7ac14d6c5d940e5e6c68463203cad19565da"
+    files:
+    - from: "df-pv.exe"
+      to: "."
+    - from: "LICENSE"
+      to: "."
+    bin: "df-pv.exe"
+  shortDescription: df (disk free) for persistent volumes
```
ahmetb: try this

Show usage (like unix df) for persistent volumes

ahmetb: Otherwise you're just saying what's in the plugin name (literally) "df" and "pv". But this explains what the plugin does better.

plugins/df-pv.yml
yashbhutwala: Done 🌮

plugins/df-pv.yml
yashbhutwala: Done 🙃

PR #381, opened 2019-12-04
Update plugins list

corneliusweig:


corneliusweig:
I would have expected your bot to approve. Why didn't it work?

PR #386, opened 2019-12-12
Cluster snapshot plugin

fbrubbo:
https://github.com/fbrubbo/kubectl-plugins#kubectl-snapshot

### DIFF
--- a/plugins/snapshot.yaml
+++ b/plugins/snapshot.yaml
@@ -1,0 +1,84 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: snapshot
+spec:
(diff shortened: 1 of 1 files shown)


fbrubbo:
Done CLA

ahmetb:
I can't find the source code for your plugin. It seems like you just checked in a binary to GitHub, which is not a great idea at all. Can you please provide the source code?

corneliusweig:
I don't think we documented that plugins without source code cannot be accepted into `krew-index`.

corneliusweig:
@fbrubbo I'm sorry, but we will not accept opaque binary plugins whose source code cannot be inspected.

If you want to open-source your work, feel free to re-submit your plugin afterwards.
/close

PR #387, opened 2019-12-13
New plugin: capture

nestorsalceda:
This PR adds support for `kubectl-capture` plugin.

This plugin uses Sysdig under the hoods to help you troubleshooting issues or doing forensic analysis in a running pod or host.

You can check more references about its usage:

* https://github.com/sysdiglabs/kubectl-capture
* https://cloud.google.com/solutions/security-controls-and-forensic-analysis-for-GKE-apps#kubectl_sysdig_capture_sysdig_inspect
* https://sysdig.com/blog/tracing-in-kubernetes-kubectl-capture-plugin/

Thank you!

### DIFF
--- a/plugins/capture.yaml
+++ b/plugins/capture.yaml
@@ -1,0 +1,54 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capture
+spec:
+  version: v0.1.0
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/sysdiglabs/kubectl-capture/releases/download/v0.1.0/kubectl-capture.v0.1.0.tar.gz
+    sha256: "86c947a9fa9acb73eeb179e29ede20755a426a4fb96aa82375246a356f3a804f"
+    bin: "kubectl-capture"
+  shortDescription: >-
+    Triggers a Sysdig capture to troubleshoot the running pod
+  homepage: https://github.com/sysdiglabs/kubectl-capture
+  caveats: |
+    To analyze the capture file you will need to use Sysdig Inspect. You can use it as docker image directly:
+
+    ```bash
+    docker run -d -v /local/path/to/captures:/captures -p8080:3000 sysdig/sysdig-inspect:latest
+    ```
+
+    Usage:
+
+    ```bash
+    $ kubectl capture pod-name -M 45
+    ```
+
+    You can set the duration in seconds of the capture with the `-M` flag, and
+    also specify the namespace using the `-ns` or `--namespace` or use eBPF to
+    trace syscalls instead the kernel module instrumentation with `--ebpf` flag.
+  description: |
+    Sysdig is a powerful open source tool for container troubleshooting,
+    performance tunning and security investigation.
+
+    This plugin triggers a capture in the underlying host which is running a
+    pod. A capture file is created for a duration of time and is download
+    locally in order to use it with Sysdig Inspect.
+
+    Example:
+
+    ```bash
+    $ kubectl capture nginx-78f5d695bd-bcbd8
+    Sysdig is starting to capture system calls:
+
+    Node: gke-sysdig-work-default-pool-e35da3a1-m8vp
+    Pod: nginx-78f5d695bd-bcbd8
+    Duration: 30 seconds
+    Parameters for Sysdig: -S -M 30 -pk -z -w /capture-nginx-78f5d695bd-bcbd8-1550246926.scap.gz
+
+    The capture has been downloaded to your hard disk at:
+    ~/captures/capture-nginx-78f5d695bd-bcbd8-1550246926.scap.gz
+    ```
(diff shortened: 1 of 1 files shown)


plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    You can set the duration in seconds of the capture with the `-M` flag, and
+    also specify the namespace using the `-ns` or `--namespace` or use eBPF to
+    trace syscalls instead the kernel module instrumentation with `--ebpf` flag.
+  description: |
+    Sysdig is a powerful open source tool for container troubleshooting,
+    performance tunning and security investigation.
+
+    This plugin triggers a capture in the underlying host which is running a
+    pod. A capture file is created for a duration of time and is download
```
corneliusweig: nit
```suggestion
    pod. A capture file is created for a duration of time and is downloaded
```

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+    locally in order to use it with Sysdig Inspect.
```
corneliusweig: > ...order to use it with Sysdig Inspect.

So if I understand correctly, the plugin only works with an additional tool that needs to be downloaded separately. I don't think we have arrived at a final decision what we should do in such cases. I lean more towards allowing such plugins, in order to not limit creativity. @ahmetb what's your opinion here?

Just to make sure: is Sysdig Inspect opens source or freely available?

And can you include a link where to get the tool?

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    Example:
+
+    ```bash
```
corneliusweig: markdown markup doesn't really make sense here, because it's just dumped to the terminal.

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+    Usage:
+
+    ```bash
+    $ kubectl capture pod-name -M 45
+    ```
+
+    You can set the duration in seconds of the capture with the `-M` flag, and
+    also specify the namespace using the `-ns` or `--namespace` or use eBPF to
+    trace syscalls instead the kernel module instrumentation with `--ebpf` flag.
+  description: |
```
corneliusweig: Can you also include some information about the pre-requisites to use this plugin? For example, you need to be allowed/able to create privileged containers. Also, it seems to only work if the container runtime is docker? Also, what tools do you need to have installed and what kernel features need to be enabled to use the BPF capabilities?

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+    ```
+
+    Usage:
+
+    ```bash
+    $ kubectl capture pod-name -M 45
+    ```
+
+    You can set the duration in seconds of the capture with the `-M` flag, and
+    also specify the namespace using the `-ns` or `--namespace` or use eBPF to
```
corneliusweig: We try to align all plugin options with standard kubectl CLI options. So would it be possible for you to rename `-ns` to `-n`?

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/sysdiglabs/kubectl-capture/releases/download/v0.1.0/kubectl-capture.v0.1.0.tar.gz
+    sha256: "86c947a9fa9acb73eeb179e29ede20755a426a4fb96aa82375246a356f3a804f"
+    bin: "kubectl-capture"
+  shortDescription: >-
+    Triggers a Sysdig capture to troubleshoot the running pod
+  homepage: https://github.com/sysdiglabs/kubectl-capture
+  caveats: |
+    To analyze the capture file you will need to use Sysdig Inspect. You can use it as docker image directly:
```
corneliusweig: This caveats read more like a short command help. It would be better to move this into a place where the user expects it, like `kubectl capture --help`.

Instead, I would expect more true caveats, such as required pre-requisites to run this plugin. For example, you need to be able to run privileged pods.

plugins/capture.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    ```bash
```
corneliusweig: Again, markdown markup doesn't really make sense here.

ahmetb:
Background: We've discussed in-person about this at KubeCon. I think it's a good candidate. I only expressed concerns around the "generic" nature of the name.

Ideally, we should make progress on multi-index proposal soon, so that commercial entities like Sysdig can maintain their own version of krew-index, and distribute their commercial plugins without having to go through us (same goes for ReplicatedHQ folks). But until then, I'm good with this.

plugins/capture.yaml
nestorsalceda: Yes, that's it. Sysdig Inspect is Open Source. I will add the link.

Thank you!

plugins/capture.yaml
nestorsalceda: Sure, good point!

plugins/capture.yaml
nestorsalceda: I added the privileged pod requirement and explained the kernel stuff needed to get sysdig up and running.

nestorsalceda:
Hey @corneliusweig I updated the PR solving your concerns.

Thank you so much for reviewing this :)

plugins/capture.yaml
corneliusweig: Thanks, this is really helpful for first-time users!

corneliusweig:
Hey @nestorsalceda, this looks very nice! Thanks for contributing :champagne: 

/lgtm
/approve

PR #391, opened 2019-12-17
Print validated files to stderr

ahmetb:
Adding a pipe to tee /dev/stderr so that we can see which files are being
passed to validate-krew-manifest. Though, we print all files before validating
any of them, but still useful since mostly 1 plugin is updated at a time.


corneliusweig:
Makes sense in principle. Another approach would be to enable debug logging for manifest validation. This could give us also other valuable hints if anything goes wrong.

/lgtm
/approve
/hold

ahmetb:
I don't think we have much logs in manifest validator?

corneliusweig:
Hm.. somewhat true. I had a look and it's only 4 log statements that could be "unlocked".

PR #398, opened 2019-12-24
Add plugin: tree

ahmetb:
🌳

I think this plugin is a good example of how to automate release of a Go-based plugin.

- [GoReleaser config](https://github.com/ahmetb/kubectl-tree/blob/master/.goreleaser.yml):
  build Go binaries in tarballs and upload them to GitHub releases

- [GitHub GoReleaser workflow](https://github.com/ahmetb/kubectl-tree/blob/master/.github/workflows/release.yml):
  do the above task on new tags, doesn't require secrets setup

- [krew-release-bot](https://github.com/rajatjindal/krew-release-bot):
  automatically make PRs from new releases using [.krew.yaml template](https://github.com/ahmetb/kubectl-tree/blob/master/.krew.yaml).
  Though, it's currently not working for some reason.


corneliusweig:
/lgtm
/approve

Very apt of you to submit a 🎄 plugin on Christmas 🤶 🌨 ❄

This is really nice work, thanks for contributing this! Let's think about how we can advertise this as a best practices example, I think many people could benefit from it.

Btw, I think you could safely add a windows build as well. For me this has always worked very well for native Go plugins (however you need to do a jump through a hoop to get colors working properly).

ahmetb:
The special utf chars I use in the table aren’t available on Windows. I need to set simpler ascii alternatives for windows. I think colors will work thanks to fatih/colors pkg.

PR #401, opened 2019-12-26
add spy plugin

huazhihao:
kubespy is a shell-implemented non-invasive debugging plugin for kubectl

### DIFF
--- a/plugins/spy.yaml
+++ b/plugins/spy.yaml
@@ -1,0 +1,47 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: spy
+spec:
+  homepage: https://github.com/huazhihao/kubespy
+  shortDescription: pod debugging tool for kubernetes clusters with docker runtimes
+  version: v0.3.1
+  description: |
+    `kubespy` is a kubectl plugin implemented in bash to debug a running pod.
+    It starts a temporary `spy container` which joins the namespaces of the
+    target container (eg. pid/net/ipc). You can specify the image of `spy
+    container` which should include all the required debugging tools. Thus,
+    the debugging tools need not unnecessarily be bundled with the main
+    application container image.
+
+    `kubespy` is similar to [kubectl-debug](https://github.com/verb/kubectl-debug).
+    In contrast to the latter, kubespy works without the EphemeralContainers
+    feature which is an experimental alpha feature and needs to be activated
+    per pod.
+
+    Meanwhile `kubespy` has its prerequisites - the cluster must use docker as
+    container runtime and you need to be able to run privileged pods.
+
+    Usage:
+      kubectl spy POD [-c CONTAINER] [--spy-image SPY_IMAGE]
+
+  caveats: |
+    This plugin can only work with nodes with a docker runtime with admin
+    privileges and you need to be able to run privileged pods.
+
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/huazhihao/kubespy/releases/download/v0.3.1/kubespy.tar.gz
+    sha256: 904f24c189003b4fa4c2b93e902b1d8ee35b191b75c0276bf599a981eb0f168a
+    bin: kubespy
+    files:
+    - from: ./kubespy
+      to: kubespy
+    - from: ./LICENSE
+      to: .


huazhihao:
CLA signed already

ahmetb:
Thanks for this submission.
It looks a bit like the `debug` plugin we have, and the other [debug](https://github.com/aylei/kubectl-debug) that we don't have here. Can you quickly explain how it compares to those?

huazhihao:
> Thanks for this submission.
> It looks a bit like the `debug` plugin we have, and the other [debug](https://github.com/aylei/kubectl-debug) that we don't have here. Can you quickly explain how it compares to those?

Thank you for taking your time for the review.

Yes the basic idea is same. The major difference is this one is implemented in bash, combining existing kubectl commands instead of trying reinventing something. I guess the benefit is: 

- lightweight
- platform flexibility
- extensibility (you can edit in place)
- more fitting to unix philosophy?

ahmetb:
It's critical to us that users can easily understand the plugins and how they compare to each other. Based on your explanation, it's still not clear to me.

Do you mind spending time with those plugins to figure out how your plugin compares to others so you can come up with a better description? I think we'll be inquiring how it's unique before accepting it.

huazhihao:
Sure I will try to focus on these two: `kubespy` and  `verb/kubectl-debug`. For aylei/kubectl-debug, I will just brief it a little.

At a high level, they all intend to create a ephemeral container to debug a running pod, although the implementations are different in details.

`kubespy` creates a new pod to get control of the node with chroot firstly, then create the ephemeral container with docker cli. These are doable with existing kubectl commands.  So kubespy is implemented in bash.

`verb/kubectl-debug` which already enrolled in this index, calls the low level api to create the ephemeral container to join the pid/net/ipc namespace of target container. It is more efficient and safe, but requires 1.  the EphemeralContainers feature to be enabled in the cluster 2. shareProcessNamespace to be enabled for the target pod.

`aylei/kubectl-debug` has multiple modes. But something in common is they all requires delicate setup both on the client side and cluster side.

In conclusion:

`kubespy`: 

pros: implemented in bash, lightweight and extensible, compatible with all platforms, no cluster or pod prerequisites

cons: gets nodes exposed to potential vulnerability

`verb/kubectl-debug`:

pros: more efficient with low level api

cons: has cluster and pod prerequisites

If users bump into a random issue and don't want to destroy the current pod, I think `kubespy` has irreplaceable advantage. Otherwise `verb/kubectl-debug` can debug more securely.

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: spy
+spec:
+  homepage: https://github.com/huazhihao/kubespy
+  shortDescription: Non-invasive debugging tool for kubernetes
```
corneliusweig:
```suggestion
  shortDescription: Debug containers on docker runtimes
```

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: spy
```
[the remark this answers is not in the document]
ahmetb: In my opinion it's unique enough that people probably won't try to create another spy plugin. Thats said, the name tells what the plugin does only a little bit, so this will impact its discoverability big time.

huazhihao:
Thanks @ahmetb and @corneliusweig  so much for your reviews and improvement suggestions. I have updated the description with more accurate wording accordingly.

For the name selection of `spy`, I prefer this name because it is short and gives the impression that all actions will be non-invasive - that is the characteristic I try to differentiate with other debugging plugins. And I think it is also understandable for a debugging purpose.

plugins/spy.yaml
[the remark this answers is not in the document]
huazhihao: Thanks @ahmetb and @corneliusweig so much for your reviews and improvement suggestions. I have updated the description with more accurate wording accordingly.

For the name selection of spy, I prefer this name because it is short and gives the impression that all actions will be non-invasive - that is the characteristic I try to differentiate with other debugging plugins. And I think it is also understandable for a debugging purpose.

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
...
+    tools into a particular running pod for debugging. So you don't have to
+    modify the spec of the pod or the image of the container just for
+    debugging purpose.
+
+    Usage:
+      kubectl spy POD [-c CONTAINER] [--spy SPY_IMAGE]
+
+    See https://github.com/huazhihao/kubespy for more details.
+  caveats: |
+    This plugin will create a privileged pod, mounting the root directory of
```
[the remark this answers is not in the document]
huazhihao: Sure. Updated.

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: spy
+spec:
+  homepage: https://github.com/huazhihao/kubespy
+  shortDescription: pod debugging tool for kubernetes clusters with docker runtimes
```
corneliusweig: It's better to stay below ~50chars. Would that work too?
```suggestion
  shortDescription: Debug pods on clusters with docker runtime
```

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
...
+  version: v0.3.1
+  description: |
+    `kubespy` is a kubectl plugin implemented in bash to debug a application pod during runtime by creating and running an temporary `spy container` to join its docker namespace(eg. pid/net/ipc). You can specify the image of `spy container` which should include all the debug tools required and can load them into the debugging environment, so you don't have to unnecessarily bundle those debug tools with the application pod image.
```
corneliusweig:
```suggestion
    `kubespy` is a kubectl plugin implemented in bash to debug a running pod. It starts a temporary `spy container` which joins the namespaces of the target container (eg. pid/net/ipc). You can specify the image of `spy container` which should include all the required debugging tools. Thus, the debugging tools need not unnecessarily be bundled with the main application container image.
```

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    Unlike [kubectl-debug](https://github.com/verb/kubectl-debug), `kubespy` doesn't require the prerequisites of 1. `EphemeralContainers` to be enabled in the cluster 2. `shareProcessNamespace` to be enabled for the application pod. The reasons are 1. `EphemeralContainers` is still in early alpha state and is not suitable for production clusters 2. modifying the spec of `shareProcessNamespace` will destroy the original application pod as well as the evidences inside.
+
+    Meanwhile `kubespy` has its prerequisite - the node that hosting the application pod needs to run on a docker runtime with admin privileges.
```
corneliusweig:
```suggestion
    Meanwhile `kubespy` has its prerequisites - the cluster must use docker as container runtime and you need to be able to run privileged pods.
```

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: spy
+spec:
+  homepage: https://github.com/huazhihao/kubespy
+  shortDescription: pod debugging tool for kubernetes clusters with docker runtimes
+  version: v0.3.1
+  description: |
+    `kubespy` is a kubectl plugin implemented in bash to debug a application pod during runtime by creating and running an temporary `spy container` to join its docker namespace(eg. pid/net/ipc). You can specify the image of `spy container` which should include all the debug tools required and can load them into the debugging environment, so you don't have to unnecessarily bundle those debug tools with the application pod image.
+
+    Unlike [kubectl-debug](https://github.com/verb/kubectl-debug), `kubespy` doesn't require the prerequisites of 1. `EphemeralContainers` to be enabled in the cluster 2. `shareProcessNamespace` to be enabled for the application pod. The reasons are 1. `EphemeralContainers` is still in early alpha state and is not suitable for production clusters 2. modifying the spec of `shareProcessNamespace` will destroy the original application pod as well as the evidences inside.
```
corneliusweig: The explanation for 2 is not straightforward to understand, because you are talking about the situation where the config option was _not_ active when the error was encountered, and activating the option would re-create the pod. Maybe just skip the explanations altogether, WDYT?

You could also say something like

> `kubespy` is similar to [kubectl-debug](...). In contrast to the latter, kubespy works without the `EphemeralContainers` feature with is an experimental alpha feature and needs to be activated per pod.

plugins/spy.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: spy
+spec:
+  homepage: https://github.com/huazhihao/kubespy
+  shortDescription: pod debugging tool for kubernetes clusters with docker runtimes
+  version: v0.3.1
+  description: |
```
corneliusweig: Can you wrap lines at ~80 chars, so that it prints nicely on smaller terminals?

plugins/spy.yaml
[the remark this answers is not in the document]
corneliusweig: Alright, then let's stick with `spy`.

plugins/spy.yaml
huazhihao: Agreed. Maybe I will just put 2 in the readme file of the plugin repo.

huazhihao:
@corneliusweig I really appreciate your help on improving the description in terms of English and the technical part.

plugins/spy.yaml
corneliusweig: Somehow that line wrapping must have slipped through. It greatly improves readability on small terminals, so I recommend doing this.

corneliusweig:
> @corneliusweig I really appreciate your help on improving the description in terms of English and the technical part.

@huazhihao I was already beginning to fear that I'm too pesky :sweat_smile:

One last nit about line wrapping and I'll merge.

huazhihao:
> > @corneliusweig I really appreciate your help on improving the description in terms of English and the technical part.
> 
> @huazhihao I was already beginning to fear that I'm too pesky 😅
> 
> One last nit about line wrapping and I'll merge.

Ah I am terribly sorry. Cannot believe I missed that one.

corneliusweig:
:confetti_ball:
/lgtm
/approve

PR #406, opened 2019-12-30
Add Snapshot plugin

fbrubbo:
See more on https://github.com/fbrubbo/kubectl-snapshot


ahmetb:
Hi, thanks for your submission. 
I have a suspicion that this does a subset of “get-all” plugin. Also the name is not clear enough to distinguish between storage volumes snapshot vs API objects.

fernandorubbo:
It is not a get all kubernetes resources plugin. It pulls information about machine resources (cpu, memory, requests, limits) and also configuration such as PDB, probes, etc all at once in a unique line.For

ahmetb:
Yep, looks authentic. We have several other plugins (`view-allocations` and `resource-usage` that show workload resource usage and allocations on nodes) but this plugin differentiates enough to be here.

Thanks, we'll review.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+        arch: amd64
+    uri: https://github.com/fbrubbo/kubectl-snapshot/archive/v0.1.0.tar.gz
+    sha256: "369e4bbbe1afa76d5f363f7c123235ac556bd12989bc3fca286982f338086610"
+    files:
+    - from: "kubectl-snapshot-*/bin/kubectl-snapshot-linux-amd64"
+      to:  "."
+    bin: "./kubectl-snapshot-linux-amd64"
+  - selector:
+      matchLabels:
+        os: linux
```
ahmetb: you can just say `os: linux` and omit arch to match all arches.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+    uri: https://github.com/fbrubbo/kubectl-snapshot/archive/v0.1.0.tar.gz
+    sha256: "369e4bbbe1afa76d5f363f7c123235ac556bd12989bc3fca286982f338086610"
+    files:
+    - from: "kubectl-snapshot-*/bin/kubectl-snapshot-linux-386"
+      to:  "."
+    bin: "./kubectl-snapshot-linux-386"
+
+  - selector:
+      matchLabels:
+        os: darwin
```
ahmetb: actually it seems like this plugin works on all `linux` and `darwin` architectures, just do this: 

https://github.com/kubernetes-sigs/krew-index/blob/1adc818356eb63d643b0d544ff28e0805c51f34e/plugins/debug-shell.yaml#L13-L15

it seems like all platform specs seem identical.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+        arch: amd64
+    uri: https://github.com/fbrubbo/kubectl-snapshot/archive/v0.1.0.tar.gz
+    sha256: "369e4bbbe1afa76d5f363f7c123235ac556bd12989bc3fca286982f338086610"
+    files:
+    - from: "kubectl-snapshot-*/bin/kubectl-snapshot-darwin-amd64"
+      to:  "."
```
ahmetb: please also copy out LICENSE next to the executable.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+    sha256: "369e4bbbe1afa76d5f363f7c123235ac556bd12989bc3fca286982f338086610"
+    files:
+    - from: "kubectl-snapshot-*/bin/kubectl-snapshot-windows-386.exe"
+      to:  "."
+    bin: "./kubectl-snapshot-windows-386.exe"    
+
+  shortDescription: >-    
+    Takes a snapshot of your cluster and print it in either standard io or csv files
+  homepage: https://github.com/fbrubbo/kubectl-snapshot
+  caveats: |
```
ahmetb: this is really the `description:` section, as the content here are not caveats.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+        arch: 386
+    uri: https://github.com/fbrubbo/kubectl-snapshot/archive/v0.1.0.tar.gz
+    sha256: "369e4bbbe1afa76d5f363f7c123235ac556bd12989bc3fca286982f338086610"
+    files:
+    - from: "kubectl-snapshot-*/bin/kubectl-snapshot-windows-386.exe"
+      to:  "."
+    bin: "./kubectl-snapshot-windows-386.exe"    
+
+  shortDescription: >-    
+    Takes a snapshot of your cluster and print it in either standard io or csv files
```
ahmetb: Let's make it shorter:

> Print a snapshot of nodes, pods, HPAs resource usage

plugins/snapshot.yaml
ahmetb: ignore that, looks like you actually compile separate binaries for each platform.

plugins/snapshot.yaml
fernandorubbo: I've put in the caveats field because I've when I install it using krew, this seems what is being printed out

ahmetb:
/lgtm
I'll have @corneliusweig take another look and approve.

plugins/snapshot.yaml
```diff
@@ -0,0 +1,109 @@
...
+    in your cluster. 
+
+    Usage:
+      kubectl snapshot
+      kubectl snapshot -csv-output <NAME>
+
+    Other fine tunning are allowed, to understand more type:
+      kubectl snapshot -h
+
+  caveats: |
```
corneliusweig: IMHO, this information is still better suited for general documentation and not the caveats section. The caveats section is really meant for pre-requisites or necessary rights to run this plugin. For example, you should mention that it requires a metrics collection component (heapster) in the cluster.

fbrubbo: Agreed. But as I mention to ahmetb, the caveat section is the only one being printed when a user installs the plugin. To be honest, I couldn't figure out where the description field is being used. 
Anyways, I will follow your guidance and add everything in the description section.

fernandorubbo:
/assign corneliusweig
Thanks for the comments. I implemented most of them. 
About the docs: I did a few tweaks now and I have the intention to improve it even further later on  
About the output: yes, I agree. Actually I've already started to put some info in the csv only. I plan to improve this in a few weeks from now. The year started a bit busy.

fernandorubbo:
@ahmetb and @corneliusweig
Do you need anything else from me?

corneliusweig:
@fernandorubbo Thanks for pushing this through, and sorry for my slow response times :/

PR #428, opened 2020-01-13
adding kubectl duck to krew index.

n3wscott:
- Named duck, stands for ducktype, main work is spinning out of Knative.
- Verified I can install via `kubectl krew install --manifest=plugins/duck.yaml`

### DIFF
--- a/plugins/duck.yaml
+++ b/plugins/duck.yaml
@@ -5,9 +5,10 @@ metadata:
 spec:
   version: v0.0.1
   homepage: https://github.com/n3wscott/kubectl-duck
-  shortDescription: Ducktype support for kubectl.
+  shortDescription: List custom resources with ducktype support
   description: |
-    This plugin allows you to list and get custom resources based on ducktype labels on CRDs.
+    This plugin allows you to list and get custom resources based on ducktype
+    labels on CRDs.
   caveats: |
     * Ducktype labels are currenly hardcoded to known types from Knative.
   platforms:
@@ -19,7 +20,6 @@ spec:
     sha256: 86294517e15b77297eeeafcb515ed7dd5f1c0ea13f6bfa4e1ecdc9d8096ab009
     files:
     - from: "*"
-      to: "."
     bin: kubectl-duck
   - selector:
       matchLabels:
@@ -29,7 +29,6 @@ spec:
     sha256: 47cc0bce7b01ca4fcde89ea30d9dad862591477e0647e30eb3367606968069e0
     files:
     - from: "*"
-      to: "."
     bin: kubectl-duck
   - selector:
       matchLabels:
@@ -39,5 +38,4 @@ spec:
     sha256: e48aa627f39d338a3b86f7c08d45885b800df70b759a267cb120f11957cbfe9b
     files:
     - from: "*"
-      to: "."
     bin: kubectl-duck.exe
--- a/plugins/duck.yaml
+++ b/plugins/duck.yaml
@@ -18,8 +18,6 @@ spec:
         arch: amd64
     uri: https://github.com/n3wscott/kubectl-duck/releases/download/v0.0.1/kubectl-duck_v0.0.1_darwin_amd64.tar.gz
     sha256: 86294517e15b77297eeeafcb515ed7dd5f1c0ea13f6bfa4e1ecdc9d8096ab009
-    files:
-    - from: "*"
     bin: kubectl-duck
   - selector:
       matchLabels:
@@ -27,8 +25,6 @@ spec:
         arch: amd64
     uri: https://github.com/n3wscott/kubectl-duck/releases/download/v0.0.1/kubectl-duck_v0.0.1_linux_amd64.tar.gz
     sha256: 47cc0bce7b01ca4fcde89ea30d9dad862591477e0647e30eb3367606968069e0
-    files:
-    - from: "*"
     bin: kubectl-duck
   - selector:
       matchLabels:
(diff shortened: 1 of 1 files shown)


plugins/duck.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: duck
+spec:
+  version: v0.0.1
+  homepage: https://github.com/n3wscott/kubectl-duck
+  shortDescription: Ducktype support for kubectl.
```
ahmetb: Can you be more descriptive? Since it's a kubectl, it's already `for kubectl`, that part is redundant. I would probably say something like:

> List custom resources with ducktype support

plugins/duck.yaml
```diff
@@ -0,0 +1,43 @@
...
+  description: |
+    This plugin allows you to list and get custom resources based on ducktype labels on CRDs.
```
ahmetb: Please 80-char wrap this for terminals.

plugins/duck.yaml
```diff
@@ -0,0 +1,43 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/n3wscott/kubectl-duck/releases/download/v0.0.1/kubectl-duck_v0.0.1_darwin_amd64.tar.gz
+    sha256: 86294517e15b77297eeeafcb515ed7dd5f1c0ea13f6bfa4e1ecdc9d8096ab009
+    files:
+    - from: "*"
+      to: "."
```
ahmetb: this is the default, you can omit.

plugins/duck.yaml
```diff
@@ -19,7 +20,6 @@ spec:
     sha256: 86294517e15b77297eeeafcb515ed7dd5f1c0ea13f6bfa4e1ecdc9d8096ab009
     files:
     - from: "*"
```
ahmetb: you need to remove both from/to to take advantage of the default.

plugins/duck.yaml
n3wscott: OK, I will use this for now but I will be updating to also support things like podspecable which are things like Deployments and Services.sources.knative.dev in the next release, will update this then.

plugins/duck.yaml
n3wscott: sounds good! I am going off what kubectl-tree is doing.

n3wscott:
It seems to reinstall correctly with the changes removing `file:`

ahmetb:
/lgtm
/cc @corneliusweig 
for /approve

plugins/duck.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: duck
+spec:
+  version: v0.0.1
+  homepage: https://github.com/n3wscott/kubectl-duck
+  shortDescription: List custom resources with ducktype support
+  description: |
+    This plugin allows you to list and get custom resources based on ducktype
+    labels on CRDs.
+  caveats: |
+    * Ducktype labels are currenly hardcoded to known types from Knative.
```
corneliusweig: Is that leading star intentional?

plugins/duck.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: duck
+spec:
+  version: v0.0.1
+  homepage: https://github.com/n3wscott/kubectl-duck
+  shortDescription: List custom resources with ducktype support
+  description: |
+    This plugin allows you to list and get custom resources based on ducktype
```
corneliusweig: As far as I understand, this is purely a companion tool for Knative, right? If so, this should be mentioned in the description. This helps both knative users to recognize relevance and non-users to filter out.

I'm also not entirely sure, how prevalent the term `ducktype` is in knative. Does it need some explanation?

plugins/duck.yaml
```diff
@@ -0,0 +1,35 @@
...
+    labels on CRDs.
```
corneliusweig:
```suggestion
    labels on Knative CRDs.
```

ahmetb: I think it works on all CRDs (plus maybe core types if OP hardcodes them).

plugins/duck.yaml
ahmetb: I think it's independent of knative, though the concept of duck types are currently hosted in Knative codebase. That said, for example, a k8s Service can fit to "addressable" duck type, and so could some CRDs. So it's a generic concept, but not defined in k8s core. And it can still be used without Knative installed.

plugins/duck.yaml
n3wscott: I think it is a list. but has only one point.

n3wscott:
> In general, this looks quite useful, but the scope is limited to Knative users 

This is not limited to Knative. Anyone can implement a ducktype and the goal longer term is that you can bring your own ducktypes.

plugins/duck.yaml
n3wscott: Yes, it is not just for Knative. Knative currenly champions the ducktype concepts, but things like triggers in tekton are also addressable and if they label their CRDs correctly this tool will pick it up.

plugins/duck.yaml
n3wscott: it works on all CRDs. At the moment the list of label selectors is limited to ducktypes know by knative, but that will be opened up in future releases.

plugins/duck.yaml
ahmetb: that's fine I think.

corneliusweig:
> This is not limited to Knative

I see. Seems like I have to do some reading :wink:

PR #429, opened 2020-01-13
adding kubectl select to krew index.

n3wscott:
- select stands for `select a context for kubectl`.
- installed tested via ` kubectl krew install --manifest=plugins/select.yaml`

### DIFF
--- a/plugins/select.yaml
+++ b/plugins/select.yaml
@@ -1,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: select
+spec:
+  version: v0.0.3
+  homepage: https://github.com/n3wscott/kubectl-select
+  shortDescription: Select the active Kubernetes context using a menu.
+  description: |
+    This plugin allows you see all configured Kubernetes contexts as a human
+    friendly list and use your arrow keys to select the active context.
+  caveats: |
+    * Contexts must already be setup in the local context file.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/n3wscott/kubectl-select/releases/download/v0.0.3/kubectl-select_v0.0.3_darwin_amd64.tar.gz
+    sha256: acbd4dc29c4228e7cd392a305eb3d7155bc22c7d550172cc2a49779cdaa0f715
+    bin: kubectl-select
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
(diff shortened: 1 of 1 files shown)


ahmetb:
(My comment based on the description) Since the well-known kubectx tool is already a plugin here, named `ctx`, it's unlikely we'll go ahead with this. We have #351 as the precedent here (which we've closed, as well).

/hold
/cc @corneliusweig 
for additional review if I'm missing anything (if not, we can /close)

n3wscott:
Does ctx have a menu?? You should try it. 😜😜

ahmetb:
Yeah kubectx has an interactive menu with support for arrow keys and fuzzy string search. Further, it can rename contexts, delete contexts and switch back and forth using `-` syntax. You should try it. 😉😉

n3wscott:
OH SWEET!! ok I will try it. I hacked up kubectl-select several months ago when I could not find one, but if there is a better one, AWESOME.

n3wscott:
hm, how do I get the menu? 

```(╯°□°)╯︵  kubectl ctx
d2k
d2k-admin
docker-desktop
docker-for-desktop
mink
[01:33:06] snichols@snichols-a02 ~/go/src/github.com/kubernetes-sigs/krew-index (select)
(╯°□°)╯︵  kubectl ctx --help
USAGE:
  kubectl ctx                       : list the contexts
  kubectl ctx <NAME>                : switch to context <NAME>
  kubectl ctx -                     : switch to the previous context
  kubectl ctx -c, --current         : show the current context name
  kubectl ctx <NEW_NAME>=<NAME>     : rename context <NAME> to <NEW_NAME>
  kubectl ctx <NEW_NAME>=.          : rename current-context to <NEW_NAME>
  kubectl ctx -d <NAME> [<NAME...>] : delete context <NAME> ('.' for current-context)
                                  (this command won't delete the user/cluster entry
                                  that is used by the context)

  kubectl ctx -h,--help             : show this message

```

n3wscott:
when I get some time, maybe I can look into adding the menu to this version. I don't want two, happy to upstream...

ahmetb:
https://kubectx.dev#interactive-mode

corneliusweig:
As far as I can tell, `select` switches the context with a terminal UI.
`ctx` can already do that (albeit with `fzf` helper) so I see no reason why we should have two very similar options in `krew-index`. In fact, `ctx` has other big advantages:
- it has been honed by a large user-basis on all platforms
- it keeps record of the previous context for fast back-and-forth
- it's bash and therefore very lightweight

So let's not confuse our users and just stick to `ctx`. But thanks for submitting this anyway!

/close

PR #432, opened 2020-01-14
Add new plugin, `kubectl resources`

howardjohn:
Fixes https://github.com/howardjohn/kubectl-resources/issues/24

### DIFF
--- a/plugins/resources.yaml
+++ b/plugins/resources.yaml
@@ -1,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: resources
+spec:
+  version: "v0.1.0"
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+        arch: amd64
+      uri: https://github.com/howardjohn/kubectl-resources/releases/download/v0.1.0/kubectl-resources_0.1.0_Darwin_x86_64.tar.gz
+      sha256: f27cc3474df627d0d9b55eee6514a9d7c7858d1a878af75c07ad15b3a5272d75
+      bin: kubectl-resources
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/howardjohn/kubectl-resources/releases/download/v0.1.0/kubectl-resources_0.1.0_Linux_x86_64.tar.gz
+      sha256: ad3d1a126b106a38fb92ae76bb728b00ed8b4ef02eef574456b73f21c28affb3
+      bin: kubectl-resources
+  shortDescription: >-
+    Access Kubernetes resource requests, limits, and usage.
+  homepage: https://github.com/howardjohn/kubectl-resources
+  description: |
+    Access Kubernetes resource requests, limits, and usage.
--- a/plugins/resources.yaml
+++ b/plugins/resources.yaml
@@ -20,7 +20,8 @@ spec:
       sha256: ad3d1a126b106a38fb92ae76bb728b00ed8b4ef02eef574456b73f21c28affb3
       bin: kubectl-resources
   shortDescription: >-
-    Access Kubernetes resource requests, limits, and usage.
+    Overview of Kubernetes resource requests, limits, and usage.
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for submitting. Do you mind explaining how is it different than the several resource/allocation plugins we have?

ahmetb:
I guess it is similar to view-allocations, just curious if it’s entirely overlapping or not.

howardjohn:
As far as I know this is the only one aggregating resource requests/limits with usage. I realize it isn't groundbreakingly innovative, I made it since I find it useful and someone opened an issue recommending me to add it to krew

ahmetb:
For reference, I'm attaching screenshots:

- kubectl view-allocations: 
image

- kubectl resources (with "pods" section of the "view-allocations" in the same screen) 
image

- `kubectl view-allocations` is also very similar to this, but it shows Node statuses.

The `kubectl view-allocations` has a funny separation of "cpu" "memory" and "pods", whereas `kubectl resources` is a cleaner representation of Pod resources.

Per our guidelines, this plugin likely would require renaming to accurately represent what it does (like `pod-resources`), or `pod-usage`. I'll have @corneliusweig also review and provide input.

howardjohn:
You can also aggregate by namespace or node for what its worth, with `--by namespace` or `--by node`, so `pod-resources` doesn't seem quite accurate

ahmetb:
Yeah, in that case --by=node part is overlapping with `resource-capacity` plugin (also overlapping in name). I believe @corneliusweig opened the issue so we are not leaving any plugin authors uninformed about Krew, but let's see if it fits. Right now there's considerable overlap with existing plugins.

image

plugins/resources.yaml
```diff
@@ -0,0 +1,26 @@
...
+          os: linux
+          arch: amd64
+      uri: https://github.com/howardjohn/kubectl-resources/releases/download/v0.1.0/kubectl-resources_0.1.0_Linux_x86_64.tar.gz
+      sha256: ad3d1a126b106a38fb92ae76bb728b00ed8b4ef02eef574456b73f21c28affb3
+      bin: kubectl-resources
+  shortDescription: >-
+    Access Kubernetes resource requests, limits, and usage.
+  homepage: https://github.com/howardjohn/kubectl-resources
+  description: |
+    Access Kubernetes resource requests, limits, and usage.
```
corneliusweig: Maybe just briefly mention why `kubectl top` doesn't cut it?

howardjohn:
I don't have any current plans to add any more features to the plugin (maybe a --watch flag, but not a high priority), but I am open to suggestions.

I am fine with changing the name, `resource-overview` sounds right. Let me know if that is sufficient or if I would need additional distinguishing

ahmetb:
I'm good with both `resource-overview` and `resource-usage` (consider renaming your repo, and --help messages). 

This would help it distinguish from `resource-capacity` which is geared towards showing cluster-wide total and percentages, despite both plugins are largely overlapping.

corneliusweig:
Then let's add this plugin to the index too. Please let us know if you plan to rename your repo as suggested.

plugins/resources.yaml
```diff
@@ -0,0 +1,27 @@
...
+          os: linux
+          arch: amd64
+      uri: https://github.com/howardjohn/kubectl-resources/releases/download/v0.1.0/kubectl-resources_0.1.0_Linux_x86_64.tar.gz
+      sha256: ad3d1a126b106a38fb92ae76bb728b00ed8b4ef02eef574456b73f21c28affb3
+      bin: kubectl-resources
+  shortDescription: >-
+    Overview of Kubernetes resource requests, limits, and usage.
+  homepage: https://github.com/howardjohn/kubectl-resources
+  description: |
+    Overview of Kubernetes resource requests, limits, and usage. This plugin aggregates resource usage, similar to `kubectl top`,
```
corneliusweig: Can you wrap this at ~80 chars for easier reading on small terminals?

howardjohn:
Sounds good. I'll rename everything in my repo then send out an updated PR later. Thanks!

saiharshitachava:
@howardjohn is this availble with krew yet?

PR #437, opened 2020-01-14
Migrate to GitHub workflows for CI

### LINKED ISSUE #430: Move to Github Actions for CI
Ref: https://github.com/kubernetes-sigs/krew/issues/451

Let's do the same for krew-index repo.

/priority P3
/good-first-issue

alonyb:
This PR contain a new CI with github workflows, logs of the pipeline you can check [here](https://github.com/RubenBaez/krew-index/commit/2ea4312d11b2a83ddab67df8dc826ae935c49173/checks?check_suite_id=400451488)
Fixes #430

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: CI
+name: krew-index CI
 
 on: [push]
 
@@ -22,19 +22,19 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: install kubectl
+    - name: Install kubectl
       run: |
         sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
         sudo chmod +x /usr/bin/kubectl
 
-    - name: install krew (temporarily hardcode version)
+    - name: Install krew (temporarily hardcode version)
       run: |
         cd "$(mktemp -d)" &&
         curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
         tar zxvf krew.tar.gz &&
         ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
         
-    - name: get repository differences
+    - name: Get repository differences
       run: git diff --name-only --diff-filter=AM $TRAVIS_BRANCH...HEAD |
         grep -E 'plugins/.*' | tee /dev/stderr |
         xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,6 @@ jobs:
     - name: Validate updated plugin manifests
       run: |
         set -euo pipefail
-        git diff --name-only --diff-filter=AM $GITHUB_REF...HEAD |
+        git diff --name-only --diff-filter=AM $GITHUB_REF |
         grep -E 'plugins/.*' | tee /dev/stderr |
         xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         
     - name: Validate updated plugin manifests
       run: |
-        ls $HOME/bin
+        export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
         set -euo pipefail
         git diff --name-only --diff-filter=AM ${GITHUB_REF##*/} |
           grep -E 'plugins/.*' | tee /dev/stderr |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,16 +1,8 @@
 name: krew-index CI
 
-on: 
-  push:
-    branches: 
-      - '*'
-    tags: 
-      - 'v*.*.*'
-  pull_request:
-    branches: 
-      - '*'
-  release:
-    types: [published, created, edited]
+on:
+- push
+- pull_request
 
 jobs:
   build:
@@ -31,12 +23,15 @@ jobs:
 
     - name: Install krew
       run: |
+        set -euo pipefail
         cd "$(mktemp -d)" &&
         curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
         tar zxvf krew.tar.gz &&
         ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
         
     - name: Validate updated plugin manifests
-      run: git diff --name-only --diff-filter=AM $TRAVIS_BRANCH...HEAD |
+      run: |
+        set -euo pipefail
+        git diff --name-only --diff-filter=AM $GITHUB_REF...HEAD |
         grep -E 'plugins/.*' | tee /dev/stderr |
         xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,6 @@ jobs:
         set -euo pipefail
         sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
         sudo chmod +x /usr/bin/kubectl
-
-    - name: Set GOBIN
       env: 
         GOBIN: $HOME/bin go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,9 +38,8 @@ jobs:
         
     - name: Validate updated plugin manifests
       run: |
+        export GOBIN=$HOME/bin
         set -euo pipefail
         git diff --name-only --diff-filter=AM ${GITHUB_REF##*/} |
           grep -E 'plugins/.*' | tee /dev/stderr |
           xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
-      env:
-        GOBIN: $HOME/bin
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
         sudo chmod +x /usr/bin/kubectl
 
     - name: Install krew
+      shell: bash
       run: |
-        set -euo pipefail
         cd "$(mktemp -d)" &&
         curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
         tar zxvf krew.tar.gz &&
@@ -32,9 +32,6 @@ jobs:
     - name: Validate updated plugin manifests
       shell: bash
       run: |
-        echo ${GITHUB_REF##*/}
-        echo $HOME
-        set -euo pipefail
         git diff --name-only --diff-filter=AM ${GITHUB_REF##*/} |
           grep -E 'plugins/.*' | tee /dev/stderr |
           xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,23 +1,25 @@
 name: krew-index CI
 
-on: [push]
+on: 
+  push:
+    branches: 
+      - '*'
+    tags: 
+      - 'v*.*.*'
+  pull_request:
+    branches: 
+      - '*'
+  release:
+    types: [published, created, edited]
 
 jobs:
-
   build:
     name: Build and Tags
     runs-on: ubuntu-latest
     env: 
       KREW_VERSION: v0.3.3
       GO111MODULE: on
-
     steps:
-    
-    - name: Set up Go 1.13
-      uses: actions/setup-go@v1
-      with:
-        go-version: 1.13
-      id: go
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -27,14 +29,14 @@ jobs:
         sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
         sudo chmod +x /usr/bin/kubectl
 
-    - name: Install krew (temporarily hardcode version)
+    - name: Install krew
       run: |
         cd "$(mktemp -d)" &&
         curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
         tar zxvf krew.tar.gz &&
         ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
         
-    - name: Get repository differences
+    - name: Validate updated plugin manifests
       run: git diff --name-only --diff-filter=AM $TRAVIS_BRANCH...HEAD |
         grep -E 'plugins/.*' | tee /dev/stderr |
         xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,9 @@ jobs:
     steps:
 
     - name: Check out code into the Go module directory
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # disable shallow clones
 
     - name: Install kubectl
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+      KREW_VERSION: v0.3.3
+      GO111MODULE: on
+
+    steps:
+    
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
+      id: go
```
ahmetb: this is probably redundant since we don't depend on it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Install kubectl
+      run: |
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
+        sudo chmod +x /usr/bin/kubectl
+
+    - name: Install krew (temporarily hardcode version)
```
ahmetb: you can remove (...)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        cd "$(mktemp -d)" &&
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
+        tar zxvf krew.tar.gz &&
+        ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
+        
+    - name: Get repository differences
```
ahmetb: `name: Validate updated plugin manifests`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
+name: krew-index CI
+
+on: [push]
```
ahmetb: Does this work on PRs? We need it to primarily work on PRs than pushes actually :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+
```
ahmetb: consider eliminating empty lines

ahmetb:
Looks clean I like it!

corneliusweig:
Yeah, after applying the comments by @ahmetb, this looks good to go.

Let's first run this alongside travis, before we disable it.

We also need to ensure that prow does not merge before GH actions has succeeded, when the auto-approver bot was faster than the CI build.

/retitle Migrate to GitHub workflow
@RubenBaez  Can you add `Fixes #430` to the PR description?

ahmetb:
> We also need to ensure that prow does not merge before GH actions has succeeded, when the auto-approver bot was faster than the CI build.

That's valid, but with GH actions, GH actually knows there's a "check" much faster than Travis posting back something. So I highly doubt that'll be a problem.

We can run side-by-side for a month.

RubenBaez:
All suggestions were made, anyone else can tell me :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    branches: 
+      - '*'
+    tags: 
+      - 'v*.*.*'
+  pull_request:
+    branches: 
+      - '*'
+  release:
+    types: [published, created, edited]
```
ahmetb: can we simply just do:

```
on:
- push
- pull_request
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Install krew
+      run: |
+        cd "$(mktemp -d)" &&
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
+        tar zxvf krew.tar.gz &&
+        ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
+        
+    - name: Validate updated plugin manifests
+      run: git diff --name-only --diff-filter=AM $TRAVIS_BRANCH...HEAD |
```
ahmetb: as you can imagine $TRAVIS_BRANCH won't work here.
please find the github equivalent.

@corneliusweig my this comment is also a good proof that without bash strict mode enabled, github scripts are powerless. they don't even notify us about undefined variables.

@RubenBaez please add `set -euo pipefail;` to the beginning of this script (as well as the script above).

corneliusweig: Aww cr*p. Good that you found this!

.github/workflows/ci.yml
alonyb: of course, it's done :)

.github/workflows/ci.yml
alonyb: Yes, good observation. I fixed it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
+        sudo chmod +x /usr/bin/kubectl
+
+    - name: Install krew
+      shell: bash
+      run: |
+        cd "$(mktemp -d)" &&
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
+        tar zxvf krew.tar.gz &&
+        ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
```
alonyb: @ahmetb I can't use pipefail with pipes in the last stage, pipefail does not work with pipes. But in this stage I can, just adding a shell label with bash, it contains pipefail.

ahmetb: we'll need at least some `set -u`. I also must admit I don't fully understand your comment. What happens if you just add `set -euo pipefail;` to beginning of every script?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Install kubectl
+      run: |
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
+        sudo chmod +x /usr/bin/kubectl
+
+    - name: Install krew
+      shell: bash
```
corneliusweig: Do you have some reference where these options are documented? I wonder what the implications of `shell: bash` are (I'm not saying this is bad in any way).

ahmetb: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#using-a-specific-shell

bash is the best option here.

corneliusweig: Looks like `set -u` is still required.

alonyb: @corneliusweig I agree with you, it's done :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+      run: |
+        cd "$(mktemp -d)" &&
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
+        tar zxvf krew.tar.gz &&
+        ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
+        
+    - name: Validate updated plugin manifests
+      run: |
+        set -u
```
ahmetb: I honestly don't know why don't you just add `set -euo pipefail` which we discussed before.

Imagine `git diff` failed. This `run` shell will NOT fail, because `git` is not the last command in the pipe. No data will flow to `xargs`, so it will succeed, which will make us think everything worked.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+      KREW_VERSION: v0.3.3
+      GO111MODULE: on
+    steps:
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Install kubectl
+      run: |
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
```
ahmetb: you need `set -e`.
we discussed this before, only last command of a shell causes a shell failure. if curl fails here, this command will still succeed because you didn't do `set -e` or use &&.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+        sudo chmod +x /usr/bin/kubectl
+
+    - name: Install krew
+      shell: bash
+      run: |
+        cd "$(mktemp -d)" &&
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}" &&
```
ahmetb: similarly if this can't access ${KREW_VERSION} it will not fail.

I want to see `set -euo pipefail` everywhere more than a single command is executed, or there are pipes or there are variables in the `run:` block.

.github/workflows/ci.yml
corneliusweig: The `shell: bash` setting enables `set -eo pipefail`. However since it doesn't enable `-u` which we also want, I'd recommend to not use `shell: bash` and instead be very explicit about `set -euo pipefail` in all scripts.

ahmetb: I think we can keep `shell: bash` (I assume that is the default), and still `set -euo pipefail` instead of relying on Github to pull the rug under us in the future.

corneliusweig: I fully agree. And you're right, `bash` is the default shell.

alonyb: it's done you can check it.

.github/workflows/ci.yml
alonyb: I already added it.

.github/workflows/ci.yml
alonyb: I already added it `set -euo pipefail`

alonyb:
@ahmetb
 I take into account all your observations you can check it :)
@corneliusweig 
I had to change the checkout version from v2 to v1 for git themes.
[here](https://github.com/alonyb/krew-index/commit/608f50a32a34640e0c901a44da7584b0fb7a16fd/checks?check_suite_id=409970843) is the logs of the final file.
One more thing, if someone makes a change (push, PR) without changing plugins / *, it will throw an error, I imagine that this should be the behavior.

corneliusweig:
> One more thing, if someone makes a change (push, PR) without changing plugins / *, it will throw an error, I imagine that this should be the behavior.

:thinking: why is that? This is not ideal, because there are maintenance PRs from time to time where we don't change any manifest files.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+  build:
+    name: Build and Tags
+    runs-on: ubuntu-latest
+    env: 
+      KREW_VERSION: v0.3.3
+      GO111MODULE: on
+    steps:
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
```
corneliusweig: Instead of switching back to v1, you could also set
```yaml
with:
    fetch-depth: 0  # disable shallow clones
```
(see https://github.com/actions/checkout)

This should be more forward-compatible.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Install kubectl
+      run: |
+        set -euo pipefail
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl
+        sudo chmod +x /usr/bin/kubectl        
+
+    - name: Install krew
+      shell: bash
```
corneliusweig: as it's the default, let's remove `shell: bash` here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        set -euo pipefail
+        cd "$(mktemp -d)" &&
```
corneliusweig: I believe the `&&` in this script is no longer necessary, because of `set -e` above?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+        
+    - name: Install validate-krew-manifest plugin
+      run : |
+        set -euo pipefail
+        export GOBIN=$HOME/bin
+        go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
+        
+    - name: Validate updated plugin manifests
+      run : |
+        set -euo pipefail
```
corneliusweig: It worries me that this fails, if there are no plugins in the changeset. Can you try to run with `set -x` to see what are the exact commands?

alonyb:
> > One more thing, if someone makes a change (push, PR) without changing plugins / *, it will throw an error, I imagine that this should be the behavior.
> 
> 🤔 why is that? This is not ideal, because there are maintenance PRs from time to time where we don't change any manifest files.

@corneliusweig 
So, I could put an IF statement or something to validate it and run that step only when there are changes on settings/* and if there are no changes in settings/* I just pass that step. What do you think?

corneliusweig:
I wonder why this worked on travis..

Anyway, if `grep` is making problems, then let's get rid of it. Can you try this:

```sh
git diff --name-only --diff-filter=AM ${{ github.event.before }} ${{ github.sha }} -- plugins |
   tee /dev/stderr |
   xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
```

plugins/whoami.yaml
```diff
@@ -36,4 +36,3 @@ spec:
       https://github.com/rajatjindal/kubectl-whoami
   description: |
     This plugin show the subject that's currently authenticated as.
-
```
corneliusweig: That extra line should not be necessary if you fix that validation script.

corneliusweig:
Hey @alonyb, I left some more comments, did you see them?

.github/workflows/ci.yml
alonyb: Yes, it's a better option.

.github/workflows/ci.yml
alonyb: Sure, it's done.

.github/workflows/ci.yml
alonyb: I did, [here](https://github.com/alonyb/krew-index/runs/409226155?check_suite_focus=true) is the result. It didn't work

.github/workflows/ci.yml
alonyb: Of course, I removed that.

plugins/whoami.yaml
alonyb: I have restored the plugins folder

alonyb:
> Hey @alonyb, I left some more comments, did you see them?

@corneliusweig 
I did not see them, sorry, but I already made some changes. Check it :)

ahmetb:
Looks good. I think we can merge. 
Did you test and manifests and observe them fail or pass after being tested?

alonyb:
> Looks good. I think we can merge.
> Did you test and manifests and observe them fail or pass after being tested?

@ahmetb 
Yes, just one concern, it’s about last step. I think pipeline should run if someone does not change the plugins/* folder, currently if plugins/* is not changed, the pipeline will fail because `git diff`
What do you think?

alonyb:
> > Looks good. I think we can merge.
> > Did you test and manifests and observe them fail or pass after being tested?
> 
> @ahmetb
> Yes, just one concern, it’s about last step. I think pipeline should run if someone does not change the plugins/* folder, currently if plugins/* is not changed, the pipeline will fail because `git diff`
> What do you think?

Should I add a new validation for that. Or that is the right way :)

.github/workflows/ci.yml
corneliusweig: :thinking: The run you linked to seems to be a version with `grep` and has a change in `access-matrix.yaml`. Have you tried the variant without `grep`:
```sh
git diff --name-only --diff-filter=AM ${{ github.event.before }} ${{ github.sha }} -- plugins/ |
          tee /dev/stderr |
          xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
```

Let's run this with `set -x` and see if it fails.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+        cd "$(mktemp -d)"
+        curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/${KREW_VERSION}/krew.{tar.gz,yaml}"
+        tar zxvf krew.tar.gz
+        ./krew-linux_amd64 install --manifest=krew.yaml --archive=krew.tar.gz
+        
+    - name: Install validate-krew-manifest plugin
+      run : |
+        set -euo pipefail
+        export GOBIN=$HOME/bin
+        go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
```
corneliusweig: I'm just wondering why this even works. In the manifest for `krew` GH actions, we had to setup go. But not here, even though this clearly needs go.

Do you know what's going on here? @alonyb

alonyb: @corneliusweig 
Yes, `ubuntu-latest` contains some versions of Go. You can see that [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#ubuntu-1804-lts), but anyway I will put go 1.13 in another step to be clear.

.github/workflows/ci.yml
alonyb: @corneliusweig 
It looks perfect, [here](https://github.com/alonyb/krew-index/runs/410034531?check_suite_focus=true#step:7:18) is the log.

alonyb:
@ahmetb @corneliusweig 
We can merge this PR.

ahmetb:
/lgtm
/approve

unnamed

corneliusweig:
Very nice! Thanks a lot for your persistence @alonyb!

PR #438, opened 2020-01-14
Add kubepug (kubectl deprecations)

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

Kubepug (Pre UpGrade check) is a plugin that validates existing Kubernetes objects against a versioned swagger.json from k/k repo. 

It's useful to verify if some existing object will be discontinued / deprecated or even removed in a future Kubernetes version

### DIFF
--- a/plugins/deprecations.yaml
+++ b/plugins/deprecations.yaml
@@ -1,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: deprecations
+spec:
+  shortDescription: >-    
+    Checks for deprecated objects in a cluster 
+  homepage: https://github.com/rikatz/kubepug
+  caveats: |
+    * By default, deprecations finds deprecated object relative to the current kubernetes
+    master branch. To target a different kubernetes release, use the --k8s-version
+    argument.
+    
+    * Deprecations needs permission to GET all objects in the Cluster
+  description: |     
+    This plugin shows all the deprecated objects in a Kubernetes cluster allowing 
+    the operator to verify them before upgrading the cluster. It uses the 
+    swagger.json version available in master branch of Kubernetes repository
+    (github.com/kubernetes/kubernetes) as a reference. The branch can be changed 
+    to some other desired Kubernetes version
+  version: "v0.2.0"
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/rikatz/kubepug/releases/download/v0.2.0/kubepug_darwin_amd64.tar.gz
+    sha256: "647a56e53b81ea4349fee6a515173bf9cfbef1173f20f591cb91ead6a77b1002"
+    bin: "kubepug" 
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rikatz/kubepug/releases/download/v0.2.0/kubepug_linux_amd64.tar.gz
+    sha256: "82f83df4a2896e6b12a62fc8d6a2f295cb926ca1ee29085ba83d54ca12a5018e"
+    bin: "kubepug" 
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/rikatz/kubepug/releases/download/v0.2.0/kubepug_windows_amd64.zip
+    sha256: "4a26e35e6b6a84c679fc73beaaa9ee4e169ae25eb750d886dd0559d9a0d9934c"
+    bin: "kubepug.exe" 


plugins/pug.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - selector:             # a regular Kubernetes selector
+      matchLabels:
+        os: linux
+        arch: amd64
+    # url for downloading the package archive:
+    uri: https://github.com/rikatz/kubepug/releases/download/v0.1.1/kubepug_linux_amd64.tar.gz
+    # sha256sum of the above archive file:
+    sha256: "d25e8a055ad535e2e85527dbc19ee900205213e393c468a7c5a1b3d1c5ad7db8"
+    files:
+    - from: "*"
```
ahmetb: Omit files: section since this is the default value

plugins/pug.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pug               # plugin name must match your manifest file name (e.g. foo.yaml)
```
ahmetb: Please remove comments

plugins/pug.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  shortDescription: >-    # short description gets truncated at ~50 chars
+    Checks existing deprecated objects in a cluster 
+  # (optional) url of the project homepage
+  homepage: https://github.com/rikatz/kubepug
+  # (optional) use caveats field to show post-installation recommendations
+  caveats: |
+    * Kubepug downloads Kubernetes swagger.json to check objects and which are deprecated
```
ahmetb: These aren’t really caveats though. Maybe just delete.

plugins/pug.yaml
```diff
@@ -0,0 +1,56 @@
...
+  homepage: https://github.com/rikatz/kubepug
+  # (optional) use caveats field to show post-installation recommendations
+  caveats: |
+    * Kubepug downloads Kubernetes swagger.json to check objects and which are deprecated  
+    according to its description. It uses by default "master" branch and should be changed 
+    to the desired Kubernetes version with the flag --k8s-version
+  description: |     
+    This plugin shows all the deprecated objects in a Kubernetes cluster allowing the operator
+    to verify them before upgrading the cluster. It uses the swagger.json version available in
+    master branch of Kubernetes repository (github.com/kuebrnetes/kubernetes) as a reference.
```
ahmetb: Typo

plugins/pug.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pug               # plugin name must match your manifest file name (e.g. foo.yaml)
+spec:
+  shortDescription: >-    # short description gets truncated at ~50 chars
+    Checks existing deprecated objects in a cluster
```
corneliusweig:
```suggestion
    Checks for deprecated objects in a cluster 
```

plugins/pug.yaml
corneliusweig: I think the part that talks about "master" default and that one should set the k8s version is good for the caveats section.

I presume, you also need to be allowed to list all objects? This could also be mentioned here.

plugins/pug.yaml
corneliusweig:
```suggestion
    master branch of Kubernetes repository (github.com/kubernetes/kubernetes) as a reference. 
```

plugins/pug.yaml
rikatz: Tks

plugins/pug.yaml
rikatz: Yes, the thing is that the behavior might change from the default flags to an specific flag.

As an example, if I'm upgrading from k8s 1.15 to 1.17 and verify against master (and not against v1.17.0) it produces different results in NetworkPolicy alert (extensions/v1beta1 still exists in 1.17.0 but is already removed in master branch).

And yes, it needs a permission allowing to get all objects in the cluster so it will be mentioned! Thank you!!

rikatz:
@ahmetb @corneliusweig changed as suggested.

Additionally I've changed the description and caveats to have a maximum of 80 cols to fit better in terminals :)

plugins/pug.yaml
```diff
@@ -0,0 +1,45 @@
...
+spec:
+  shortDescription: >-    
+    Checks for deprecated objects in a cluster 
+  homepage: https://github.com/rikatz/kubepug
+  caveats: |
+    * Kubepug downloads Kubernetes swagger.json to check objects and which are
+    deprecated according to its description. It uses by default "master" branch and
+    should be changed to the desired Kubernetes version with the flag --k8s-version
+    
+    * Kubepug needs permission to GET all objects in the Cluster, so you should run
```
corneliusweig:
```suggestion
    * Kubepug needs permission to GET all objects in the Cluster
```
I think that already says enough. No need to restate it in a different way.

plugins/pug.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pug
+spec:
+  shortDescription: >-    
+    Checks for deprecated objects in a cluster 
+  homepage: https://github.com/rikatz/kubepug
+  caveats: |
+    * Kubepug downloads Kubernetes swagger.json to check objects and which are
```
corneliusweig: I just read this as "the swagger.json is deprecated". Probably not what you mean :)

Maybe change to something like

> By default, pug finds deprecated object relative to the current kubernetes master branch. To target a different kubernetes release, use the --k8s-version argument.

plugins/pug.yaml
```diff
@@ -0,0 +1,45 @@
...
+    should be changed to the desired Kubernetes version with the flag --k8s-version
+    
+    * Kubepug needs permission to GET all objects in the Cluster, so you should run 
+    it as a Cluster Admin or give the right permissions to the user executing the
+    plugin
+  description: |     
+    This plugin shows all the deprecated objects in a Kubernetes cluster allowing 
+    the operator to verify them before upgrading the cluster. It uses the 
+    swagger.json version available in master branch of Kubernetes repository
+    (github.com/kubernetes/kubernetes) as a reference. The branch might be changed
```
corneliusweig:
```suggestion
    (github.com/kubernetes/kubernetes) as a reference. The branch can be changed 
```

rikatz:
@corneliusweig thank you! I've made the suggested changes :)

corneliusweig:
Sorry if I'm super annoying. But while reviewing, I consistently thought "pug -- what was that good for again?". And I think I'm not the only one. Given that your plugin will not be used very frequently, why not use a more descriptive name? This will help your users remember more easily.
So what about something like `deprecated-objects` or `outdated-objects` or just `deprecations`?

ahmetb:
I'm actually not against the unique/creative names.
But it comes at a cost of discovery. If the plugin authors want to retain the name in this case, we should probably allow them. But they should be aware that they won't get as much visibility from krew-index as they could, with a descriptive name.

rikatz:
Actually you’re right about that. I’ve started kubepug not with the idea of
being a kubectl plugin but just happened :)

I’ll think tomorrow, maybe I can keep the naming but, when using krew it
gots an alias of “deprecated-objects” or something like that, WDYT?

Em qua, 15 de jan de 2020 às 22:02, Ahmet Alp Balkan <
notifications@github.com> escreveu:

ahmetb:
Yeah, that absolutely works. We have many plugins like that. 
Make sure to also update your README and -help messages to reflect the krew name when run as a krew plugin (args[0]).

corneliusweig:
Oh, I would be totally fine with `pug` too. But as Ahmet said, it comes at a price. However, long names also have a price because nobody wants to type long names.
In this case, I think it's not executed frequently so that a long and descriptive name would work better. In the end, it's your choice. Just let us know when you have decided :)

rikatz:
> Oh, I would be totally fine with `pug` too. But as Ahmet said, it comes at a price. However, long names also have a price because nobody wants to type long names.
> In this case, I think it's not executed frequently so that a long and descriptive name would work better. In the end, it's your choice. Just let us know when you have decided :)

Yeap, you both are right. Changing here to 'deprecations', the day was pretty busy today to work on this but will continue and finish tomorrow :)

corneliusweig:
So you mean you want to rename to `deprecations`?

rikatz:
@corneliusweig I will keep the 'pug' name in the original problem (sorry, I have some affection with the name and the logo) but in Krew I'll change it to 'deprecations' and also make the suggested changes from @ahmetb in the cmd/kubepug.go to reflect the binary name :)

rikatz:
@ahmetb @corneliusweig PTAL to see if this is fine now :)

ahmetb:
> I will keep the 'pug' name in the original problem

now I'm thinking we can keep krew name `pug` as well. Since it's unique enough I don't have a problem with it, just that the discoverability-by-name would be low. If you are ok with it as the author. We can keep it?

rikatz:
@ahmetb I've been also thinking, for krew 'deprecations' is a good name :)

I will follow your suggestions and have a better discoverability and keep the 'Pug' name only in the original repo.

ahmetb:
Thanks. We'll go this way.

I still recommend these two strongly:

1. Make sure your repo mentions krew installation method with the name here.
2. When your plugin is executed with args[0]=*/kubectl-deprecations* , please adjust your -h/--help messages and usage strings to show this name, not 'pug', to not to confuse users installing this way.

rikatz:
Yeap, about the args I'm already using os.args[0] :) 

https://github.com/rikatz/kubepug/blob/master/cmd/kubepug.go#L25-L32

I'll change the README right now mentioning how to use it through krew!

Thank you very much!

PR #457, opened 2020-01-26
add vagrant plugin

huazhihao:
`kube-vagrant` is a kubectl plugin to create a multi-node kubernetes cluster locally with vagrant, so you can quickly set up a dev cluster with one command line.

### DIFF
--- a/plugins/vagrant.yaml
+++ b/plugins/vagrant.yaml
@@ -1,0 +1,44 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: vagrant
+spec:
+  homepage: https://github.com/huazhihao/kube-vagrant
+  shortDescription: create a multi-node kubernetes cluster locally with vagrant
+  version: v0.1.0
+  description: |
+    `kube-vagrant` is a kubectl plugin to create a multi-node kubernetes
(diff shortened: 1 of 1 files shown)


ahmetb:
Yes, cluster bootstrap tooling it not great for plugins. 
Similarly Vagrant is a trademark of HashiCorp so we’ll not be able to accept it.
You’re free to still distribute it outside krew, however.

huazhihao:
thanks @corneliusweig @ahmetb for your review comments. I will leave it as-is.

huazhihao:
> Hey @huazhihao,
> 
> this is great work, but I'm not sure it qualifies for a good plugin. As far as I can tell, this is usually a one-time setup, so that the plugin gets to run just a single time. Scripts like these have also previously been rejected as plugins, because krew's focus is on repeatable commands. For example, there was a request to have a tiller installer for helm-2 which eventually got rejected (#87).
> 
> Let's hear @ahmetb's opinion about this.
> 
> Btw, this looks dangerous 😉 https://github.com/huazhihao/kube-vagrant/blob/c3eee010eea50be7fdfd9882c657b76a43d1bebe/kube-vagrant#L194

that wasn't under home directory.

corneliusweig:
> that wasn't under home directory.

As far as I can tell, the directory is unspecified. So it could be any directory, including $HOME. Right?

PR #472, opened 2020-02-04
Add doker as a plugin

Pothulapati:
Adds `doker` as a plugin.

- do**k**er is a cli tool to get a docker-cli experience but on a kubernetes cluster. This requires a `daemonset` to be installed. More Information https://github.com/pothulapati/doker
- Currently, The main use-case is to deal with docker images present on the nodes i.e list, load, prune, etc. But the plan is to replicate more commands from the docker-cli.
- Running `kubectl krew install --manifest=plugins/doker.yaml` works.

Thanks :)

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

### DIFF
--- a/plugins/doker.yaml
+++ b/plugins/doker.yaml
@@ -1,0 +1,57 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: doker
+spec:
+  version: v0.1.0
+  homepage: https://github.com/pothulapati/doker
+  shortDescription: Gives docker-cli experience but for a kubernetes cluster.
+  description: |
+    This plugin gives users a docker-cli like experinece but in the context of
+    a kubernetes cluster, Allowing users to control and manage images, containers 
+    running in the kubernetes nodes.
+  caveats: |
+    * A Kubernetes daemonset has to be installed first which acts like an agent to 
+    which the plugin sends commands. This can be done by running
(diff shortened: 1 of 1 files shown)


ahmetb:
It appears like this isn’t really meant to enhance “Kubernetes” experience but rather a separate tool that is meant to imitate Docker experience on Kubernetes. 

I am currently -1 on accepting this to the central plugin index but I’ll wait for input from my maintainers.

Pothulapati:
@ahmetb but my thinking was that it uses `kubeconfig` just like any other krew tool, and talks to the `daemonset` to perform some action.

corneliusweig:
While the answers you can get with `doker` may be interesting, I feel they don't matter on the abstraction level of kubernetes. Think about listing the container images on all nodes of your cluster. Even on my laptop, this list is huge. Take that times the number of nodes and it becomes an even bigger mess.

Also it't not clear to me what you want to achieve by pulling an image on all nodes. Isn't that the job of the kubelet? You shouldn't need to care about this.

ahmetb:
We probably shouldn't deliberate/advise on the contents of the plugins much. :) 

Just because something uses kubeconfig file, it doesn't make them a kubeconfig plugin. It should enhance the kubectl experience as well. I recommend you distribute this as a "standalone tool".

ahmetb:
BTW You can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

PR #475, opened 2020-02-05
Add plugin: cluster-group

u2takey:
This [plugin](https://github.com/u2takey/kubectl-clusters) allows users exec kubectl commands in multiple clusters with group.

PR #476, opened 2020-02-05
add if statement for pr & push event

alonyb:
This should fix #462 and #475 (`force-pushed` cases) [logs](https://github.com/alonyb/krew-index/runs/427491287?check_suite_focus=true)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
         export GOBIN=$HOME/bin
         go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
       
-    - name: Validate updated plugin manifests PR case
+    - name: Validate updated plugin manifests
       run : |
         set -x
         set -euo pipefail
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 name: krew-index CI
 
-on:
-- push
-- pull_request
+on: pull_request
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


ahmetb:
Thanks. I am currently -1 for this duplication.  
I am ~80% sure there’s a better way to do this, if not, I am questioning capabilities of Github Workflows system. Do you mind asking on their community forums if there’s a way to do this cleanly?

.github/workflows/ci.yml
```diff
@@ -44,11 +44,23 @@ jobs:
         export GOBIN=$HOME/bin
         go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
         
-    - name: Validate updated plugin manifests
+    - name: Validate updated plugin manifests PUSH case
+      if: github.event_name == 'push'
```
ahmetb: We need plugin validation to run only in PRs in all honesty.

Once merged, it’s too late to check manifests as we can’t act on it. Am i right @corneliusweig ?

alonyb: > Thanks. I am currently -1 for this duplication.
> I am ~80% sure there’s a better way to do this, if not, I am questioning capabilities of Github Workflows system. Do you mind asking on their community forums if there’s a way to do this cleanly?

Sure, that was a quick solution. For those cases. But there are better options.

corneliusweig: I agree. The tests should only run on PR and pushes to pull requests.

The duplication between the steps does not make sense to me. I still don't understand why we don't just do 
```sh
git diff --name-only --diff-filter=AM origin/master ${{ github.sha }} -- plugins/
```

This would work fine with force push AND it would _always_ run the test for everything that is about to be merged into master. Doesn't that sound right?

ahmetb: Yeah. And travis was gracious enough to provide the right value as env vars.

ahmetb:
@alonyb can you update with the suggested changes? I think we don't need to build on "push" (I'm assuming it means pushing to master). We still need to rebuild on pushes to PR branches.

.github/workflows/ci.yml
alonyb: > I agree. The tests should only run on PR and pushes to pull requests.
> 
> The duplication between the steps does not make sense to me. I still don't understand why we don't just do
> 
> ```shell
> git diff --name-only --diff-filter=AM origin/master ${{ github.sha }} -- plugins/
> ```
> 
> This would work fine with force push AND it would _always_ run the checks for everything that is about to be merged into master. Doesn't that sound right?

In PR event it is the better option. But for push event, I needed a hash to compare. Now that it's just PR it's simple.

alonyb:
> can you update with the suggested changes? I think we don't need to build on "push" (I'm assuming it means pushing to master). We still need to rebuild on pushes to PR branches.

I just deleted the push event, however, I could only be able to push on the master branch. Or only in PR, depending on what is needed now.

ahmetb:
Thanks! Let's give this a shot.
/lgtm
/approve

corneliusweig:
@alonyb Looks like we have a problem, see https://github.com/kubernetes-sigs/krew-index/pull/485

corneliusweig:
PR #486 should fix this.

PR #486, opened 2020-02-08
Fix manifest validation in GitHub actions pipeline

corneliusweig:
Previously, the changed manifests were determined by a git command
    equivalent to `git diff <commit>`. This checks for dirty files in the
    working directory, so this never detects any changes in the CI where
    everything is clean.
    
    Instead, we need to run a command `git diff origin/master <commit>` to
    detect all changes that are new in the current branch and are not
    present in ref origin/master. For that to work, the master branch needs
    to be pulled explicitly.

This fixes a problem that was introduced in #476.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,9 @@ jobs:
       with:
         fetch-depth: 0  # disable shallow clones
 
-    - name: Fetch origin/master
+    - name: Fetch base branch
       run: |
-        git fetch --no-tags --prune --depth=1 origin +refs/heads/master:refs/remotes/origin/master
+        git fetch --no-tags --prune --depth=1 origin +refs/heads/${GITHUB_BASE_REF}:refs/remotes/origin/${GITHUB_BASE_REF}
 
     - name: Install kubectl
       run: |
@@ -51,7 +51,7 @@ jobs:
         set -x
         set -euo pipefail
         export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
-        git diff --name-only --diff-filter=AM origin/master ${{ github.sha }} -- plugins/ |
+        git diff --name-only --diff-filter=AM origin/${GITHUB_BASE_REF} ${{ github.sha }} -- plugins/ |
           tee /dev/stderr |
           xargs -I _ $HOME/bin/validate-krew-manifest -manifest _
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,27 +6,31 @@ jobs:
   build:
     name: Build and Tags
     runs-on: ubuntu-latest
-    env: 
+    env:
       KREW_VERSION: v0.3.3
       GO111MODULE: on
     steps:
-    
+
     - name: Set up Go 1.13
       uses: actions/setup-go@v1
       with:
         go-version: 1.13
       id: go
 
-    - name: Check out code into the Go module directory
+    - name: Check out PR branch
       uses: actions/checkout@v2
       with:
         fetch-depth: 0  # disable shallow clones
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -35,19 +39,19 @@ jobs:
...
         go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
-      
+
     - name: Validate updated plugin manifests
       run : |
         set -x
         set -euo pipefail
         export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
-        git diff --name-only --diff-filter=AM ${{ github.sha }} -- plugins/ |
+        git diff --name-only --diff-filter=AM origin/master ${{ github.sha }} -- plugins/ |
```
ahmetb: I think this will save the day but i don’t think it’s the ultimate solution.

GH should be providing us what the base branch is. 
This wouldn’t work if we had multiple branches for some reason and sent PRs to them.

ahmetb: Quick search reveals https://stackoverflow.com/a/57862507/54929 env vars like  "$GITHUB_BASE_REF..$GITHUB_SHA" could these be useful?

alonyb: It could be useful, but I don't know how Github will get the difference.

I mean I already tried this one, but the problem I found:
The local repository in your build doesn't contain the commits you're referencing between original and forked repo, then it will through `fatal: bad object` or `fatal: bad revision` as [here](https://github.com/alonyb/krew-index/runs/435257792?check_suite_focus=true#step:7:18)

Maybe using something like [this](https://github.com/bdougie/fetch-upstream)  to fetch repos, my concerns appear from [here](https://stackoverflow.com/questions/25192708/how-can-i-compare-my-local-forked-repository-with-changes-that-may-have-been-mad)

It just a concern maybe I am wrong.

corneliusweig: Good point. Let's see if the modified version does the trick.

corneliusweig: Ok, looks good. If you have had a look, I'll remove the broken manifest so that we can merge.

ahmetb:
Looks good but isn't `GITHUB_BASE_REF` a commit number already? `origin/` prefix might not be necessary? Or is it a branch name?

corneliusweig:
As far as I understand the docs (https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables), it is just the branch name in the merge base repo. The script runs with `set -x` and shows that the `origin/` prefix is indeed needed.

corneliusweig:
I removed the on-purpose broken commit. Once the build is green, we could consider merging it.

ahmetb:
It's still surprising that a step like

```
    - name: Fetch base branch
      run: |
        git fetch --no-tags --prune --depth=1 origin +refs/heads/${GITHUB_BASE_REF}:refs/remotes/origin/${GITHUB_BASE_REF}

```

is necessary –given most everyone should be using Workflows to verify PRs.

Could this actually be an option for the `checkout` action we already have? 

But if you think this is the way to go, I'll lgtm. It's just so surprising to me it takes so much effort.

corneliusweig:
The closest thing I can find in the examples is this: https://github.com/actions/checkout#fetch-all-branches

It looks like this is the standard way. We can only make it terser by skipping `name` or so.

corneliusweig:
I guess that the great majority of users just runs tests at the tip and does not care so much about the changes between to refs. That would explain why `actions/checkout` has such a weird way for it.

ahmetb:
Possibly a feature request to actions/checkout but ain't nobody got time fo that. :)

/lgtm

PR #497, opened 2020-02-14
Add 'kubectl-topology'.

bmcstdio:
This PR adds the first version of `kubectl-topology`, a plugin meant to provide insights into the topology of Kubernetes clusters (i.e. into the distribution of nodes and pods per region and zone).

### DIFF
--- a/plugins/topology.yaml
+++ b/plugins/topology.yaml
@@ -1,0 +1,65 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: topology
+spec:
+  version: v0.1.0
+  shortDescription: Explore region topology for nodes or pods
+  homepage: https://github.com/bmcstdio/kubectl-topology
+  description: |
+    Provides insight into the topology of a Kubernetes cluster (i.e. into the
+    distribution of nodes and pods per region and zone).
+
+    NOTE: Requires nodes to be adequately labeled with the
+    'topology.kubernetes.io/{region,zone}' labels or their deprecated
+    equivalents, 'failure-domain.beta.kubernetes.io/{region,zone}'.
+
+    # List all pods across all namespaces and the zones and regions they are in.
+    $ kubectl-topology pod --all-namespaces
+    NAMESPACE     NAME                                     NODE                          REGION         ZONE
+    kube-system   kube-dns-5f886bf8d8-s7pcc                gke-gke-1-p-1-50120dfc-26gm   europe-west1   europe-west1-d
+    kube-system   kube-dns-5f886bf8d8-zhm42                gke-gke-1-p-1-8e1077f6-17st   europe-west1   europe-west1-b
+    kube-system   kube-dns-autoscaler-8687c64fc-d5jvm      gke-gke-1-p-1-8e1077f6-17st   europe-west1   europe-west1-b
+    kube-system   kube-proxy-gke-gke-1-p-1-50120dfc-26gm   gke-gke-1-p-1-50120dfc-26gm   europe-west1   europe-west1-d
+    kube-system   kube-proxy-gke-gke-1-p-1-7023cbca-cz4b   gke-gke-1-p-1-7023cbca-cz4b   europe-west1   europe-west1-c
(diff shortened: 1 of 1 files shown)


plugins/topology.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: topology
+spec:
+  version: v0.0.1
+  shortDescription: Provides insight into the topology of a Kubernetes cluster.
```
ahmetb: Rm trailing period

plugins/topology.yaml
```diff
@@ -0,0 +1,46 @@
...
+  homepage: https://github.com/bmcstdio/kubectl-topology
+  description: Provides insight into the topology of a Kubernetes cluster.
```
ahmetb: Please write something more thorough. 
I still cannot understand what this plugin does. 
Ideally users should not be forced to download to understand.

bmcstdio: Done. I was going to expand on this later on, but I agree now's probably as good a time as any other.

bmcstdio:
@ahmetb thank you for reviewing, as well as for your feedback. I have addressed your comments. Could you please re-review?

plugins/topology.yaml
```diff
@@ -0,0 +1,84 @@
...
+    # List all nodes in the cluster and the zones and regions they belong to.
+    $ kubectl topology node
+    NAME     REGION         ZONE
+    node-1   europe-west1   europe-west1-a
+    node-2   europe-west1   europe-west1-a
+    node-3   europe-west1   europe-west1-b
+    node-4   europe-west1   europe-west1-c
+
+    # List nodes in the 'europe-west1-a' zone only.
+    $ kubectl topology node --zone europe-west1-a
```
ahmetb: I think 1 example is more than enough. 
You can probably remove the rest And make the last paragraph more prominent.

plugins/topology.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: topology
+spec:
+  version: v0.0.1
+  shortDescription: Provides insight into the topology of a Kubernetes cluster
```
ahmetb: Better description for you: 

> Browse zones/regions of Nodes or Pods

ahmetb: > Explore region topology for nodes or pods

plugins/topology.yaml
bmcstdio: Done!

plugins/topology.yaml
bmcstdio: I appreciate the suggestion, but I have this same description in the `README.md` file for now. Would you mind me changing that later on?

ahmetb: I think the question would be why not now? :)
You can have different descriptions in different places.

bmcstdio: Yeah, guess you're right. I've amended the description as per one of your suggestions.

ahmetb:
It seems like the plugin isn't usable in cloud providers right away because you aren't importing auth plugins. e.g.

```
Error: no Auth Provider found for name "gcp"
Usage:
  kubectl-topology node [flags]
```

ahmetb:
I can probably merge as is but I'd probably ask you to release a new version right away. Right now I'm not able to test this myself.

Also make sure to check out https://github.com/kubernetes-sigs/krew/blob/master/docs/DEVELOPER_GUIDE.md#automate-releasing-new-versions-on-krew-index for release automation from github.

bmcstdio:
@ahmetb I've made the required changes, released a new version and changed the manifest accordingly. Could you please check again if it is working for you? Thank you very much for your feedback.

ahmetb:
Cool. Thanks.
/lgtm
/approve

bmcstdio:
@ahmetb awesome, thank you! I've noticed that https://github.com/kubernetes-sigs/krew-index/blob/master/plugins.md wasn't updated. Does it trigger automatically or do I/you need to do something?

ahmetb:
There’s a tool to update it as mentioned in the page,you can make a PR or it will be picked up on the next PR.

bmcstdio:
@ahmetb great, thank you! I've opened #500.

PR #499, opened 2020-02-16
add plugin: kubectl node-restart

MnrGreg:
This plugin provides cluster-admins with the ability to perform graceful, rolling restarts of Kubernetes cluster Nodes.

### DIFF
--- a/plugins/node-restart.yaml
+++ b/plugins/node-restart.yaml
@@ -17,11 +17,12 @@ spec:
       to: "."
     bin: "node-restart.sh"
   shortDescription: >-
-    Restart Kubernetes Nodes sequentially & gracefully
+    Restart cluster nodes sequentially & gracefully
   homepage: https://github.com/mnrgreg/kubectl-node-restart
   caveats: |
-    Execution of this plugin requires Kubernetes cluster-admin Rolebindings.
+    Execution of this plugin requires Kubernetes cluster-admin Rolebindings
+    and the ability to schedule Privileged Pods.
   description: |
     This plugin performs a sequential, rolling restart of selected nodes by first
-    Draining each node, then running a Kubernetes Job to reboot each node, and
-    finally Uncordoning each node when Ready. 
+    draining each node, then running a Kubernetes Job to reboot each node, and
+    finally uncordoning each node when Ready. 
--- a/plugins/node-restart.yaml
+++ b/plugins/node-restart.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: node-restart
+spec:
+  version: "v0.0.2"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/MnrGreg/kubectl-node-restart/releases/download/v0.0.2/v0.0.2.zip
+    sha256: "44db0eee9603085652837d26c5352287864f158b557f78ae19325724640535f8"
+    files:
+    - from: "*.sh"
+      to:  "."
+    - from: "LICENSE"
+      to: "."
+    bin: "node-restart.sh"
+  shortDescription: >-
+    Restart Kubernetes Nodes sequentially & gracefully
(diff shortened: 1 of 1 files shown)


plugins/node-restart.yaml
```diff
@@ -0,0 +1,27 @@
...
+    files:
+    - from: "*.sh"
+      to:  "."
+    - from: "LICENSE"
+      to: "."
+    bin: "node-restart.sh"
+  shortDescription: >-
+    Restart Kubernetes Nodes sequentially & gracefully
+  homepage: https://github.com/mnrgreg/kubectl-node-restart
+  caveats: |
```
corneliusweig: You could also mention that you need to be allowed to run privileged pods.

plugins/node-restart.yaml
```diff
@@ -0,0 +1,27 @@
...
+    uri: https://github.com/MnrGreg/kubectl-node-restart/releases/download/v0.0.2/v0.0.2.zip
+    sha256: "44db0eee9603085652837d26c5352287864f158b557f78ae19325724640535f8"
+    files:
+    - from: "*.sh"
+      to:  "."
+    - from: "LICENSE"
+      to: "."
+    bin: "node-restart.sh"
+  shortDescription: >-
+    Restart Kubernetes Nodes sequentially & gracefully
```
corneliusweig:
```suggestion
    Restart cluster nodes sequentially & gracefully
```
Kubernetes is not necessary, because this is `kubectl` :)

plugins/node-restart.yaml
```diff
@@ -0,0 +1,27 @@
...
+  homepage: https://github.com/mnrgreg/kubectl-node-restart
+  caveats: |
+    Execution of this plugin requires Kubernetes cluster-admin Rolebindings.
+  description: |
+    This plugin performs a sequential, rolling restart of selected nodes by first
+    Draining each node, then running a Kubernetes Job to reboot each node, and
+    finally Uncordoning each node when Ready.
```
corneliusweig:
```suggestion
    finally uncordoning each node when Ready. 
```

plugins/node-restart.yaml
```diff
@@ -0,0 +1,27 @@
...
+      to: "."
+    bin: "node-restart.sh"
+  shortDescription: >-
+    Restart Kubernetes Nodes sequentially & gracefully
+  homepage: https://github.com/mnrgreg/kubectl-node-restart
+  caveats: |
+    Execution of this plugin requires Kubernetes cluster-admin Rolebindings.
+  description: |
+    This plugin performs a sequential, rolling restart of selected nodes by first
+    Draining each node, then running a Kubernetes Job to reboot each node, and
```
corneliusweig:
```suggestion
    draining each node, then running a Kubernetes Job to reboot each node, and
```

plugins/node-restart.yaml
MnrGreg: added

plugins/node-restart.yaml
MnrGreg: revised

plugins/node-restart.yaml
MnrGreg: revised

plugins/node-restart.yaml
MnrGreg: revised

MnrGreg:
@ahmetb could you review this when you get a gap?

/assign @ahmetb

plugins/node-restart.yaml
```diff
@@ -0,0 +1,28 @@
...
+    uri: https://github.com/MnrGreg/kubectl-node-restart/releases/download/v0.0.2/v0.0.2.zip
+    sha256: "44db0eee9603085652837d26c5352287864f158b557f78ae19325724640535f8"
+    files:
+    - from: "*.sh"
+      to:  "."
+    - from: "LICENSE"
+      to: "."
+    bin: "node-restart.sh"
+  shortDescription: >-
+    Restart cluster nodes sequentially & gracefully
```
ahmetb: ideally we'd use `and` instead of `&`.
but you can fix this later.

ahmetb:
/lgtm
/approve
but left a note.

PR #500, opened 2020-02-17
Update 'plugins.md'.

bmcstdio:


corneliusweig:
Thanks! Looks like we should automate this at some point :)
/lgtm
/approve

PR #502, opened 2020-02-18
Add manifest for "hotfix" plugin

steadysupply:
Closes #492 (and follows on from it)
Documentation at https://steady.supply/git/kubectl-hotfix
Source code available with `git clone https://steady.supply/git/kubectl-hotfix`

### DIFF
--- a/plugins/hotfix.yaml
+++ b/plugins/hotfix.yaml
@@ -1,0 +1,75 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hotfix
+spec:
+  version: "v0.0.3"
+  shortDescription: dumb file transfer to your labelled pods
+  homepage: https://steady.supply/git/kubectl-hotfix
+  description: |
+
+      hotfix
+
+      dumb file sync/transfer to your labelled pods
+
+      command line options are -v, etc for glog, and -c to specify a configuration
+      file, eg.
+
+        namespace: xyz
+        ignore:
+          - "*.tmp"
+          - "*.{html,css,js}"
+        include:
+          - "*.py"
+        shell: ash
+        path: src:/app
+        target:
+          - label: service in (web, api)
+          - label: service=postgres
+            path: src/database:/app/database
+          - label: service=chrome
+            shell: bash
+          - label: service=static
+            include:
+              - "*.{html,css,js}"
+
+      which specifes defaults for the namespace to work in, the transfer paths (in
+      standard local:remote format), the globs to be ignore’d / include’d and the
+      shell offered by the default container in matched pods. this configuration will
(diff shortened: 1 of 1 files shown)


plugins/hotfix.yaml
```diff
@@ -0,0 +1,75 @@
...
+        path: src:/app
+        target:
+          - label: service in (web, api)
+          - label: service=postgres
+            path: src/database:/app/database
+          - label: service=chrome
+            shell: bash
+          - label: service=static
+            include:
+              - "*.{html,css,js}"
```
corneliusweig: This should be moved into a dedicated help for this command like `hotfix --help`. The description is meant to explain _what_ a plugin does and not how to configure it.

plugins/hotfix.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hotfix
```
corneliusweig: `hotfix` is not very descriptive. I think your name should mention `sync` or `transfer`, like your short description does.

plugins/hotfix.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+      uri: https://steady.supply/git/kubectl-hotfix/release/v0.0.3/kubectl-hotfix-v0.0.3-darwin-amd64.tar.gz
+      sha256: "8a879f5256630866bfe033bae70b66256065c817c193c3e31ca9dd0aa78530bc"
+      files:
+        - from: "*"
+          to:  "."
```
corneliusweig: This is the default so you can also omit the copy instructions.

plugins/hotfix.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hotfix
+spec:
+  version: "v0.0.3"
+  shortDescription: dumb file transfer to your labelled pods
+  homepage: https://steady.supply/git/kubectl-hotfix
+  description: |
```
corneliusweig: This description reads more like a man page.

Please focus on what your plugin does and why a user would need it. The details you give here should be moved to the commands `--help` output.

plugins/hotfix.yaml
steadysupply: Thanks

steadysupply: Wait, I can omit `to: "."` or the whole `files` key?

ahmetb: Whole files key.

corneliusweig:
@steadysupply What's your opinion on the above points? If you don't intend to follow up, I'd like to close this issue.

PR #521, opened 2020-03-02
Adding popeye to the krew...

derailed:
Hi - here is a proposal to add [popeye](https://popeyecli.io) as a new member of this awesome krew! 
Hopefully we're kool?
 
Thank you for your assistance in making this a reality!

### DIFF
--- a/plugins/popeye.yaml
+++ b/plugins/popeye.yaml
@@ -4,7 +4,7 @@ metadata:
   name: popeye
 spec:
   homepage: https://popeyecli.io
-  shortDescription: 👀 A Kubernetes cluster resource sanitizer
+  shortDescription: Scans your clusters for potential resource issues.
   version: v0.7.1
   description: |
     Popeye is a utility that scans live Kubernetes clusters and reports
--- a/plugins/popeye.yaml
+++ b/plugins/popeye.yaml
@@ -4,7 +4,7 @@ metadata:
   name: popeye
 spec:
   homepage: https://popeyecli.io
-  shortDescription: Scans your clusters for potential resource issues.
+  shortDescription: Scans your clusters for potential resource issues
   version: v0.7.1
   description: |
     Popeye is a utility that scans live Kubernetes clusters and reports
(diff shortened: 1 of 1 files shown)


plugins/popeye.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: popeye
+spec:
+  homepage: https://popeyecli.io
+  shortDescription: 👀 A Kubernetes cluster resource sanitizer
```
ahmetb: Let's avoid emojis as they aren't available everywhere. (e.g. windows terminals).

Also you repeat `Kubernetes` word which is obvious in this context. 

I'll help you rewrite this:

> Scan cluster and resources for potential issues

ahmetb:
Right now the only issue I have is the usage string doesn't show kubectl when invoked as plugins:

```
k popeye -h
Popeye scans your Kubernetes clusters and reports potential resource issues.

Usage:
  popeye [flags]
  popeye [command]
```

We tend to advise developers to change that when the binary is run as a plugin.

plugins/popeye.yaml
corneliusweig: > Let's avoid emojis as they aren't available everywhere.

Just a question: would we accept ascii art?

corneliusweig:
The manifest looks good to me. Pretty sweet.

To me the proper command name in the help text is certainly good, but not a blocker to getting merged. However, if @derailed can fix that before it's merged that would be even better :)

plugins/popeye.yaml
ahmetb: Not in shortDescription since it’s meant to be one-line. 

For description, also probably not as we might show it on the website someday.

derailed:
@ahmetb @corneliusweig  Thank you both for reviewing this! 

I will change the short description. As for the help message, I'll have to rev popeye to make it so ;(

plugins/popeye.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: popeye
 spec:
   homepage: https://popeyecli.io
-  shortDescription: 👀 A Kubernetes cluster resource sanitizer
+  shortDescription: Scans your clusters for potential resource issues.
```
ahmetb: do you mind removing the trailing period
sorry for the back and forth.

derailed:
@ahmetb Tx!!

PR #522, opened 2020-03-03
Add ssh plugin

temal-:
Hi,

I learned about krew today via [this blog post](https://medium.com/@marko.luksa/a-consistent-provider-agnostic-way-to-ssh-into-any-kubernetes-node-99328372d09f) and wanted to contribute this plugin so that others can also use this useful helper.

Thanks!

### DIFF
--- a/plugins/ssh.yaml
+++ b/plugins/ssh.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ssh
+spec:
+  homepage: https://github.com/luksa/kubectl-plugins#kubectl-ssh-node
+  shortDescription: Provider-agnostic way of opening a remote shell to a node
+  version: v0.1.0
+  description: |
+    Enables you to access a node even when it doesn't run an SSH server
+    or when you don't have the required credentials. Also, the way you log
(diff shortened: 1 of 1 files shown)


corneliusweig:
Hey, thanks for bringing this up! As far as I can tell, this is very similar to https://github.com/kvaps/kubectl-node-shell, which is available as the `node-shell` plugin in krew. Does `ssh` solve a problem that is not solved by node-shell which I'm missing right now?

Otherwise I'd argue that krew-index does not need a second tool for the same job. `ssh` is also a pretty generic name. The main focus of the plugin seems to be to get a root shell on a node, so this should be part of the name.

ahmetb:
+we have the `sshd` plugin as well.

temal-:
> Hey, thanks for bringing this up!

Thanks for taking the time to review this!

> As far as I can tell, this is very similar to https://github.com/kvaps/kubectl-node-shell, which is available as the node-shell plugin in krew.

You're absolutely right. I only searched for `ssh` but not for shell. My fault, sorry.

> Does ssh solve a problem that is not solved by node-shell which I'm missing right now?

Not that I know of. The implementation is a bit different, but other than that I see no difference.

> Otherwise I'd argue that krew-index does not need a second tool for the same job.

I'd argue the same. 👍 

> +we have the sshd plugin as well.

Which does something completely different, afaik.

Thanks for your time!

PR #523, opened 2020-03-04
Add tmux-exec to index

predatorray:
This plugin provides the ability to operate (eg, `exec bash`) multiple pods by using Tmux.


corneliusweig:
I'll merge as-is, because my previous comment is purely on the plugin side and has nothing to do with the manifest. Nice job! :confetti_ball: 

/lgtm
/approve

predatorray:
@corneliusweig , thanks for your advice! In fact, I think there are no cases when Tmux window will be opened without tty. I will make them turned on by default in the next release. Thanks again!

PR #532, opened 2020-03-09
Add sql plugin

yaacov:
Add [kubectl-sql](https://github.com/yaacov/kubectl-sql) plugin.

### DIFF
--- a/plugins/sql.yaml
+++ b/plugins/sql.yaml
@@ -12,16 +12,7 @@ spec:
         arch: amd64
     uri: https://github.com/yaacov/kubectl-sql/releases/download/v0.2.11/kubectl-sql.tar.gz
     sha256: 34b42e057cb7bf4d09340469d18ea7ecaca6f109d893a3fc5782dad84c8a0c94
-    files:
-    - from: "*"
-      to: "."
     bin: kubectl-sql
-  shortDescription: Use SQL like language to query the Kubernetes cluster manager.
+  shortDescription: Query the cluster via pseudo-SQL
   description: |
-    This plugin use SQL like language to query the Kubernetes cluster manager.
-  caveats: |
-    Usage:
-      $ kubectl sql
-    For additional options:
-      $ kubectl sql --help
-      or https://github.com/yaacov/kubectl-sql
+    This plugin enables to query the kubernetes API server via a SQL like language.
--- a/plugins/sql.yaml
+++ b/plugins/sql.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: sql
+spec:
+  version: v0.2.11
+  homepage: https://github.com/yaacov/kubectl-sql
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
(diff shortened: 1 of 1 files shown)


plugins/sql.yaml
```diff
@@ -0,0 +1,27 @@
...
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/yaacov/kubectl-sql/releases/download/v0.2.11/kubectl-sql.tar.gz
+    sha256: 34b42e057cb7bf4d09340469d18ea7ecaca6f109d893a3fc5782dad84c8a0c94
+    files:
+    - from: "*"
+      to: "."
+    bin: kubectl-sql
+  shortDescription: Use SQL like language to query the Kubernetes cluster manager.
```
corneliusweig: It's most important to be very concise here, because the short description will be truncated at ~50 chars.
```suggestion
  shortDescription: Query the cluster via pseudo-SQL
```

plugins/sql.yaml
```diff
@@ -0,0 +1,27 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/yaacov/kubectl-sql/releases/download/v0.2.11/kubectl-sql.tar.gz
+    sha256: 34b42e057cb7bf4d09340469d18ea7ecaca6f109d893a3fc5782dad84c8a0c94
+    files:
+    - from: "*"
+      to: "."
```
corneliusweig: This is the default, so you can drop the `files` spec altogether.

plugins/sql.yaml
```diff
@@ -0,0 +1,27 @@
...
+    bin: kubectl-sql
+  shortDescription: Use SQL like language to query the Kubernetes cluster manager.
+  description: |
+    This plugin use SQL like language to query the Kubernetes cluster manager.
```
corneliusweig:
```suggestion
    This plugin enables to query the kubernetes API server via a SQL like language.
```

plugins/sql.yaml
```diff
@@ -0,0 +1,27 @@
...
+  caveats: |
+    Usage:
+      $ kubectl sql
```
corneliusweig: Basic invocation is now printed after installation by default. So no need to do it explicitly.

plugins/sql.yaml
```diff
@@ -0,0 +1,27 @@
...
+    For additional options:
+      $ kubectl sql --help
+      or https://github.com/yaacov/kubectl-sql
```
corneliusweig: Homepage is also shown by default, so no need to do it explicitly.

yaacov:
@corneliusweig thanks for the review :+1: 
updated the descriptions and remove the unneeded default fields.

ahmetb:
> The name make sense, but is pretty generic. I think it's ok to use `sql`.
> @ahmetb, WDYT?

I think the name is ok. There's been similar tools in the space, like GQL or something I forgot. But it's a not-so-crowded space anyway.

corneliusweig:
/lgtm
/approve
Thanks for your contribution!
In case you haven't seen it: there is tooling to automate releases to krew-index via GitHub actions (see  https://krew.sigs.k8s.io/docs/developer-guide/release/updating-plugins.md/)

PR #541, opened 2020-03-12
Adding wls plugin

mohatb:

PR #543, opened 2020-03-13
Adding wls.yaml

mohatb:

### DIFF
--- a/plugins/wls.yaml
+++ b/plugins/wls.yaml
@@ -19,7 +19,7 @@ spec:
         - darwin
         - linux
     uri: https://github.com/mohatb/kubectl-wls/archive/v1.0.0.tar.gz
-    sha256: "43e571371da3ae71ed804c7baf78f1ddcc35127e3375d815245034511d8913b5"
+    sha256: "70fd0e8f7e9c31801e887d1ce698c36601d09c0bcceb5b66c5cbb751909d936a"
     files:
       - from: "./*/kubectl-wls"
         to: .
--- a/plugins/wls.yaml
+++ b/plugins/wls.yaml
@@ -1,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: wls
+spec:
+  version: "v1.0.0"
+  homepage: https://github.com/mohatb/kubectl-wls
+  shortDescription: Windows-linux-shell allows you to start a shell on windows or linux nodes
+  description: |
+    kubectl-wls it is an interactive shell script that allows you to access a kubernetes windows or linux node shell.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/mohatb/kubectl-wls/archive/v1.0.0.tar.gz
+    sha256: "43e571371da3ae71ed804c7baf78f1ddcc35127e3375d815245034511d8913b5"
+    files:
+      - from: "./*/kubectl-wls"
+        to: .
+      - from: "./*/LICENSE"
+        to: .
+    bin: "kubectl-wls"
--- a/plugins/wls.yaml
+++ b/plugins/wls.yaml
@@ -5,9 +5,11 @@ metadata:
 spec:
   version: "v1.0.0"
   homepage: https://github.com/mohatb/kubectl-wls
-  shortDescription: Windows-linux-shell allows you to start a shell on windows or linux nodes
+  shortDescription: simple and interactive plugin to access a shell on windows or linux nodes
   description: |
-    kubectl-wls it is an interactive shell script that allows you to access a kubernetes windows or linux node shell.
+    kubectl interactive plugin to access a shell on windows or Linux nodes.
+    For Linux nodes: It works by creating a pod (with a privileged container) in the node you specified and using nsenter for getting a shell into your Kubernetes nodes.
+    For windows nodes: It creates a pod in any available Linux node, and it uses ssh to access windows nodes that have ssh enabled.
   platforms:
   - selector:
       matchExpressions:


mohatb:
@ahmetb I am not sure why the other pull was closed. I've created another one here and tested locally without any issues.

This is different than the other tools, but a little similar to nodeshell for linux as it is a fork from kubectl-enter script.

> it has an interactive option, which runs kubectl get nodes and prompts the user to chose from the node node name.
> it mainly supports accessing cmd/powershell, by creating a container on a linux node, and use it a jump box to windows nodes that has OpenSSH installed by default as of AKS VMs.

ahmetb:
It seems like you closed it, let's not do that again.
That loses context.
/cc @corneliusweig 
There was some discussion at #541 to keep in mind while evaluating.

plugins/wls.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: wls
+spec:
+  version: "v1.0.0"
+  homepage: https://github.com/mohatb/kubectl-wls
+  shortDescription: simple and interactive plugin to access a shell on windows or linux nodes
+  description: |
+    kubectl interactive plugin to access a shell on windows or Linux nodes.
+    For Linux nodes: It works by creating a pod (with a privileged container) in the node you specified and using nsenter for getting a shell into your Kubernetes nodes.
+    For windows nodes: It creates a pod in any available Linux node, and it uses ssh to access windows nodes that have ssh enabled.
```
ahmetb: 80 line char wraps please

plugins/wls.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: wls
+spec:
+  version: "v1.0.0"
+  homepage: https://github.com/mohatb/kubectl-wls
+  shortDescription: simple and interactive plugin to access a shell on windows or linux nodes
```
ahmetb: Too long. Cap at 50 chars.

> simple and interactive plugin

almost nobody cares about this part.

mohatb:
So what is required from my side now ? Just yo update the description and short description to 50 chars ? Anything else ?

ahmetb:
Another round of reviews will be done by another maintainer.

ahmetb:
+1 That's what I mentioned in other issue.


> it has an interactive option, which runs kubectl get nodes and prompts the user to chose from the node node name.

This is not really a differentiator.

> it mainly supports accessing cmd/powershell, by creating a container on a linux node, and use it a jump box to windows nodes that has OpenSSH installed by default as of AKS VMs.

Sounds like this plugin is primarily created for "windows" nodes, so why support Linux when there are already plugins that support Linux?

I recommend refining and reducing scope.

mohatb:
Why to have a tool for windows and another for linux when you can access both in the same one ?

mohatb:
Also, for "interactive" since you said "no body cares" I guess that is wrong, as other tools requires you to remember or copy the node name before running the tool, and if you did not, you need to go back and figure out the node name. In cloud environments usually the node name is to long and unmemorable

ahmetb:
> Also, for "interactive" since you said "no body cares" I guess that is wrong, 

I meant it’s not relevant in short description. Try to shorten it with interactive in it. You don’t need to say “plugin” in short description. It’s implied. 

> when you can access both in the same one ?

We try not to be prescriptive about these. 
However, in this case this is a plugin that hugely overlaps with other plugins (which is something we usually reject). But you have a unique handle (windows) but the rest of the plugin is probably not as great as plugins purely focused on Linux I assume. Hope this makes sense.

mohatb:
ok so I guess we should close this one.

ahmetb:
@pduchnovsky all the plugins you listed were among some of the first plugins we accepted (and are planning to remove), therefore they're from a time when no criteria were in place. 

We're always happy to re-evaluate and we do accept plugins with similar functionality if they are better or have a differentiation.

mohatb:
Thanks @ahmetb @pduchnovsky - Well I think they might all lead to the same results, however, wls might be easier to use based on the feedback from many people that are using it. specially for hosted k8s as the name of the node is very long, it gives the option to select the node number instead of copy/paste the node name which requires you to go back to another terminal to get the name. I am happy to resubmit this if this as it would be easier for many people to use directly from krew.

ahmetb:
I have tried the plugin and I would say its functionality is nearly identical to `node-shell` plugin.

`wls` didn't work for me out of the box.

1. I had to explicitly chmod+x, but we can look past that.
2. You mentioned it makes it easy to deal with long node names, but the version I installed from this manifest doesn't have node numbers or substring match.
3. Error handling of wls is not great, edge cases seem rough:

        Error from server (NotFound): nodes "tul5" not found
        /Users/ahmetb/.krew/bin/kubectl-wls: line 118: [: =: unary operator expected

4. When I entered a node name, didn't work, as it used the wrong kubectl argument:

		    Enter the node name: gke-integration-tests-pool-1-cfcd1dab-tul5
		    creating pod "gke-integration-tests-pool-1-cfcd1dab-tul5-exec-k4ets2" on node "gke-integration-tests-pool-1-cfcd1dab-tul5"
		    Error: unknown flag: --generator
		    See 'kubectl run --help' for usage.

On the flip side, `node-shell` plugin works much better today, and it already has 500+ stars on GitHub, which makes me think there's a proven user base there:

```
~ $ k node-shell
Please specify node name
~ $ k node-shell gke-integration-tests-pool-1-cfcd1dab-tul5
spawning "nsenter-nzinc2" on "gke-integration-tests-pool-1-cfcd1dab-tul5"
If you don't see a command prompt, try pressing enter.
gke-integration-tests-pool-1-cfcd1dab-tul5 / #
```

The functionality in your plugin seems 95% identical to an already very-popular plugin in the ecosystem. Is there a reason why you don't contribute there?

If your value-add (versus `node-shell`) is the Windows support, and an easier way to choose nodes (which I have not observed in my hands-on trial), perhaps you can submit these features to `node-shell` (which is _also_ a bash-script based plugin). If you have trouble convincing its maintainer, I will do my best to help there.

pduchnovsky:
@ahmetb 95% ?
How did you come up with such a number ?

node-shell can only
- connect to a linux node that you know name of

wls can:
- connect to a linux node that you know name of
- connect to a linux node that you don't know name of interactively
- connect to a windows node that you know name of
- connect to a windows node that you don't know name of interactively
- mount pod to copy data between local machine and a node
- be called much quicker, `k wls` vs `k node-shell`

mohatb:
Well, am also not sure which tool did you test, as I do not even have a print statement with **"Enter the node name"**. I agree that node-shell has a strong community behind it but I guess we might get the same as I also have a good number of clones on both kubectl-exec [which was originally created but renamed to wls so it does not conflict with exec command] https://github.com/mohatb/kubectl-exec and kubectl-wls, am not saying that we need to replace anything, but more of having the option for the people to choose. you may try kubectl-exec and there is a video to walk you through how it works. in your example, you've also used a gke node with a very long name your process will be as the following:

(Simplicity) - You run the command, if you did not copy the node name already, you have to cancel the command, run kubectl get nodes, copy the node name, and run the tool again. or, to open a new terminal window, and run kubectl get nodes  and copy the node name (this is where we solve it by having an interactive where you can choose a node number 0, 1, 2 ..etc. and this will also avoid typos if you already memorizing the long node name.

In addition, there is also a huge adoption to windows containers, and people depends on this a lot to access powershell in windows containers. instead of having to create a jump server/pod ..etc

PR #180, opened 2019-07-16
New plugin : role-summary

nlamirault:
**Checklist for plugin developers:**

- [x] Read the [Plugin Naming Guide](https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md) (for new plugins)
- [x] Verify the installation from URL or a local archive works (`kubectl krew install --manifest=[...] --archive=[...]`)

PR #546, opened 2020-03-17
Add rolesum(ex-bindrole) plugin to index

Ladicle:
Re-Opened the PR to add a bindrole plugin.
The changes from last time are as follow:
- Reduced .spec.shortDescription less than 50 characters.
- Add the LICENSE file to kubectl-bindrole repository.
- Change the help message to make it easier to understand, and updated [README](https://github.com/Ladicle/kubectl-bindrole#usage).

ref: https://github.com/kubernetes-sigs/krew-index/pull/180
ref: https://github.com/Ladicle/kubectl-bindrole/issues/6


plugins/bindrole.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: bindrole
+spec:
+  shortDescription: Summarize RBAC roles tied to the given subject
```
ahmetb: “roles for subjects” work?

plugins/bindrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+  homepage: https://github.com/Ladicle/kubectl-bindrole
+  version: v1.3.0
+  platforms:
+    - uri: https://github.com/Ladicle/kubectl-bindrole/releases/download/v1.3.0/kubectl-bindrole_linux-amd64.tar.gz
+      sha256: "40e012d2687df0d31cd68bc6abaf2fe6e47de0a3d66a10b6800ae3c56c787ced"
+      bin: kubectl-bindrole
+      files:
+        - from: ./kubectl-bindrole_linux-amd64/kubectl-bindrole
```
ahmetb: Please also copy out LICENSE

plugins/bindrole.yaml
Ladicle: yes, it works. fix: 81af304

plugins/bindrole.yaml
Ladicle: copied. fix: 81af304

plugins/bindrole.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: bindrole
```
corneliusweig: I think we discussed before (https://github.com/kubernetes-sigs/krew-index/pull/180) that `bindrole` is not an ideal name for this plugin. As far as I understand, this plugin shows an overview over roles for a given subject. The current name suggests that the plugin is used to _change_ the roles.

The last suggestion from #180 was `role-summary`. WDYT about that?

ahmetb: +1 that was also the outcome of https://github.com/Ladicle/kubectl-bindrole/issues/6

Ladicle: I know `bindrole` is not an ideal name, but I'd rather not rename because it'd confuse current users. If it needs to change, I prefer the short name so I'd like to make it `sumrole`. Is that ok?

corneliusweig: Creative names are ok. That said, I recommend to choose a descriptive name, because many users will not run the command on a daily basis but once in a while. And it is easier to recognise a descriptive plugin name than a cryptic one.
However, in the space of acceptable plugin names, the choice is yours.

For your concrete suggestion, I would question what does it mean to "sum" a role? It's not at all what you intended, but users may come with similar questions.

ahmetb: `rolesum`?

Ladicle: sounds nice! I'll change it to that name.

Ladicle: fix: 07a3827

corneliusweig:
I think you should also add a `description` field. This is your chance to sell your plugin to potential users who might have a use for it. Don't waste this chance :wink:

Ladicle:
Thanks, @corneliusweig! Added description in ae1d35c.

plugins/rolesum.yaml
```diff
@@ -0,0 +1,58 @@
...
+      selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+    - uri: https://github.com/Ladicle/kubectl-rolesum/releases/download/v1.4.0/kubectl-rolesum_windows-amd64.tar.gz
+      sha256: "deb3a8eab654cd95ce93c246aa6b570fa41ece1a4f321404d037aef2534565db"
+      bin: kubectl-rolesum
+      files:
+        - from: ./kubectl-rolesum_windows-amd64/kubectl-rolesum
+          to: "."
```
ahmetb: I am guessing you didn't test this but this won't work.

For something to be executable on windows, it needs an .exe prefix. (you need to add it to `to:` and `bin:` fields).

Ladicle: fix: 1b6a46f

PR #559, opened 2020-03-28
Add 'kubectl-cilium'.

bmcstdio:
This PR adds the first version of `kubectl-cilium`, a plugin meant to provide easier interaction with [Cilium](https://cilium.io) agents running on a Kubernetes cluster.

### DIFF
--- a/plugins/cilium.yaml
+++ b/plugins/cilium.yaml
@@ -1,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: cilium
+spec:
+  version: v0.1.0
+  shortDescription: Easily interact with Cilium agents.
+  homepage: https://github.com/bmcstdio/kubectl-cilium
+  description: |
+    Easily interact with Cilium agents.
+    Useful, for example, to open a shell on the Cilium agent running on a
+    particular node or co-located with a particular pod:
+
+    # Open a shell on the Cilium agent running on node 'kind-cilium-mesh-worker':
+    $ kubectl cilium exec kind-cilium-mesh-worker
+
+    # Run 'cilium monitor' on the Cilium agent co-located with the 'ns-1/nginx' pod:
+    $ kubectl cilium exec ns-1/nginx cilium monitor
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/bmcstdio/kubectl-cilium/releases/download/v0.1.0/kubectl-cilium_v0.1.0_darwin_amd64.tar.gz
+    sha256: "c8466c94337db34f0cee5be60a0715682bfabf5b5dc041d1a30b4cd9eaae0b11"
+    files:
+    - from: kubectl-cilium
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-cilium
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/bmcstdio/kubectl-cilium/releases/download/v0.1.0/kubectl-cilium_v0.1.0_linux_amd64.tar.gz
+    sha256: "4173139333fd1e8993dcfdc747ae11b54ccd123ae9668296bdb67c3d90521f0b"
+    files:
+    - from: kubectl-cilium
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-cilium
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/bmcstdio/kubectl-cilium/releases/download/v0.1.0/kubectl-cilium_v0.1.0_windows_amd64.tar.gz
+    sha256: "801d66d8fc06c04ac8f7ea0a243c1a742166eef8e1c97caf3ec322f30de92a38"
+    files:
+    - from: kubectl-cilium.exe
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-cilium.exe


ahmetb:
Hi, per naming guidelines, I’m assuming Cilium is a trademarked name. Do you work at/represent cilium?

tgraf:
> Hi, per naming guidelines, I’m assuming Cilium is a trademarked name. Do you work at/represent Cilium at official capacity?

Hi, speaking as one of the Cilium maintainers, Bruno has asked us for permission and we are absolutely happy with the naming of the plugin.

plugins/cilium.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: cilium
```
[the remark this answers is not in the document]
tgraf: Hi, speaking as one of the Cilium maintainers, Bruno has asked us for permission. We are excited about this and Bruno has offered to make this part of the Cilium project as well. We have no problem with this.

plugins/cilium.yaml
```diff
@@ -0,0 +1,54 @@
...
+spec:
+  version: v0.1.0
+  shortDescription: Easily interact with Cilium agents running on a Kubernetes cluster.
+  homepage: https://github.com/bmcstdio/kubectl-cilium
+  description: |
+    Easily interact with Cilium agents running on a Kubernetes cluster.
+    Useful, for example, to open a shell on the Cilium agent running on a particular node or co-located with a particular pod:
+
+    # Open a shell on the Cilium agent running on node 'kind-cilium-mesh-worker':
+    $ kubectl cilium exec gke-gke-1-p-1-50120dfc-26gm
```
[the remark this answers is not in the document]
bmcstdio: That was a mistake, which I've amended in the meantime. :+1:

bmcstdio:
> The main purpose of the tool is to more easily select the pod to kubectl exec into. Is that correct?

No, that's not correct.

> There are already a few plugins in that niche. Can you tell us what is so different about yours?

The plugin is supposed to evolve into much more than that, and if you look at the [repository](https://github.com/bmcstdio/kubectl-cilium) and at the description, it's defined as a plugin for easier _interaction_ with Cilium, which may (and will) encompass much more than just picking a pod to exec into.

bmcstdio:
@tgraf thank you very much for your support! :+1:

corneliusweig:
> The plugin is supposed to evolve into much more than that, and if you look at the [repository](https://github.com/bmcstdio/kubectl-cilium) and at the description, it's defined as a plugin for easier _interaction_ with Cilium, which may (and will) encompass much more than just picking a pod to exec into.

I can only assess what the plugin already does, since I don't know your vision for the plugin. So far it only does exec and there are already other plugins to help with that. That's why I'm reluctant to merge just now. To me it looks like its current functionality is quite generic (non-Cilium specific) and there are various tools for the same task.
So I suggest to extend its functionality with cilium specific tasks before merging to krew-index. Until then, we can put this PR on hold to not lose context.

---

As to naming: Plugin names cannot be transferred to other implementations. So the name `cilium` will not be available to any future alternatives. @tgraf if this is ok for the Cilium project then we can proceed with this name. However, please be aware that using the project name will also look like an official plugin. So you should consider if the proper location for this plugin is really the cilium github org (i.e. `github.com/cilium/kubectl-cilium`).

/hold

ahmetb:
It seems like the plugin is on a course to potentially be owned by Cilium. So I don't have a problem with that part anymore. Very soon we'll be allowing people to self-host their index repos, so Cilium can run a small version of this repo if they need.

My current concern is the same as Cornelius: it seems like the plugin only does "kubectl exec". What we strive to avoid (in the central index) is that leaving users confused. We can merge with the promise of adding more functionality soon –however, if that doesn't happen, as you might appreciate the plugin will look odd there, so we might remove it (which is not something we've done before).

bmcstdio:
> So far it only does exec and there are already other plugins to help with that. (...) To me it looks like its current functionality is quite generic (non-Cilium specific) and there are various tools for the same task.

I wonder if there's one that takes the name of a node or a `namespace/name` key targeting a pod and proceeds to (1) discovering in what namespace Cilium is installed, (2) discovering what Cilium pod is running on that node/alongside that pod and (3) execing into that particular Cilium pod. It seems to me like you're too focused on the "exec" part, which is just the end result. From a quick look into the Krew index, all plugins mentioning "exec" would require some additional interaction from the user instead of just taking a node name / pod coordinates. In any case, I am beyond arguing at this point. 🙃

> We can merge with the promise of adding more functionality soon –however, if that doesn't happen, as you might appreciate the plugin will look odd there, so we might remove it (which is not something we've done before).

IMHO this sounds the most reasonable of all. I wasn't expecting such an entry barrier, TBH, but again I won't argue—will only appreciate if this can be merged as it currently stands. 🙂

plugins/cilium.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: cilium
+spec:
+  version: v0.1.0
+  shortDescription: Easily interact with Cilium agents running on a Kubernetes cluster.
```
corneliusweig:
```suggestion
  shortDescription: Easily interact with Cilium agents
```
kubernetes is redundant and everything above ~50chars is usually truncated.

plugins/cilium.yaml
```diff
@@ -0,0 +1,54 @@
...
+  homepage: https://github.com/bmcstdio/kubectl-cilium
+  description: |
+    Easily interact with Cilium agents running on a Kubernetes cluster.
+    Useful, for example, to open a shell on the Cilium agent running on a particular node or co-located with a particular pod:
```
corneliusweig: Can you warp at ~80 chars so this prints nicely regardless of screen size?

bmcstdio:
> With your additional context, I'm convinced that this already adds unique value to cilium users and that we should merge. Thanks for your endurance in explaining this

Awesome, glad to hear that! :) I've addressed your feedback, thank you for that.

plugins/cilium.yaml
corneliusweig: Did you mean to shorten the `shortDescription`? AFAICT, it's unchanged.
Feel free to reword it such that it makes most sense to you.

bmcstdio: Sure I did, but I had forgotten. Thanks @corneliusweig!

bmcstdio:
@corneliusweig @ahmetb I believe everything's OK now. Let me know if anything else is necessary before merging, and thank you for your time!

corneliusweig:
/lgtm
/approve
Thanks for your contribution! Looking forward to see how this will grow in the future.

PR #562, opened 2020-04-01
new plugin: roll

branttaylor:
Don't be too hard on me! I was able to install my plugin locally, but I don't think it linked correctly. Let me know if anything is very obvious that I've done wrong, I'm a little new at this.

### DIFF
--- a/plugins/roll.yaml
+++ b/plugins/roll.yaml
@@ -4,11 +4,15 @@ metadata:
   name: roll
 spec:
   homepage: https://github.com/LifeWay/kubectl-roll-plugin
-  shortDescription: Slowly delete pods by namespace
-  version: v1.0.2
+  shortDescription: Rolling restart of all persistent pods in a namespace
+  version: v1.0.4
   description: |
-    A utility to slowly delete pods by namespace with
-    a set amount of seconds between each deleted resource.
+    A utility to slowly delete all pods in a namespace that are
+    controlled by Deployments, StatefulSets, and DaemonSets. You must
+    correctly configure readiness probes to avoid any downtime.
+
+    Usage:
+      kubectl roll -n <namespace>
   platforms:
   - selector:
       matchExpressions:
@@ -17,9 +21,9 @@ spec:
         values:
         - darwin
         - linux
-    uri: https://github.com/LifeWay/kubectl-roll-plugin/archive/1.0.2.tar.gz
-    sha256: 9022434c06e3e782fb1809ae36ea48c77ab982b3480b48c233d3320df8ccb8df
+    uri: https://github.com/LifeWay/kubectl-roll-plugin/archive/1.0.4.tar.gz
+    sha256: d0e408fefdbc078410f4021644d9b03236be1f215aaaba79e2e786ef3ae10344
     bin: roll
     files:
-    - from: kubectl-roll-plugin-1.0.2/kubectl-roll
+    - from: kubectl-roll-plugin-1.0.4/kubectl-roll
       to: roll
--- a/plugins/roll.yaml
+++ b/plugins/roll.yaml
@@ -1,0 +1,22 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: roll
+spec:
+  homepage: https://github.com/LifeWay/kubectl-roll-plugin
+  shortDescription: Slowly delete pods by namespace or label
+  version: v0.1.0
+  description: |
+    A utility to slowly delete pods by namespace or by label with
+    a set amount of seconds between each pod deletion.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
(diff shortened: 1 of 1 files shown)


ahmetb:
Don’t link to master.(zip/tar.gz). It will break **immediately** as you push to that repo. 

Please stop  and read the pre-release checklist https://krew.sigs.k8s.io/docs/developer-guide/release/new-plugin/

branttaylor:
@ahmetb Sorry about that. This is fixed. I'm working through the CLA process with my organization now.

branttaylor:
@corneliusweig Definitely fair points, and if this plugin is not suitable for krew, I'll withdraw it.
The plugin absolutely assumes that you have properly deployed the pods via a deployment, replication controller, daemonset, or some other mechanism to restart the pods once deleted. 

Maybe it would help to describe some situations where we've used this plugin internally:
- A namespace's default pod security policy was changed
- A namespace's default pod resource limits (`limitranges`) were changed (memory and/or cpu)
- A namespace's default pod priority class was changed

There may be more, but all of the above scenarios only trigger a change to a pod _once started_, but does not apply to any running pods. In these cases, we used the plugin to slowly restart pods in a namespace in order to limit any potential downtime. Does that help explain the reasoning? The `kubectl rollout restart` command does restart pods but with a different scope (e.g. deployment vs. namespace or label), so this plugin was more ideal for the scenarios I listed above.

To your point about the delay, for us internally we unfortunately have a lot of Java-based pods which take an eternity to start up, so we wanted a configurable way to restart pods in a really slow way if needed, and even better in a way that our devs would not even notice. There are definitely some fair arguments to be made that pods that take so long to start up need some work to fix that, and I agree, but this let us work around our dev's "issues".

Let me know what you think!

corneliusweig:
Hi, sorry for the slow response. I see that your command addresses an issue where you need to restart everything in a _namespace_ as opposed to a _deployment_. My major concern was about the way this restart is performed, though. Your plugin does this with a hard delay and does not take readiness probes into account. This makes it slow and potentially dangerous. I suggest to work on that before merging to krew-index.
A nice way to do this would be to base your logic on replicaset instead of the pod level, and use `kubectl rollout restart` instead of `kubectl delete pod` to do the restarts. This would miss standalone pods, but your current version just kills those (which is questionable). WDYT?

branttaylor:
@corneliusweig I think you're right, that's a good change to make. `kubectl rollout restart` didn't exist when I originally wrote this last year. Thanks for the suggestion!

branttaylor:
@corneliusweig I rewrote this to target Deployments, StatefulSets, and DaemonSets instead of pods. Please let me know what you think.

plugins/roll.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: Plugin
+metadata:
+  name: roll
+spec:
+  homepage: https://github.com/LifeWay/kubectl-roll-plugin
+  shortDescription: Slowly delete pods by namespace
+  version: v1.0.2
+  description: |
+    A utility to slowly delete pods by namespace with
+    a set amount of seconds between each deleted resource.
```
corneliusweig, line 11: You should mention that this triggers a rolling update of Deployments, ReplicaSets, and StatefulSets. In particular, you need to point out that users must set up readiness probes and use the rolling-update strategy to avoid downtimes.

Btw, waiting between each deleted resource is not doing what it suggests.

branttaylor:
@corneliusweig Thank you, great suggestions! In it's current form, it will quickly run through scheduling restarts on the named resources. It's not as good as running each in parallel, but it's still very fast. I will plan on rewriting part of this to make it faster, but it won't make a huge difference unless there are a large amount of resources in the namespace. As an example, on a test cluster, it ran through scheduling restarts on 5 deployments in 1 second.

What are your thoughts on how it looks now?

plugins/roll.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: roll
+spec:
+  homepage: https://github.com/LifeWay/kubectl-roll-plugin
+  shortDescription: Rolling restart of all persistent pods in a namespace
+  version: v1.0.4
+  description: |
+    A utility to slowly delete all pods in a namespace that are
+    controlled by Deployments, StatefulSets, and DaemonSets. You must
+    correctly configure readiness probes to avoid any downtime.
```
corneliusweig, line 12: I'm not sure if this is implicit or obvious, but the rollout strategy also plays a role in avoiding downtime. If it's set to `recreate`, there will be downtime, only `RollingUpdate` is safe (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy).

corneliusweig:
Hey @branttaylor,
sorry for the delay. This looks quite nice now. Maybe you can incorporate a small warning about picking the right update strategy in the next manifest iteration?

Thanks for all your additional effort that you put in!
/lgtm
/approve

PR #563, opened 2020-04-01
new plugin: std-json

fleeto:
Export Kubernetes object in JSON format with a clean and sorted style. Remove some auto-generated fields and then sort fields in it.

### DIFF
--- a/plugins/std-json.yaml
+++ b/plugins/std-json.yaml
@@ -7,7 +7,7 @@ spec:
   shortDescription: Export Kubernetes object in JSON format with a clean and sorted style.
   version: v0.1.0
   description: Export Kubernetes object in JSON format, remove some auto generated fields and then sort fields in it.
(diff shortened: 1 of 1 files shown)


ahmetb:
Hi, thanks for your submission. This seems quite overlapping with existing `-o=json` flag for kubectl get + [eksporter](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/eksporter.yaml) plugin which we already have. When we run into this situation, we look for significant differentiation before accepting very similar plugins.

corneliusweig:
There are even two plugins in that niche: https://github.com/itaysk/kubectl-neat

Please let us know how your plugin is different.

fleeto:
neat
The first one is neat, and the second one is eksporter. 

But you are right, there is no significant difference, it's just simple and correct, just for me.

PR #568, opened 2020-04-05
adding new plugin 'context'

stefanm8:

### DIFF
--- a/plugins/context.yaml
+++ b/plugins/context.yaml
@@ -1,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: context
+spec:
+  version: "v0.0.1"
+  homepage: https://github.com/stefanm8/kubectl-select-context
+  shortDescription: "Simpler way of viewing and selecting contexts"
+  description: |
+    Displays contexts, when given a argument it searches across contexts
+    and switch to the found context.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: "os"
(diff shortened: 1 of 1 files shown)


stefanm8:
I signed the CLA

ahmetb:
Thanks for submitting this plugin. Since we already have a plugin called ctx that does nearly the same functionality, it’s very likely we’ll turn this down. Can you please compare the two?

stefanm8:
Thanks for looking over it.

The main difference is that it matches a substring in the name of the context and displays the information as kubectl config get-contexts does, it just applies color to the current context.

corneliusweig:
The `ctx` plugin already highlights the current context and it allows to select a context by fuzzy matching via fzf. So your plugin does not really add anything that's not already there. Or did I miss something?

There is precedent that we turn down plugins that are too similar to other plugins that are already in the index (https://github.com/kubernetes-sigs/krew-index/pull/429, https://github.com/kubernetes-sigs/krew-index/pull/351). Especially in the realm of switching context, there is no added value of having several similar alternatives in the index.
My recommendation: if you find that `ctx` is missing something, open an issue there and submit a PR if you get good feedback.

ahmetb:
> The main difference is that it matches a substring in the name of the context and displays the information as kubectl config get-contexts does, it just applies color to the current context.

ctx plugin supports all of those when fzf is installed.

stefanm8:
Fair enough, I was not familiar with fzf, doesn't look like it can match a substring but a prompt serves pretty well. 

Thanks for commentaries

PR #616, opened 2020-05-18
Add volume-reclaim plugins

dirathea:
# Summary

Kubectl plugins to gather all PVC and check whether it used in any workloads on the cluster or not.
This could help users to remove unnecessary PVCs that lead to reducing unecesary cost by unused persistent disk

### DIFF
--- a/plugins/volume-reclaim.yaml
+++ b/plugins/volume-reclaim.yaml
@@ -1,0 +1,56 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: volume-reclaim
+spec:
+  version: "v0.1.0"
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/dirathea/kubectl-volume-reclaim/releases/download/v0.1.0/kubectl-volume-reclaim_linux_amd64.tar.gz
+    sha256: "64fd81c6115312980b8bba9b9036e34b7608ca2f3d64e5cb0c143ce56b39ee41"
+    files:
+    - from: "./volume-reclaim"
+      to: "."
+    - from: LICENSE
+      to: "."
+    bin: "volume-reclaim"
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/dirathea/kubectl-volume-reclaim/releases/download/v0.1.0/kubectl-volume-reclaim_darwin_amd64.tar.gz
+    sha256: "a84b614b62c4410b33a7c6d6dd8442c3020c95f4edb5c379000fe87c3ff8187b"
+    files:
+    - from: "./volume-reclaim"
+      to: "."
+    - from: LICENSE
+      to: "."
+    bin: "volume-reclaim"
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/dirathea/kubectl-volume-reclaim/releases/download/v0.1.0/kubectl-volume-reclaim_windows_amd64.zip
+    sha256: "3e3351447c40e378bf73b0c0361aa1b8b63842fb3b6a5f15ab2442f9649ea7a9"
+    files:
+    - from: "/volume-reclaim.exe"
+      to: "."
+    - from: LICENSE
+      to: "."
+    bin: "volume-reclaim.exe"
+  shortDescription: Gather all PVC and output all pvcs that doesn't belong to any workloads.
+  homepage: https://github.com/dirathea/kubectl-volume-reclaim
+  caveats: |
+    Usage:
+      $ kubectl volume-reclaim
+
+    For additional options:
+      $ kubectl volume-reclaim --help
+      or https://github.com/dirathea/kubectl-volume-reclaim/blob/v0.1.0/doc/USAGE.md
+
+  description: |
+    Kubectl plugins to gather all PVC and check whether it used in any workloads on cluster or not.
+    This could help users to remove unecessary PVCs that lead to reducing unecesary cost by unused persistent disk


dirathea:
Already sign up on linuxfoundation.

corneliusweig:
Hey, thanks for submitting!
Can you comment how your plugin is different from https://github.com/yashbhutwala/kubectl-df-pv#kubectl-df-pv? As far as I can tell, there is considerable overlap and we try to not have multiple implementations for the same idea.

dirathea:
The main purpose of this plugin is to determine whether the PVC is referred by any Kubernetes resources. For example, when creating new StatefulSet, it automatically creates new PVC, PV, and Persistent Disk on the provider.
But when StatefulSet is deleted, PVC still around and so the PV and Persistent Disk. Listing PVC only describes the status between PV and PVC, but not the user of PVC. This makes confusion on whether to delete unused PVC or not.

The plugin that you refer to is about free space of PVC, while this check whether a PVC is used by Deployment, StatefulSet, DaemonSet, or even Job

dirathea:
is it clear enough about the differences @corneliusweig ?

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: volume-reclaim
```
corneliusweig, line 4:
```suggestion
  name: unused-volumes
```
`volume-reclaim` sounds like it actually does reclaim something. But it only lists volumes, right?

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+        arch: amd64
+    uri: https://github.com/dirathea/kubectl-volume-reclaim/releases/download/v0.1.0/kubectl-volume-reclaim_windows_amd64.zip
+    sha256: "3e3351447c40e378bf73b0c0361aa1b8b63842fb3b6a5f15ab2442f9649ea7a9"
+    files:
+    - from: "/volume-reclaim.exe"
+      to: "."
+    - from: LICENSE
+      to: "."
+    bin: "volume-reclaim.exe"
+  shortDescription: Gather all PVC and output all pvcs that doesn't belong to any workloads.
```
corneliusweig, line 44: Try to keep your short description below 50 characters. Everything above this will usually be truncuated.

What about `List unused PVCs`?

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+  homepage: https://github.com/dirathea/kubectl-volume-reclaim
+  caveats: |
+    Usage:
+      $ kubectl volume-reclaim
+
```
corneliusweig, line 49:
```suggestion
```
Newer `krew` versions print instructions how to run the plugin out of the box. So this is obsolet.

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+    For additional options:
+      $ kubectl volume-reclaim --help
```
corneliusweig, line 51: Usually the `caveats` section is about essential information that you need _before_ trying out the plugin. Like a note about required dependencies. If you are absolutely attached to print info about getting help, you can do this, but I wouldn't say it's necessary.

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+      or https://github.com/dirathea/kubectl-volume-reclaim/blob/v0.1.0/doc/USAGE.md
```
corneliusweig, line 52:
```suggestion
      or https://github.com/dirathea/kubectl-volume-reclaim/blob/master/doc/USAGE.md
```
Please link to the master version instead. I understand why you want to direct to the exact (and correct) usage file, but this will prevent automatic version bumps for your plugin.

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  description: |
+    Kubectl plugins to gather all PVC and check whether it used in any workloads on cluster or not.
+    This could help users to remove unecessary PVCs that lead to reducing unecesary cost by unused persistent disk
```
corneliusweig, line 56:
```suggestion
    This could help users to remove unecessary PVCs that lead to reducing unecesary cost by unused persistent volume
```
Persistent disk is GCP specific, right?

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+  caveats: |
+    Usage:
+      $ kubectl volume-reclaim
+
+    For additional options:
+      $ kubectl volume-reclaim --help
+      or https://github.com/dirathea/kubectl-volume-reclaim/blob/v0.1.0/doc/USAGE.md
+
+  description: |
+    Kubectl plugins to gather all PVC and check whether it used in any workloads on cluster or not.
```
corneliusweig, line 55: There's an inaccurate promise here. It does not evaluate _all_ workloads, but only a subset (https://github.com/dirathea/kubectl-volume-reclaim/blob/master/pkg/plugin/plugin.go#L39-L42). I think it's okay to just consider these workloads, but you should be specific here as well. In particular, the plugin homepage doesn't talk about this either.

corneliusweig, line 55: Just a suggestion, feel  free to use or reword:
```suggestion
    Kubectl plugins to gather all PVC and check whether it used in any workloads on cluster or not.
    This plugin lists all PVCs that are not used by any
      - DaemonSet
      - Deployment
      - Job
      - StatefulSet
    Thus it helps to find unused PersitentVolumes which just cost money.
```

plugins/volume-reclaim.yaml
```diff
@@ -0,0 +1,56 @@
...
+  homepage: https://github.com/dirathea/kubectl-volume-reclaim
+  caveats: |
+    Usage:
+      $ kubectl volume-reclaim
+
+    For additional options:
+      $ kubectl volume-reclaim --help
+      or https://github.com/dirathea/kubectl-volume-reclaim/blob/v0.1.0/doc/USAGE.md
+
+  description: |
```
corneliusweig, line 54: Please keep line length below 80 chars, so that it prints nicely on wide and narrow terminals.

dirathea:
please check the updates @corneliusweig

PR #627, opened 2020-05-28
Add SchemaHero to krew index

marccampbell:
This PR adds [SchemaHero](https://schemahero.io) to the krew index.

SchemaHero is preparing to release a 0.9.0 version that will include (hopefully) a krew installer and CLI to approve and reject migrations (https://schemahero.io/cli/install/). This is the prerelease version (it does work, just an alpha release of this version). 

Because this is the first version distributed in krew, we are starting with an alpha of the upcoming release, but will not be distributing alpha versions this way in the future.

- [x] Make sure you read the Plugin Naming Guide
- [x] Verify you can install your plugin locally

### DIFF
--- a/plugins/schemahero.yaml
+++ b/plugins/schemahero.yaml
@@ -1,0 +1,60 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: schemahero
+spec:
+  version: v0.9.0-alpha.1
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/schemahero/schemahero/releases/download/v0.9.0-alpha.1/kubectl-schemahero_linux_amd64.tar.gz
+    sha256: 78119f420f10fd9cc5d0d341cdc0404aea4ccc6d8da8646a0a34826e71d4dcd6
+    files:
+    - from: kubectl-schemahero
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-schemahero
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/schemahero/schemahero/releases/download/v0.9.0-alpha.1/kubectl-schemahero_darwin_amd64.tar.gz
+    sha256: 79144272c3dc396b7b950bbd2a3ddd9045c49b248577dd00de7e1ed034b0f2ad
+    files:
+    - from: kubectl-schemahero
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-schemahero
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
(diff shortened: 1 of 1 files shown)


plugins/schemahero.yaml
```diff
@@ -0,0 +1,60 @@
...
+    files:
+    - from: kubectl-schemahero.exe
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-schemahero.exe
+  shortDescription: Cloud native, declarative database schema migrations
+  homepage: https://schemahero.io
+  caveats: |
+    Installation:
```
corneliusweig, line 47: Can you be a bit more explicit that this is a step that comes after installation via krew? Something like

> Schemahero requires a cluster-side operator. To install it, run:

plugins/schemahero.yaml
```diff
@@ -0,0 +1,60 @@
...
+      $ kubectl schemahero install
+
+    Documentation is available at:
+      https://schemahero.io
```
corneliusweig, line 51: The homepage is automatically shown after installs nowadays. So no need to put it here.

plugins/schemahero.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+
+  description: |
+    SchemaHero is an open-source database schema migration tool that converts 
+    a schema definition into migration scripts that can be applied in any 
+    environment. Written as both a CLI utility and a Kubernetes Operator, 
+    SchemaHero eliminates the task of creating and managing sequenced 
+    migration scripts that are compatible with all environments that an 
+    application is running in.
```
corneliusweig, line 60: This sounds a bit like a marketing pitch. Can you give more details how the plugin works? In particular, what do you mean by "all environments"?

plugins/schemahero.yaml
```diff
@@ -0,0 +1,60 @@
...
+        arch: amd64
+    uri: https://github.com/schemahero/schemahero/releases/download/v0.9.0-alpha.1/kubectl-schemahero_windows_amd64.tar.gz
+    sha256: 5a293788484762fbc1c7d27737e40a39db9f7f162b482b651516c06a5d3a6479
+    files:
+    - from: kubectl-schemahero.exe
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-schemahero.exe
+  shortDescription: Cloud native, declarative database schema migrations
```
corneliusweig, line 44: IIUC, it currently only supports postgres. This should be mentioned somewhere.

corneliusweig, line 44:
```suggestion
  shortDescription: Declarative database schema migrations via YAMLs
```
It's unclear what cloud native even means. So I'd recommend to be more descriptive here.

marccampbell:
Thanks @corneliusweig. All good points...

I've updated the manifest based on your comments.

ahmetb:
/lgtm
/approve
Thanks for your submission as usual!

PR #631, opened 2020-05-29
Add kubectl diagnose and podevents

alecjacobs5401:
Adds `kubectl podevents` and `kubectl diagnose` plugins.

Both of these binaries exist in the same repository, which is why they share download URIs

### DIFF
--- a/plugins/diagnose.yaml
+++ b/plugins/diagnose.yaml
@@ -1,0 +1,51 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Kubernetes Pods that are "Not Ready" (that have failing Pod Conditions or Containers)
+  description: |
+    This plugin shows finds and displays debugging information for Pods that are "Not Ready" in the current namespace.
+    In addition, you can filter which pods you want to show based on labels or field selectors (as well as by pod name(s)).
+    Use --help to see all supported flags.
+  caveats: |
+    * The -A/--all-namespaces flag is currently not supported
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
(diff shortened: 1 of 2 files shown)


ahmetb:
Can you please submit separate PRs as the decisions and discussions for two plugins are independent.

ahmetb:
Ideally we also strongly recommend splitting repos. That way you can release and revision them independently. For example when you set up release automation to update one, you can have only one of them update but not the other. 

It’s not a hard requirement, but you’ll find out it makes the life easier in the long run for your users to read docs, report issues etc as well.

alecjacobs5401:
@ahmetb Sounds good, thanks for looking!

>  That way you can release and revision them independently. For example when you set up release automation to update one, you can have only one of them update but not the other.

Yeah, I struggled with this too but both implementations use pretty much exactly the same code, so I didn't see the benefit of splitting repos and duplicating code (or turning duplicated code into _another_ third repo for DRYness)

I'll keep it in mind for the future though.

plugins/diagnose.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Kubernetes Pods that are "Not Ready" (that have failing Pod Conditions or Containers)
```
ahmetb, line 8: This is too long. Cap at 50-60 chars.

plugins/podevents.yaml
```diff
@@ -0,0 +1,51 @@
...
+metadata:
+  name: podevents
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Show events for pods
+  description: |
+    This plugin shows events for all pods in the current namespace. In addition, you can filter which pods to show
+    events for as well as filter which events you want to show.
+    Use --help to see all supported flags.
```
ahmetb, line 12: No need to say —help as it is implied.

PR #632, opened 2020-05-29
add sick-pods plugin

alecjacobs5401:
Adds `kubectl diagnose` plugin.

This has been split into its own branch as recommended in https://github.com/kubernetes-sigs/krew-index/pull/631

### DIFF
--- a/plugins/diagnose.yaml
+++ b/plugins/diagnose.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   version: v0.1.1
   homepage: https://github.com/alecjacobs5401/kubectl-diagnose
-  shortDescription: Find and debug Kubernetes Pods that are "Not Ready"
+  shortDescription: Find and debug Pods that are "Not Ready"
   description: |
     This plugin shows finds and displays debugging information for Pods that are "Not Ready" in the current namespace.
     In addition, you can filter which pods you want to show based on labels or field selectors (as well as by pod name(s)).
--- a/plugins/diagnose.yaml
+++ b/plugins/diagnose.yaml
@@ -1,0 +1,51 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Kubernetes Pods that are "Not Ready" (that have failing Pod Conditions or Containers)
+  description: |
+    This plugin shows finds and displays debugging information for Pods that are "Not Ready" in the current namespace.
+    In addition, you can filter which pods you want to show based on labels or field selectors (as well as by pod name(s)).
+    Use --help to see all supported flags.
+  caveats: |
+    * The -A/--all-namespaces flag is currently not supported
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/alecjacobs5401/kubectl-diagnose/releases/download/v0.1.1/kubectl-diagnose_0.1.1_darwin_amd64.tar.gz
+    sha256: a61ab7faa2b73db1cd52397330c1bccaec0ccf1909c3db059377749a248007bb
+    files:
+    - from: LICENSE
+      to: .
+    - from: kubectl-diagnose
+      to: .
+    bin: kubectl-diagnose
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/alecjacobs5401/kubectl-diagnose/releases/download/v0.1.1/kubectl-diagnose_0.1.1_linux_amd64.tar.gz
+    sha256: 2cf1a20ff8607a3e729aa86b67188beb9aab1146da532b03be03fc9841860249
+    files:
+    - from: LICENSE
+      to: .
+    - from: kubectl-diagnose
+      to: .
+    bin: kubectl-diagnose
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/alecjacobs5401/kubectl-diagnose/releases/download/v0.1.1/kubectl-diagnose_0.1.1_windows_amd64.tar.gz
+    sha256: deb96ab4acf683073b70ef9d2f3a259212ba3c48ba712c064a9e3d3980a82fdf
+    files:
+    - from: LICENSE
+      to: .
+    - from: kubectl-diagnose.exe
+      to: .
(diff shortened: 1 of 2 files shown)


plugins/diagnose.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Kubernetes Pods that are "Not Ready"
+  description: |
```
ahmetb, line 9: please wrap each line at 80 chars

plugins/diagnose.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Kubernetes Pods that are "Not Ready"
```
ahmetb, line 8: no need to say "Kubernetes", in kubectl context, Pod is specific enough.

plugins/diagnose.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
```
ahmetb, line 7: off-topic: I recommend you list some example outputs so people can see how it works without installing.

alecjacobs5401: Thats a good suggestion, forgot to update my README with that!

Will do

plugins/diagnose.yaml
alecjacobs5401: 🤦 right lol. Thanks!

plugins/diagnose.yaml
alecjacobs5401: [Added!](https://github.com/alecjacobs5401/kubectl-diagnose/blob/master/README.md#example-usages)

plugins/diagnose.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: diagnose
```
ahmetb, line 4: This `diagnose` name is sadly too generic.
It violates our "first comers advantage" rule in the Naming Guide.

As you might already see, we already have plenty of plugins in this repo that provide some form of "diagnosing" capabilities or another.

The exception to this has been the `debug` plugin which has now graduated into kubectl core. Other than that, we tend to reject names that put dibs on an entire class of verbs or resources. (e.g. `kubectl statefulset`, `kubectl service`, `kubectl inspect`...)

alecjacobs5401: @ahmetb 
Thats totally fair.

Would something like `kubectl poddiagnose` or `kubectl pod-diagnose` better fit? I can just rename it in the manifest to copy over as one of those

ahmetb: Ideally we like dash between these words but there's precedent for both.
Pick and choose what’s better for your users, our Naming Guide recommends dashes in between.
I'll let @corneliusweig also chime in the proposed names.

plugins/diagnose.yaml
```diff
@@ -0,0 +1,54 @@
...
+spec:
+  version: v0.1.1
+  homepage: https://github.com/alecjacobs5401/kubectl-diagnose
+  shortDescription: Find and debug Pods that are "Not Ready"
+  description: |
+    This plugin shows finds and displays debugging information for Pods
```
corneliusweig, line 10:
```suggestion
    This plugin finds and displays debugging information for Pods
```

plugins/diagnose.yaml
corneliusweig: Yeah, `diagnose` is a bit too generic. `pod-diagnose` is certainly an option, but diagnose sounds more interactive to me. What about `sick-pods`? To me that sounds more like it would just find list unhealthy pods (which your plugin does, AFAIU).

plugins/diagnose.yaml
```diff
@@ -0,0 +1,54 @@
...
+    that are "Not Ready" in the current namespace.
+    Pods are deemed as "Not Ready" if they have any failing Pod Conditions
+    or have any containers that do not have a "Ready" value.
+    In addition, you can filter which pods you want to show based on labels
+    or field selectors (as well as by pod name(s)).
+  caveats: |
+    * The -A/--all-namespaces flag is currently not supported
```
corneliusweig, line 17: Please use `k8s.io/cli-runtime` to talk to the cluster. This will take care of all the standard flags.

plugins/diagnose.yaml
alecjacobs5401: I like `sick-pods`! That fits the theme and does make a lot of sense. I'll change it to that 👍

plugins/diagnose.yaml
alecjacobs5401: If possible, I'd like to push this to a later release version. The `podevents` plugin that was already approved and merged had the same caveat, so I'd like to take the time to update both to adhere to better standards, but still get this `sick-pods` in.

alecjacobs5401:
Renamed `diagnose` to `sick-pods`

ahmetb:
> This has been split into its own branch as recommended in #631

long-term I recommend you really consider splitting by repo.
branches aren't really for maintaining totally separate codebases.
but that's off topic for krew-index.

ahmetb:
@alecjacobs5401 Great job with this plugin.
I've just tried it and it works like a champ! I have several other points to mention

1. Your --help message doesn't print the right plugin name currently:

      kubectl sick-pods -h
      usage: kubectl diagnose [<flags>] [<pod>...]

  Hope you can fix this in an upcoming release.

2. Hope you can integrate with our release automation via GitHub Workflows https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/

/lgtm
/approve

alecjacobs5401:
@ahmetb @corneliusweig 

Thanks a bunch to you both! Excited to be contributing back!

>Your --help message doesn't print the right plugin name currently

Oof, yeah great point. I'm definitely planning on taking the suggestion to use the `k8s.io/cli-runtime`, which would probably include an overall rename so I can take the time to fix it there as well.

>Hope you can integrate with our release automation via GitHub Workflows

Absolutely!

PR #628, opened 2020-05-28
Add gadget plugin

mauriciovasquezbernal:
[Inspektor Gadget](https://github.com/kinvolk/inspektor-gadget) is a collection of tools (gadgets) for debugging and
introspecting Kubernetes applications.

More info is at https://github.com/kinvolk/inspektor-gadget.

- [x] Make sure you read the Plugin Naming Guide: https://sigs.k8s.io/krew/docs/NAMING_GUIDE.md
- [x] Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...]

PR #634, opened 2020-05-29
gadget: Update description

mauriciovasquezbernal:


ahmetb:

-----

This pull request **does not** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/gadget.yaml is not a straightforward version bump: could not find the old/new version spec in the diff

corneliusweig:
I guess this is in response to https://github.com/kubernetes-sigs/krew-index/pull/628#issuecomment-636200418.

@ahmetb I suggested this wording, because `app` sounds somewhat implicit to me. Besides, the list of subcommands from here https://github.com/kinvolk/inspektor-gadget#how-to-use does not suggest that it's very app-specific, but rather a general swiss-army knife toolset. WDYT?

ahmetb:
It seems like it’s pretty process-oriented so I’m good with this. 👍🏼

mauriciovasquezbernal:
I'd say that it's container-oriented. Even if the "app" word is somehow implicit, "Gadgets for debugging and introspecting" sound like is missing something.

PR #643, opened 2020-06-10
Add sample plugin

seredot:
This PR adds `example` plugin to Krew. The plugin prints out an example resource file for a given resource type. 

```
$ kubectl example deploy
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.14.2
          ports:
            - containerPort: 80
```


plugins/sample.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: sample
+spec:
+  homepage: https://github.com/seredot/kubectl-sample
+  shortDescription: Prints out sample manifest YAMLs
+  version: v1.0.0
+  description: |
+    Prints out a sample manifest YAML for a given resource type.
```
corneliusweig, line 10: Can you add something like

> Similar to `kubectl create deploy --dry-run` but for arbitrary resources.

seredot: I added it to the description.

plugins/sample.yaml
```diff
@@ -0,0 +1,78 @@
...
+kind: Plugin
+metadata:
+  name: sample
+spec:
+  homepage: https://github.com/seredot/kubectl-sample
+  shortDescription: Prints out sample manifest YAMLs
+  version: v1.0.0
+  description: |
+    Prints out a sample manifest YAML for a given resource type. 
+    The plugin works similar to `kubectl create deployment nginx --image=nginx --dry-run=client -o yaml` but for arbitrary resources.
```
ahmetb, line 11: I'm not sure if there's a need to list that here + breaks the 80-char linewrap.

ahmetb:
I think I'm open to this. I've been hoping to develop something like this myself for a while. As for criticism, naming, `example` could be better, but `sample` will do just fine. (Just gotta keep in mind that sampling is also a verb, that can be used for other reasons like perf benchmarking).

I hope you can extend this project and make it modular so people can contribute Istio, Knative, and many other CRDs too.

plugins/sample.yaml
seredot: It is now adjusted.

seredot:
> I think I'm open to this. I've been hoping to develop something like this myself for a while. As for criticism, naming, `example` could be better, but `sample` will do just fine. (Just gotta keep in mind that sampling is also a verb, that can be used for other reasons like perf benchmarking).
> 
> I hope you can extend this project and make it modular so people can contribute Istio, Knative, and many other CRDs too.

That's a good idea. I'm open for contributions.

ahmetb:
> I'm open for contributions.

It's more like you’re reserving a pretty generic word on Krew-index, so you better make good use of it. :)

plugins/sample.yaml
```diff
@@ -0,0 +1,78 @@
...
+  version: v1.0.0
+  description: |
+    Prints out a sample manifest YAML for a given resource type. 
+    The plugin works similar to a dry run output but for arbitrary resources.
+    All samples originate from kubernetes.io.
+
+    Usage:
+      kubectl sample deployment
+      kubectl sample Pod
+      kubectl sample svc
```
ahmetb, line 17: no need for this; --help should show it.

corneliusweig:
I'm a bit worried about the generality and imprecision of `sample`. It could be many things:
* sample manifest
* collect requests samples
* collect statistics
and probably many more.

I think it's fine to have the plugin. But would you be open to rename to `example`? I think that would overall be a better name.

seredot:
> I think it's fine to have the plugin. But would you be open to rename to `example`? I think that would overall be a better name.

Sure, why not. I'll rename it and update the PR.

corneliusweig:
@seredot Thanks!
/hold

seredot:
The project name is changed to 'example'. @ahmetb @corneliusweig

ahmetb:
/hold cancel
/lgtm
/approve
Hope you set up Github actions in your repo to automate releases: https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/.

PR #646, opened 2020-06-12
Add node-rolling plugin

yogeek:
This PR aims to add [node-rolling](https://github.com/yogeek/kubectl-node-rolling) plugin to krew index.

As explained in the README of the plugin directory, I had a need similar to the [node-restart](https://github.com/MnrGreg/kubectl-node-restart) plugin but in a context where a restart was not enough to update the node : as they were managed by an AWS ASG, I needed to shutdown the nodes to allow the ASG to replace them by a new version (new AMI in Launch Template).

I originally thought of a more generic version where the command to execute (restart, shutdown, etc.) would be a parameter of the plugin but I did not see use cases for other command and this would imply the deprecation of the original plugin... but I am open to suggestions to ease the user experience !

### DIFF
--- a/plugins/node-rolling.yaml
+++ b/plugins/node-rolling.yaml
@@ -1,0 +1,29 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: node-rolling
+spec:
+  version: "v1.0.0"
+  platforms:
+  - selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+    uri: https://github.com/yogeek/kubectl-node-rolling/releases/download/v1.0.0/v1.0.0.zip
+    sha256: "9e9e22724016724765419354509f9238f34ec1f7bb6dc9eeb55af248a559bf93"
+    files:
+    - from: "*.sh"
+      to:  "."
+    - from: "LICENSE"
+      to: "."
+    bin: "node-rolling.sh"
+  shortDescription: >-
+    Shutdown cluster nodes sequentially & gracefully
+  homepage: https://github.com/yogeek/kubectl-node-rolling
+  caveats: |
+    Execution of this plugin requires Kubernetes cluster-admin Rolebindings
+    and the ability to schedule Privileged Pods.
+  description: |
+    This plugin performs a sequential, rolling shutdown of selected nodes by first
+    draining each node, then running a Kubernetes Job to shutdown each node, and
+    finally waiting for the node to be replaced by a new one 
+    by an external component (like an AutoScaling Group). 


yogeek:
I signed it

ahmetb:
I'll let @corneliusweig decide this one.

corneliusweig:
@yogeek Your project description sounds a lot like it is specific to AWS. Is that correct?

yogeek:
@corneliusweig I currently use a K8S cluster where nodes are in AWS AutoScaling Group but I don't think this kind of need is specific to AWS.
All nodes managed by a central component (it can be AWS  ASG but also GCP MIG or other instance group manager in other situations) needs a sort of instance template that define their configuration. And updating a node in this case needs the update of this template (update of the OS image, of the cloud-init script,etc.) with infrastructure as code and then the termination of the node so the controller can boot another one with the new template.
Consequently, a shutdown is in this case a way to make the controller detecting that the instances is not here anymore and replacing it with an updated node.

corneliusweig:
@yogeek Ok, I think this use-case is a good one. However, thinking about our users, it is less helpful to have many plugins that solve very similar (albeit different) problems, than having _one_ plugin which does it's job well for various variants.
So I wonder, if your use-case could be integrated into the `node-restart` plugin. It looks feasible, given that they share most of their code. WDYT?
cc @MnrGreg

yogeek:
@corneliusweig Fair point ! (that's why I mentionned the possibility to make a more generic plugin in my first comment on this PR ;) )

On the other hand, making the script generic could lead to a more complex code because I used a technique to discover the new node (the one started by the ASG to replace the shutdowned one) that may not be needed in the restart use-case... 

Indeed, it would be interesting to have @MnrGreg in the loop :)

MnrGreg:
Having an optional input parameter that captures a command that must be executed, prior to reboot, is a use-case that many could benefit from. For example, applying an OS security patch prior to reboot. It's also simple to implement with little side affects.

The function that checks for a different node name, however may have less use-cases outside of the ASG.

Out of interest, is the rolling deployment of nodes, not better solved for you @yogeek with something like ClusterAPI?

yogeek:
@MnrGreg Do you mean `a command to be executed prior to reboot` or `a command to be executed` (and this command would be either "restart", or "reboot" or any command needed on the node, for example "update && reboot"...)
Maybe the second option would be more generic ? 
But, in this case, it may be interesting to include all the command logic as a script which would be passed to the plugin as a parameter (and so my function to check for a different node name would be included in this script for example) but it does not seem very easy to do... (it would imply sending the script into the node in order to execute it... ?)
It reminds me of the behaviour of https://github.com/rancher/system-upgrade-controller project but using a kubectl plugin is much more simple ! :D 

About your question regarding CLusterAPI, we do not use it today to manage our clusters because it did not exist at the time we had to develop our deployment ^^ But I looked a bit into it for the past few weeks and I find it very interesting, and wish to test an adapatation of our current deployment towards it.

One difference between our current deployment and ClusterAPI is that :
- our current deployment is based on an immutable pattern, where all the needed elements for a component (master, nodes, external etcd...) are included in an pre-baked AMI, firstly deployed with Terraform which configures this AMI as part of a Launch Template that is used by an ASG to manage instances for this component. The main logic of the cluster deployment is based on kubadm which is used in instances cloud-init scripts to form (or join) a cluster at boot time. The logic is then managed from within the cluster itself which is quite challenging and that explains my initial need for this kind of plugin
- ClusterAPI is based on a "management cluster" which allows to manage the states of "workload clusters" externally

MnrGreg:
I was thinking of having a place holder for any command and then still perform the `&& reboot`.  So leave it open to the admin to decide what they want to execute but it could be as simple as `yum -y update` or more complicated such as executing an Ansible pull from git `ansible-pull -d /tmp/ansible -U https://github.com/repo/ansiblepull.git`

ClusterAPI does rely on the Management Cluster. The Workload Cluster is workable in an immutable form. One can build ones own AMIs and can modify the default ClusterAPI templates to create ones own Flavors in which one can inject custom cloud-init configs and files. This is all maintained in the Management Cluster.

yogeek:
Ok so in the "terminate instance" use case, the final command will be `shutdown && reboot` ...?

MnrGreg:
Yup

corneliusweig:
Is there an agreement between you whether the functionality of `node-rolling` can and should be integrated into `node-restart`? From a user perspective, I think that _one_ more powerful tool would be better, but it depends on you if this is feasible.
/hold

MnrGreg:
@yogeek would the ability to specify an input command, that's executed prior to reboot, help your use-case?

yogeek:
@MnrGreg indeed I think this solution would be better for everyone
The only thing is : what do we do about my function to discover the new node name ...? It is mandatory in a case of an ASG as the new node name is unknown, but I think it can be applied also in your 'restart' case and will return the same node name, don't you think ?

MnrGreg:
@yogeek PR #772 submitted which takes a pre-restart command as input.

corneliusweig:
I haven't followed the latest discussion thoroughly. Is there a common understanding now, that the functionality can be included in `node-restart`? If so, could you please close this issue?

yogeek:
I still have the questions of my previous comment unanswered : 
"what do we do about my function to discover the new node name that is needed in the case of a node replacement to check its readiness ?"

If you have any idea about making this with your pre-start command, I will be happy to close this issue ! ☺️

corneliusweig:
For that I have too little experience with the kind of setup this plugin is targeting. Maybe @MnrGreg has an idea?

ahmetb:
Closing due to inactivity.
/close

PR #658, opened 2020-06-19
Add kubectl trace v0.1.0-rc.1

fntlnz:
Happy to submit kubectl trace to the krew index as per the discussions happening one year ago  in https://github.com/iovisor/kubectl-trace/issues/17

Sorry if it took a long time!

Adding a wip since I didn't test the manifest on all platforms yet.

If anyone can test it on windows before I do and report here I'll be grateful.

Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/plugins/trace.yaml
+++ b/plugins/trace.yaml
@@ -19,7 +19,7 @@ metadata:
 spec:
   version: "v0.3.4"
   homepage: https://github.com/iovisor/kubectl-trace
-  shortDescription: bpftrace programs on the kubectl
+  shortDescription: bpftrace programs in a cluster
   description: |
     kubectl trace is now installed! 
 
--- a/plugins/trace.yaml
+++ b/plugins/trace.yaml
@@ -1,0 +1,80 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: trace
+spec:
+  version: "v0.3.4"
+  homepage: https://github.com/iovisor/kubectl-trace
+  shortDescription: bpftrace programs on the kubectl
+  description: |
+    kubectl trace is now installed! 
+
+    Examples:
+
+      # Count system calls using tracepoints on a specific node
+      kubectl trace run node/kubernetes-node-emt8.c.myproject.internal -e 'kprobe:do_sys_open { printf("%s: %s\n", comm, str(arg1)) }'
+
+      # Execute a bpftrace program from file on a specific node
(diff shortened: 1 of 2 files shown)


plugins/trace.yaml
```diff
@@ -0,0 +1,80 @@
...
+# limitations under the License.
+
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: trace
+spec:
+  version: "v0.3.4"
+  homepage: https://github.com/iovisor/kubectl-trace
+  shortDescription: Schedule bpftrace programs on your kubernetes cluster using the kubectl.
```
ahmetb, line 22: `Schedule bpftrace programs on your cluster`

would be a short summary that doesn't get truncated.

plugins/trace.yaml
```diff
@@ -0,0 +1,80 @@
...
+  caveats: |
+    kubectl trace is now installed!
```
ahmetb, line 24: we usually reserve `caveats` for software that needs to be installed after plugin is installed.

you should probably already have the following examples in your `--help`?

ahmetb:
> If anyone can test it on windows before I do and report here I'll be grateful.

I suspect neither of the maintainers have a windows box. :(

plugins/trace.yaml
```diff
@@ -0,0 +1,80 @@
...
+metadata:
+  name: trace
+spec:
+  version: "v0.3.4"
+  homepage: https://github.com/iovisor/kubectl-trace
+  shortDescription: Schedule bpftrace programs on your kubernetes cluster using the kubectl.
+  caveats: |
+    kubectl trace is now installed! 
+
+    Examples:
```
corneliusweig, line 26: I think these examples would fit better in a `description`, of as @.ahmetb said in the `--help` output.

Also, it's usually nice to put a rough description of the technical details into the description. Just to give the users an idea of what is going on when invoking the plugin.

fntlnz: That's exactly what I wanted thanks @corneliusweig !

plugins/trace.yaml
fntlnz: I added them to caveats as suggetsed by @corneliusweig

plugins/trace.yaml
fntlnz: Thanks, will change this

fntlnz:
@ahmetb @corneliusweig I just addressed all the suggestions and tested again.

Thank you two for the reviews, I really appreciate it. And thanks for this awesome tool!

fntlnz:
Created the corresponding PR on kubectl trace for when this is merged. https://github.com/iovisor/kubectl-trace/pull/118

plugins/trace.yaml
corneliusweig: I think you gave a nice collection of examples which are well suited for people familiar with `kubectl-trace`. However, I'd recommend to keep the description on a more conceptual level and give the (potential) users an idea of the key concepts of `kubectl-trace`. Also, you should try to explain why and in which situations users could benefit from this plugin.
Further, you should explain how `trace` does its magic, for example:

> `trace` loads bpf programs into the host kernels on the target node and uses these programs to collect metrics.

(not sure if that's totally accurate, but you get the point, right?)

In essence, the description is a bit of a sales pitch for potential users. So use it wisely :)

fntlnz:
Thanks for your suggestions @corneliusweig - updated!

ahmetb:
It seems `description` now only lists examples. Was that your intention @fntlnz? If so, I'll go ahead and merge. BTW you can set up an easy [release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) to your repo so you don't have to come here to edit your yaml file every time you  have a new version.

fntlnz:
Thanks everyone!

PR #660, opened 2020-06-20
Add fuzzy plugin

d-kuro:
This PR aims to add [fuzzy](https://github.com/d-kuro/kubectl-fuzzy) plugin to krew index.

This plugin is a kubectl plugin providing a fuzzy finder selector.

~~There is a tool of the same name, but it is not kubectl plugin, so I thought it was OK.~~
https://github.com/bonnefoa/kubectl-fzf

I'm open to suggestions about the problem.
Thanks!


ahmetb:
I don’t think it will be wise to assign the fzf name:

1. fzf project already exists
2. There are other plugins named fzf (https://github.com/bonnefoa/kubectl-fzf) what if they come and ask us to add their plugin (we don’t want to have first-comer advantage)

d-kuro:
@ahmetb 

Thank you comments.
Hmm.., I understand. I think you're right.

For example, what if you renamed `kubectl-fzf` to `kubectl-fuzzyfinder`?
I liked the short name fzf, but I'm considering renaming the project.

ahmetb:
I think even `fuzzy` could work. I’ll defer to @corneliusweig.

corneliusweig:
Yeah, `fzf` is pretty much a trademark in the open source world. I think people would be confused to find that your plugin isn't the real `fzf`.
`fuzzy` sounds pretty good to me.

d-kuro:
@ahmetb @corneliusweig 
Thank you!
I will rename the project to kubecl-fuzzy based on your advice.

I will modify this Pull Request after I have renamed the project and made a release.

d-kuro:
@ahmetb @corneliusweig

I renamed the project and fixed the Pull Request.
Please review 🙏

plugins/fuzzy.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - from: kubectl-fuzzy.exe
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v1.0.1"
+  shortDescription: Fuzzy Finder Pod selector plugin. Currently "kubectl logs" and "kubectl exec" are supported.
```
corneliusweig, line 44: Please find a more succinct description for you plugin. Anything above ~50 chars gets usually truncated.

plugins/fuzzy.yaml
```diff
@@ -0,0 +1,49 @@
...
+  description: |
+    Kubectl fuzzy is a kubectl plugin providing a Fuzzy Finder Pod selector.
+    Uses go-fuzzyfinder, so there is no dependency on fzf binaries or anything else.
```
corneliusweig, line 47: This is an implementation detail and can safely be skipped in the description. I think it's better to be mentioned on the project homepage.

plugins/fuzzy.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v1.0.1"
+  shortDescription: Fuzzy Finder Pod selector plugin. Currently "kubectl logs" and "kubectl exec" are supported.
+  description: |
+    Kubectl fuzzy is a kubectl plugin providing a Fuzzy Finder Pod selector.
```
corneliusweig, line 46: It's not clear to me what this plugin actually does. Is it really just for selecting pods by a fuzzy pattern? If that is indeed the case, I think this is not the right approach. The recommended way to select pods in Kubernetes is via labels. Selecting individual pods via fuzzy match seems pretty wonky to me.

My original assumption was that this plugin does some fuzzy magic on the output of some kubectl commands. Apparently, I misread this :/

FWIW, there are already a few plugins which focus on `exec` and `logs` commands. For example https://github.com/gabeduke/kubectl-iexec, https://gitlab.com/grzesuav/kubectl-mtail, https://github.com/boz/kail and probably more.

ahmetb:
I recommend installing and trying out. 
It seems the main use case is to type for example “kubectl fuzzy describe pods” and then use fzf to type a few chars to do fuzzy string match. This isn’t like a label selector, maybe the description is not great.

ahmetb:
@corneliusweig Ping for the comment above.

plugins/fuzzy.yaml
d-kuro: Thanks for the comment. I will fix the tool description.

That's right, you can select an object in the label selector.
But in order to use the label selector, you need to know which label is being used on the target Kubernetes object.
I was too lazy to look up labels, so I decided to create this tool.
A fuzzy search allows you to select an object without having to find out which label is used for the Kubernetes object.

As for exec and logs, I know there are other tools for interactively selecting pods, but I think that fuzzy searches differentiate them from other tools.

I have also added support for kubectl describe yesterday.
I'm going to add more features to this tool in the future.
https://github.com/d-kuro/kubectl-fuzzy/releases/tag/v1.1.0

plugins/fuzzy.yaml
d-kuro: Fixed, Thanks!
https://github.com/kubernetes-sigs/krew-index/pull/660/commits/6eddc7e881362e0ba3144a869bf031ac16e91c33

plugins/fuzzy.yaml
d-kuro: I updated the description.
https://github.com/kubernetes-sigs/krew-index/commit/6eddc7e881362e0ba3144a869bf031ac16e91c33

plugins/fuzzy.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - from: kubectl-fuzzy.exe
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v1.2.0"
+  shortDescription: Fuzzy Finder Kubernetes object selector plugin.
```
ahmetb, line 44: This is still not good. You should consider never using the word `selector`. It has a specific meaning in Kubernetes that is causing all the misunderstanding w/ @corneliusweig and it will confuse users too.

You should update your repo and help messages ideally but here's my attempt at writing your description:

> Fuzzy and partial string search for kubectl

full description:

> This tool uses fzf(1) to do partial or fuzzy search of Kubernetes resources. Instead of specifying full resource names to kubectl commands, you can choose them from an interactive list that you can filter by typing a few characters.

d-kuro: Thanks for the advice.
I think the description you have come up with is smart!

I'm sorry if I sound picky. 
Strictly speaking, I use a different implementation of fuzzy finder than github.com/junegunn/fzf, so I'd like to rephrase it to fzf-like.

> This tool uses a fzf(1)-like fuzzy-finder to do partial or fuzzy search of Kubernetes resources.

I will fix it. (at the same time, I will fix the repository README)

d-kuro:
@ahmetb Thanks for your support!

corneliusweig:
Sorry for being such a 🐌.. and too late for the 🎉 
I think you are taking the plugin in a great direction. After having a better look at it, this makes total sense and I think this is a great addition to the index! Looking forward to your next features!

PR #687, opened 2020-07-07
Add ipick

kaplanelad:
Hey, 

I have created a new plugin that helps you to work interactively which provides flexible interface and very easy to use. 
To get more details and see example of the tool in use, visit https://github.com/similarweb/kubectl-interactive/blob/master/README.md 

Tests:
1. Go to kubectl-interactive latest release: https://github.com/similarweb/kubectl-interactive/releases/tag/v1.0.3
2. Download the relevant binary file 
3. Run the command: `kubectl krew install --manifest=${PWD}/interactive.yaml --archive={ARCHIVE_VERSION}`

Thanks


ahmetb:
Thanks for your submission.
I’ll have a problem with the name `"interactive"`. It violates our naming guide by taking advantage of being the first-comer of a _very_ generic verb/noun. As you can imagine, you could potentially do many things interactively (viewing logs, choosing sub-commands/options, deleting objects, querying resources, choosing a service to be proxied locally...). Therefore this name will not work, we cannot assign this `interactive` since other plugins might come and do this task better in the future.

We currently have many plugins that do some form of interactive task another (`iexec`, `fuzzy`, `ctx`, `ns`, `tap`) and they have appropriate names for the task they do.

plugins/interactive.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: interactive
+spec:
+  version: v1.0.3
+  homepage: https://github.com/similarweb/kubectl-interactive
+  shortDescription: A kubectl plugin that provides interactive, easy-to-use and flexible interface
```
ahmetb, line 8: this doesn't explain nearly anything :) the phrase. has the same level of detail as the word "interactive".

plugins/interactive.yaml
```diff
@@ -0,0 +1,42 @@
...
+  description: |
+    A kubectl plugin that is easy-to-use, flexible, and interactive
+    You will have the ability to roam around the Kubernetes world with a blazing fast selection menu
```
ahmetb, line 11: again the description doesn't explain anything specific and it has multiple subjective statements such as _easy-to-use, flexible, and blazing fast_.

please explain what the command actually does.

plugins/interactive.yaml
```diff
@@ -0,0 +1,42 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/similarweb/kubectl-interactive/releases/download/v1.0.3/kubectl-interactive_1.0.3_Darwin_x86_64.tar.gz
+    sha256: d11624612a8924fa61b6e14e7d1d48731e283bbf85e2105dbc56ff07b44441a0
+    files:
+    - from: "*"
+      to: "."
```
ahmetb, line 21: this is the default, you can remove `files:` altogether.

ahmetb:
Upon closer look at the demo GIF in the readme, this plugin seems to be a subset of the capabilities of the `fuzzy` plugin. So having this in krew-index might not be something that adds value to our users as it stands today.

kaplanelad:
Thanks @ahmetb for your fast response

**Regarding your wondering, what is the difference between this plugin to the others:**

Interactive plugin, it's actually a "proxy" to kubectl with adding to the user the selection + better filtring view.
Actually this plugin supports all kubectl actions that kubectl has to offer. 
this plugin allows you to delete any resources (pod/deployment/ns etc), node actions, edit/exec/log/describe **(and any kubectl actions) with prompt selection and --like resource filtring.**

The plugins that you mentions (`iexec`, `fuzzy`, `ctx`, `ns`, `tap`) are great plugins but each plugin handle very specific functionality mostly around (exec, log and describe). 
for example :
`iexec` - show selection pods to connect

`fuzzy` - allow you to work fast only with `describe`, `exec` and `logs`

`ctx` / `ns` - allow you to switch contexts/namespaces but then you should do that command to do the action
 

In my opinion, this plugin has place in krew-index because it is different from the other plugin you mentioned, this plugin supports all kubectl actions (not a specific one), .
I think this plugin is very useful  and gives a lot of value to the users, and on your side, you give to Krew users the option to choose the relevant plugin for them

Hope that I succeeded to explain more about how the plugin different for the another  and how this plugin can help to Kubernetes community 

**Request Changes:**
I have fixed your request changes.

**Plugin Name:**
regarding the plugin name, I'll change the name of the plugin when we agree that this plugin have a place in your repository :) (very hope that will) 

Thanks

plugins/interactive.yaml
kaplanelad: removed

plugins/interactive.yaml
kaplanelad: changed, hope it makes sense now

plugins/interactive.yaml
kaplanelad: changed, hope it makes sense now

ahmetb:
/cc @corneliusweig 
for naming recommendations

kaplanelad:
Hey @ahmetb,
What do you think about `kubectl select-resource`?
I think that is a name which better describes the plugin's purpose 

Do you want me to change it?

plugins/interactive.yaml
```diff
@@ -0,0 +1,35 @@
...
+    bin: kubectl-interactive
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/similarweb/kubectl-interactive/releases/download/v1.0.3/kubectl-interactive_1.0.3_Windows_x86_64.tar.gz
+    sha256: b466ec910933d646e120ca65884f2413fe6fd1289a88408735003927ff2080cc
+    bin: kubectl-interactive.exe
+
+
```
corneliusweig, line 35: Please remove trailing whitespace

plugins/interactive.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: interactive
+spec:
+  version: v1.0.3
+  homepage: https://github.com/similarweb/kubectl-interactive
+  shortDescription: A kubectl plugin that provides an interactive wrapper on top of kubectl.
+  description: |
```
corneliusweig, line 9: Please wrap at ~80 chars so that is prints nicely on wide and narrow terminals.

plugins/interactive.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: interactive
+spec:
+  version: v1.0.3
+  homepage: https://github.com/similarweb/kubectl-interactive
+  shortDescription: A kubectl plugin that provides an interactive wrapper on top of kubectl.
```
corneliusweig, line 8:
```suggestion
  shortDescription: A kubectl wrapper for interactive resource selection
```

plugins/interactive.yaml
```diff
@@ -0,0 +1,35 @@
...
+  description: |
+    A kubectl plugin that provides an interactive wrapper on top of kubectl.
+    This plugin is similar to a proxy in front of kubectl that provides an easier method to use and explore kubernetes objects with confidence.
```
corneliusweig, line 11:
```suggestion
    This plugin is similar to a proxy in front of kubectl that provides an easier method to use and explore kubernetes objects.
```
Please try to explain what your plugin does instead of making it sound like a marketing pitch.

What about this:

> This plugin is a smart wrapper around kubectl, which lets you pick the resource to act on. For example, `kubectl interactive exec` will show a list of pods and execute `kubectl exec <selected-pod>`.

plugins/interactive.yaml
kaplanelad: removed

plugins/interactive.yaml
kaplanelad: done

plugins/interactive.yaml
kaplanelad: changed

plugins/interactive.yaml
kaplanelad: Nice. thanks for that.
the text was changed

kaplanelad:
Thanks @corneliusweig for your review and the great feedback!

1. The plugin name was changed to pick (like your recommendation)
2. I have fixed all your request changes
3. The resource selection was changed, and now the plugin using https://github.com/AlecAivazis/survey library
4. The `-f` flag was removed, and not I'm using -- instead.

Please let me know if you have more issues that you want me to fix.

Thanks!

plugins/ipick.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: ipick
+spec:
+  version: v1.0.4
+  homepage: https://github.com/similarweb/kubectl-ipick
+  shortDescription: A kubectl wrapper for interactive resource selection.
+  description: |
+    This plugin is a smart wrapper around kubectl, which lets you pick 
+    the resource to act on. For example, kubectl ipick exec will 
+    show a list of pods and execute kubectl exec <selected-pod>.
```
corneliusweig, line 12:
```suggestion
    the resource to act on. For example, `kubectl ipick exec` will 
    show a list of pods and execute `kubectl exec <selected-pod>`.
```
nit: It would improve readability to enclose the commands in backticks -- even though this is just dumped to the command line without any markup.

kaplanelad: changed. thanks !

kaplanelad:
Thanks @corneliusweig

ahmetb:
Sorry
/lgtm
/hold cancel

PR #690, opened 2020-07-08
Add `save` plugin

nathforge:

### DIFF
--- a/plugins/save.yaml
+++ b/plugins/save.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: save
+spec:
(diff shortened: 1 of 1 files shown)


nathforge:
CLA is now signed...

ahmetb:
Thanks. This name will not work. Did you have a chance to read the naming style guide?

Please read this comment as it exactly applies here: https://github.com/kubernetes-sigs/krew-index/pull/687#issuecomment-655077038

nathforge:
Thanks for the quick feedback @ahmetb - would `split-yaml` be better?

ahmetb:
Yeah we can think about the name a bit. “To-files” isn’t a bad name either. Or “dump-files”, “takeout”, “yaml2files”, or some more unique and creative name you can think of.

PR #691, opened 2020-07-08
Add split-yaml plugin

nathforge:
Installed & tested from manifest on Linux, OS X and Windows amd64.

### DIFF
--- a/plugins/split-yaml.yaml
+++ b/plugins/split-yaml.yaml
@@ -1,0 +1,69 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: split-yaml
+spec:
+  version: v0.1.0
+  homepage: https://github.com/nathforge/kubectl-split-yaml
+  shortDescription: Split YAML output into one file per resource.
+  description: |
+    Split Kubernetes YAML output into one file per resource, e.g
+    `kubectl get deploy -o yaml | kubectl split-yaml -p deployments`
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm
+    uri: https://github.com/nathforge/kubectl-split-yaml/releases/download/v0.1.0/kubectl-split-yaml_0.1.0_linux_armv6.tar.gz
+    sha256: 0886f2f8877ae5e7d2bc920dbc0cd23cab9aa24a7f820c445a9839cde7fa13ea
+    bin: kubectl-split-yaml
+  - selector:
+      matchLabels:
+        os: linux
+        arch: '386'
+    uri: https://github.com/nathforge/kubectl-split-yaml/releases/download/v0.1.0/kubectl-split-yaml_0.1.0_linux_386.tar.gz
+    sha256: 49fd3b9de779388f6e68d09131ae1a598e7f1a17ba95f2c3c1eb05161e4b87d7
+    bin: kubectl-split-yaml
+  - selector:
+      matchLabels:
+        os: windows
+        arch: '386'
+    uri: https://github.com/nathforge/kubectl-split-yaml/releases/download/v0.1.0/kubectl-split-yaml_0.1.0_windows_386.tar.gz
+    sha256: 81ebb46a8d8318026c8a1a57d9a9d2b501568bdcfc661af2ae8789b858f4ede5
+    bin: kubectl-split-yaml.exe
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/nathforge/kubectl-split-yaml/releases/download/v0.1.0/kubectl-split-yaml_0.1.0_linux_amd64.tar.gz
+    sha256: 8fa9cd92c38f745bf22c5415573067bda2955016db9e5bdcec165fc7b6a2e116
+    bin: kubectl-split-yaml
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm64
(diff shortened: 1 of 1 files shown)


plugins/split-yaml.yaml
```diff
@@ -0,0 +1,92 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/nathforge/kubectl-split-yaml/releases/download/v0.0.1/kubectl-split-yaml_0.0.1_darwin_amd64.tar.gz
+    sha256: 05ad495bb6045355c064235e945f9b1860f181046f1cc1d7fc75199bafb820c5
+    files:
+    - from: '*'
+      to: .
```
ahmetb, line 20: this statement is redundant since that's the default, you can remove.

ahmetb:
On a personal note: I really like the plugin. Though, it could probably use some customization around how the files are saved (e.g. specify output directory, maybe people don't need apiversions as directory names, maybe allow --name-template to determine filename format etc).

Most likely, I would use this with other plugins like:

   kubectl get -o=yaml | kubectl neat (or eksporter) | kubectl split-yaml -o=out/

plugins/split-yaml.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: split-yaml
+spec:
+  version: v0.0.1
+  homepage: https://github.com/nathforge/kubectl-split-yaml
+  shortDescription: Save Kubernetes YAML output into one file per resource.
+  description: This plugin takes output from `kubectl get -o yaml` and saves one file
+    per resource.
```
corneliusweig, line 10: We usually don't recommend usage instructions here, because most plugins run standalone. However, your plugin is designed to be piped into, so it would make sense to give an example here.

You could also mention that it works well with `get-all`, since this is explicitly mentioned on your landing page and `get-all` is also a krew plugin. I might have a conflict of interest here though, so it's completely up to you.

ahmetb:
Ping @nathforge for review updates.

nathforge:
> 
> 
> Ping @nathforge for review updates.

plugins/split-yaml.yaml
nathforge: Thanks, have done.

nathforge:
> On a personal note: I really like the plugin. Though, it could probably use some customization around how the files are saved (e.g. specify output directory, maybe people don't need apiversions as directory names, maybe allow --name-template to determine filename format etc).
> 
> Most likely, I would use this with other plugins like:
> 
> ```
> kubectl get -o=yaml | kubectl neat (or eksporter) | kubectl split-yaml -o=out/
> ```

@ahmetb Thank you - it did have those options though the README didn't mention them so have now updated.

nathforge:
> This looks really neat! The manifest is great, too!
> 
> I'd have a suggestion for your implementation though. You depend on `cli-runtime` only to get a struct which wraps the three standard unix streams (https://github.com/nathforge/kubectl-split-yaml/blob/f1731237efcf876af9af31f6e6df350b70885c2b/internal/cmd/splityaml.go#L25). This adds quite some weight to your binary, but you don't really need that. Unless you plan to talk to the cluster directly, consider removing that dependency by duplicating the type in your project.
> 
> Btw, I also opened an issue in `ketall` so that you don't need to work around my bug 😉

@corneliusweig Perfect, that's a much faster & smaller build. I'll try to look at the ketall issue if not beaten to it!

plugins/split-yaml.yaml
nathforge: Have done thanks. I've only added one example as `krew info` seems to collapse whitespace so wanted to keep it brief.

nathforge:
Thank you both, I've released a new build and updated the manifest.

plugins/split-yaml.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: split-yaml
+spec:
+  version: v0.1.0
+  homepage: https://github.com/nathforge/kubectl-split-yaml
+  shortDescription: Split Kubernetes YAML output into one file per resource.
+  description: Split Kubernetes YAML output into one file per resource, e.g `kubectl get deploy -o yaml | kubectl split-yaml -p deployments`
```
corneliusweig, line 9: Could you wrap that line at ~80 chars, so that it prints nicely on wide and narrow terminals?

plugins/split-yaml.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: split-yaml
+spec:
+  version: v0.1.0
+  homepage: https://github.com/nathforge/kubectl-split-yaml
+  shortDescription: Split Kubernetes YAML output into one file per resource.
```
corneliusweig, line 8: It's OK, but this is more concise :
> Split YAML output into one file per resoure

This field is usually truncated at ~50 chars, so it's better to stay below that threshold

plugins/split-yaml.yaml
nathforge: Sure, that's done.

plugins/split-yaml.yaml
nathforge: Good call, done.

nathforge:
Forgot to ping you @corneliusweig - think it's all good! Thanks for your help

PR #705, opened 2020-07-13
Add kubechonk plugin to index

nickgerace:
[This plugin ](https://github.com/nickgerace/kubechonk) finds the "[chonkiest](https://www.dictionary.com/e/slang/chonky/)" nodes in your cluster. Specifically, it finds the nodes with the highest number of CPU cores, and largest amount of memory available.

Thank you for reviewing my PR.

### DIFF
--- a/plugins/chonk.yaml
+++ b/plugins/chonk.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: chonk
+spec:
+  version: v0.1.1
+  homepage: https://github.com/nickgerace/kubechonk
+  shortDescription: Find the chonkiest nodes in your cluster.
+  description: |
+    This plugin returns all the node(s) with the highest number of CPU cores,
+    and all the node(s) with the largest amount of memory.
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/nickgerace/kubechonk/releases/download/v0.1.1/kubechonk-v0.1.1-linux-amd64.tar.gz
+    sha256: def8abd34130a797869dd6c2b3ac2a4b7af8901a1a36d288d1efa9fc33402b6e
+    bin: kubectl-chonk
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/nickgerace/kubechonk/releases/download/v0.1.1/kubechonk-v0.1.1-darwin-amd64.tar.gz
+    sha256: ea981378e1e02c0df71a095cf03f50394b7497fd8c6a3ed5fa245bc0ae85edc5
+    bin: kubectl-chonk
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/nickgerace/kubechonk/releases/download/v0.1.1/kubechonk-v0.1.1-windows-amd64.zip
+    sha256: 78ebdecf11b91e60ce4928b5a43b78f157654c6fe4267ba03048caa95782ed37
+    bin: kubectl-chonk.exe


nickgerace:
My apologies for the double PR. I thought the CLA trigger was not working. It was me!

nickgerace:
I signed it

ahmetb:
Please submit a support request at https://identity.linuxfoundation.org/projects/cncf instead of closing/opening new PRs.

plugins/chonk.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: chonk
+spec:
+  version: v0.1.1
+  homepage: https://github.com/nickgerace/kubechonk
+  shortDescription: Find the chonkiest nodes with this kubectl plugin.
```
ahmetb, line 8: `with this kubectl plugin.` redundant. it's clear it's a kubectl plugin.

plugins/chonk.yaml
```diff
@@ -0,0 +1,32 @@
...
+  description: |
+    This plugin returns all the node(s) with the highest number of CPU cores, and all the node(s) with the largest amount of memory.
```
ahmetb, line 10: wrap at 80 chars please.

plugins/chonk.yaml
```diff
@@ -0,0 +1,32 @@
...
+    sha256: def8abd34130a797869dd6c2b3ac2a4b7af8901a1a36d288d1efa9fc33402b6e
+    bin: kubectl-chonk
+  - selector:	
+      matchLabels:	
+        os: darwin	
+        arch: amd64	
+    uri: https://github.com/nickgerace/kubechonk/releases/download/v0.1.1/kubechonk-v0.1.1-darwin-amd64.tar.gz
+    sha256: ea981378e1e02c0df71a095cf03f50394b7497fd8c6a3ed5fa245bc0ae85edc5
+    bin: kubectl-chonk
+  - selector:
```
ahmetb, line 26: you have trailing `tab` characters in half of the lines, please consider cleaning those up.

ahmetb:
Thanks for your submission. We have some plugins such as `resource-capacity` that already show CPU/memory available on a node. Can you please make sure this is not a subset of another plugin that's already available?

nickgerace:
Not a problem, thank you for the review. 

I would say ```status```, ```resource-capacity```, and ```view-allocations``` occupy similar use cases. I will admit, while this plugin specifically aims to do "less" than the those plugins, you deduce the same answers with them (i.e. which nodes have the "most" $RESOURCE).

I was unaware of partial usage overlap being an issue, barring the ["case-by-case" disclaimer in the README](https://github.com/kubernetes-sigs/krew-index#submitting-new-plugins). I used the [docs](https://krew.sigs.k8s.io/docs/developer-guide/release/new-plugin/) to guide me through this process.

Regardless, if that's a non-starter, that is totally fine. I understand 😄.

nickgerace:
I signed it

nickgerace:
Even though you are already reviewing this PR @ahmetb (thank you by the way!), I'll finish the prerequisites for this PR via the CI bot instructions...

/assign corneliusweig

corneliusweig:
Hey @nickgerace, thanks for your submission. I think the overlap with the other plugins about resource usage is too big. Plus the existing plugins solve the problem in a more generic way. For example, it is a bit arbitrary to show only the largest nodes with respect to memory and cpu capacity.

We tried to formalize that plugins with too much overlap should be rejected, but it was too vague to make a general rule of it. With more and more plugins in the index, I expect this will happen more often.

So let's close this PR for now. If you think that I have missed an important point (for example why it is generally important to know the chonkiest node), then let's reopen. With the upcoming feature of private indexes, you could also add your plugin to another index.

/close

nickgerace:
@corneliusweig - Thank you for the review. I completely understand your reasoning. If that discussion ever opens back up, I'd like to add that I may have held off on making this PR if there was an indication in the README. Although, I also understand that those kinds of rules (or even encouraging guidelines) can be difficult to define clearly.

Thank you both @ahmetb @corneliusweig! In the meantime, I'll do some kubectl user research on desired features that seldom overlap with other krew-index plugins.

PR #718, opened 2020-07-22
add emit-event plugin

rajatjindal:
Hi krew-index team,

I would request you to please accept the new plugin I wrote. `emit-event`. 

This basically just allows to emit events for k8s objects

Thanks
Rajat Jindal


corneliusweig:
🤔 Can you explain what are the use-cases for this plugin?

plugins/emitevent.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-emitevent/releases/download/v0.0.1/kubectl-emitevent_v0.0.1_linux_amd64.tar.gz
+    sha256: ad811795cece68d4d8d39a63d98feae68a48c249b02bf1314e8e676d379e36b1
+    bin: kubectl-emitevent
+  shortDescription: emit event for the requested object
+  description: |
+    This plugin emits the event for the requested object
```
ahmetb, line 25: not very descriptive

rajatjindal:
the idea was to be able to emit events during ci/cd pipeline. 

e.g. lets say we updated configmap during the pipeline, it will update resourceversion, but it does not have anything like 'updatedAt' and also does not have something like "last-applied-configuration'.

in that case we can emit an event with a reason and message, that can help in debugging

Kindly let me know if that makes any sense.

Thanks

ahmetb:
I think it makes sense to "manufacture" synthetic events. I'm not aware of a CLI tool that lets you do that so it seems to be a legit use case. But it's worth capturing in `description` fields. Otherwise it's not clear what "event" actually refers to (e.g. Kubernetes Events API)

rajatjindal:
Hi Ahmet

I've updated the description. As always thanks for the feedback, and let me know if I should update it further.

Thanks

plugins/emitevent.yaml
```diff
@@ -0,0 +1,25 @@
...
+    sha256: b86b0c6c24d6db3adfd523eee1f73db8d09b92613080169cd5c6e8853e035f06
+    bin: kubectl-emitevent
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-emitevent/releases/download/v0.0.1/kubectl-emitevent_v0.0.1_linux_amd64.tar.gz
+    sha256: ad811795cece68d4d8d39a63d98feae68a48c249b02bf1314e8e676d379e36b1
+    bin: kubectl-emitevent
+  shortDescription: emit Kubernetes Events for the requested object
```
ahmetb, line 23: Capitalise please.

plugins/emitevent.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: emitevent
```
ahmetb, line 4: We often recommend people to split words using - (kebab-case, see Naming doc). 

@corneliusweig wdyt about this?

ahmetb:
can you please fix your --help message


```
emits an event to given object

Usage:
  emitevent [flags]
```

see https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/#help-messages

rajatjindal:
Hi Ahmet

I've fixed the usage msg. 

Thanks

ahmetb:
Unless you're in a hurry I will wait Cornelius to take a look at naming. 
/approve

rajatjindal:
Absolutely no hurry. Will wait for feedback on name

plugins/emitevent.yaml
corneliusweig: IMO, kebab-case is nicer to read, but harder to type. We are not completely consistent, but the great majority does use `-` between words. So I'd recommend to go with `emit-event` instead.

plugins/emitevent.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/rajatjindal/kubectl-emitevent/releases/download/v0.0.2/kubectl-emitevent_v0.0.2_linux_amd64.tar.gz
+    sha256: 058a56a9dcf3ea919fb443fea8da89649056337c0fa347f39a669b47ae2f59ea
+    bin: kubectl-emitevent
+  shortDescription: Emit Kubernetes Events for the requested object
+  description: |
+    This plugin emits the Kubernetes Events for the requested object
```
corneliusweig, line 25: We normally don't recommend this anymore, but in this case I think it would be good to include an example command. This will give interested users a better idea how the plugin works. For example

> kubectl emit-event daemonset/kube-proxy -n kube-system --reason "foo-reason" --message "bar-message"

rajatjindal:
I will make the suggested change tonight, thanks for the feedback both of you

rajatjindal:
I've renamed the plugin to emit-event and added example in description.

many thanks

rajatjindal:
woohoo !!! many thanks

PR #730, opened 2020-07-28
add kubectl fetch-yaml plugin

olegsu:
Related to https://github.com/kubernetes/kubectl/issues/894
- [X] Make sure you read the Plugin Naming Guide: https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/
- [X] Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...]


ahmetb:
Verb "fetch" is extremely generic and doesn't describe what it fetches. It's very unlikely we will accept this name. 

Please revisit the Naming Guide or watch this: https://youtu.be/8UYcmugKlUw

olegsu:
Thank you @ahmetb. 
Will update the pr with the updated name.

olegsu:
@ahmetb Renamed to `fetch-yaml`

plugins/fetch-yaml.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: Plugin
+metadata:
+  name: fetch-yaml
+spec:
+  version: "v0.1.0"
+  homepage: https://github.com/olegsu/kubectl-fetch-yaml
+  shortDescription: "Fetch Kubernetes files from remote destinations"
+  description: |
+    Fetch Kubernetes files from remote destinations
+    Example git: kubectl fetch-yaml git --repo https://github.com/kubernetes/examples --path guestbook/all-in-one | kubectl applay -f -
```
ahmetb, line 11: Please wrap at 80 chars.
Also typo.

plugins/fetch-yaml.yaml
```diff
@@ -0,0 +1,29 @@
...
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/olegsu/kubectl-fetch-yaml/releases/download/v0.1.0/kubectl-fetch-yaml_0.1.0_Darwin_x86_64.tar.gz
+      sha256: 39a6b5e97a57b2a0e9bcb883392fd13c29632bfb78ea440ad659a46356cfb234
+      files:
```
ahmetb, line 19: This is the default value. Remove “files:”

plugins/fetch-yaml.yaml
olegsu: Thanks, done

plugins/fetch-yaml.yaml
olegsu: Removed, thanks

plugins/fetch-yaml.yaml
olegsu: Done, thanks.

plugins/fetch-yaml.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: fetch-yaml
+spec:
+  version: "v0.1.0"
+  homepage: https://github.com/olegsu/kubectl-fetch-yaml
+  shortDescription: "Fetch Kubernetes files from remote destinations"
```
corneliusweig, line 8: What's the reason to chose this tool over something like `curl`, `httpie`, or `wget`?

It seems wasteful to install a new tool, and learn a new CLI language for something that can be done with standard tools.

olegsu: Thank for the feedback, @corneliusweig.

I can think of a few reasons to use plugin over native tools:
1. Cross-platform (in case it is supported)
2. People are looking for Kubernetes best practices and "native" integration.

For fetch-yaml plugin. There is a feature request to support private git repositories directly from `kubectl apply -f` command. On the last sig-cli bug scrub it was [suggested](https://github.com/kubernetes/kubectl/issues/894#issuecomment-662573257) that a better entry point for native kubectl support will be as a plugin.

corneliusweig: 1. Every platform has its own tools to download from the web. But ok, I'll accept that.
2. `curl | kubectl apply -f` is in itself an anti-pattern. Best practices for the `curl` part don't make that better.

Here is why:
1. *download-apply in one step is inherently insecure.* If you don't even have the chance to inspect the script before applying, you could install anything. And the content behind a web address may change anytime, so what was good before, does not have to be good the next time.
2. *The content is not version controlled.* Usually people pull whatever is on `main`/latest and just apply what's there. So you cannot reproduce your setup at a later time if you have to, if `main` has changed in between.

---

So if you want to implement best practices, maybe do this:

* if `fetch-yaml` downloads from a public repo, refuse unless a `--allow-insecure` option is specified.
* if `fetch-yaml` would pull from master, refuse and print a stable link instead. For example
  ```sh
  kubectl fetch-yaml github.com/kubernetes-sigs/krew/README.md  # should print an error such as:
  WARNING: Unstable URL. Force with --allow-ephemeral or use a the permanent link
  https://raw.githubusercontent.com/kubernetes-sigs/krew/605a639/README.md
  ```
  The permanent link is the real value here, IMO. It will require some work to generate this, though.

Then this plugin will be somewhat safe and implement best practices. WDYT?

/cc @ahmetb

olegsu: Thanks for the feedback, @corneliusweig 
I agree that download-apply is not the recommended way to install new stuff and examples, on the other hand, during ci/cd process it is acceptable.

I would suggest the following:
1. Accept `--revision` flag, when set, no error will be printed.
2. Accept `--allow-ephemeral` flag, when set, no error will be printed.
3. Print error message and exit otherwise.

ahmetb:
> What's the reason to chose this tool over something like curl, httpie, or wget?
> It seems wasteful to install a new tool, and learn a new CLI language for something that can be done with standard tools.

I'm also in this boat currently. If unix tools can be put together to achieve this, then we're likely looking at an anti-pattern solution here. 

> On the last sig-cli bug scrub it was suggested that a better entry point for native kubectl support will be as a plugin.

@olegsu are you working with anyone from the issue report or SIG CLI on this by any chance? maybe working longer on the idea and refining it first, then submitting it here might be a better option.

plugins/fetch-yaml.yaml
corneliusweig: IIUC, you suggest to have the `--revision` flag instead of encoding that in the URL? Should be fine in most cases, but of course I could download from a permanent URL to start with, so that no revision needs to be specified.

Just to make sure I understand correctly: according to your suggestion, I should still get an error when I use the `--revision` flag on a public repo, right?

corneliusweig:
As discussed above, the plugin needs to undergo some improvements. Just let us know when it's ready 🚀 
/hold

olegsu:
> @olegsu are you working with anyone from the issue report or SIG CLI on this by any chance? maybe working longer on the idea and refining it first, then submitting it here might be a better option.

I am not, I will link to this pr in the original feature request to get more feedback from the sig and from the author. The implementation was to support a similar use case in our ci/cd flow.

plugins/fetch-yaml.yaml
olegsu: I suggest accepting public repo without any error in case a revision flag set. So for example:
* Error:  `kubectl fetch-yaml git --repo https://github.com/kubernetes/examples --path guestbook/all-in-one` - prints an error and exit
* With **`--allow-insecure`** flag:  `kubectl fetch-yaml git --repo https://github.com/kubernetes/examples --path guestbook/all-in-one --allow-insecure` - Fetching the files, error is not printed.
* With **`--revision`** flag: `kubectl fetch-yaml git --repo https://github.com/kubernetes/examples --path guestbook/all-in-one --revision b86c9d5` - Fetching the files, error is not printed.

I was not thinking about URLs like https://raw.githubusercontent.com/kubernetes-sigs/krew/605a639/README.md. This is just another URL that is supported via `kubectl apply -f`

ahmetb:
/close
I'm closing this for now, however you can distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

PR #735, opened 2020-07-29
add kubectl node-pod plugin

mattfenwick:
This plugin shows the assignment of pods, containers, and jobs to nodes.  

I haven't found something else to do this, and it's been a handy tool for having a node-centric view of what's going on in my k8s clusters.

If this isn't a good fit for a krew plugin, I'm open to suggestions for how to address that, as well as suggestions on how to make this more useful!

#### PLUGIN DEVELOPERS: If you are submitting a new plugin

  - I tested the install by running `kubectl krew install --manifest=./deploy/krew/node-pod.yaml` from a checkout of https://github.com/mattfenwick/krew-node-pod

### DIFF
--- a/plugins/node-pod.yaml
+++ b/plugins/node-pod.yaml
@@ -1,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: node-pod
+spec:
+  version: "v0.1.0"
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/mattfenwick/krew-node-pod/releases/download/v0.1.0/kubectl-node_pod_linux_amd64.tar.gz
+    sha256: "c21488a37c7a29b43c22cf078c772858f1380d7836eaddeaf34a171fa405a821"
+    files:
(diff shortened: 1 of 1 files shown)


corneliusweig:
As far as I understand, your plugin basically lists the pods in all namespaces and prints the `nodeName` that pod is running on.

In other words, it is `kubectl get pods -A` with a custom set of columns to show. Incidentally, there is already a plugin for exactly that: https://github.com/webofmars/kubectl-custom-cols. Its author welcomes additional presets for column lists. Maybe you can upstream your set of columns there?

Overall it looks to me that the plugin so far has a narrow focus around a problem that can be better solved with a shell alias. Do you have plans to extend the plugin in a meaningful way so that this would change? Otherwise, I'm against merging.

mattfenwick:
Hi @corneliusweig thanks for taking a look, I appreciate the feedback!

I'll do some more research and look into the plugin you mentioned, and either improve this plugin or close this PR!

corneliusweig:
Sounds good!
/hold

mattfenwick:
@corneliusweig I think you're right, this plugin needs more/differentiated functionality before it should be considered for krew.

Thanks again for taking a look, and thanks again for the pointers!  Closing for now.

PR #737, opened 2020-07-30
Add operator plugin

joelanford:
This plugin _greatly_ simplifies the process of managing Kubernetes Operators with Operator Lifecycle Manager. With this plugin, users can:
* add and remove operator catalogs
* list installed catalogs and their available operators
* install, upgrade, and uninstall operators
* list installed operators

This plugin is a component of the Operator Framework, which is an incubating CNCF project.

### DIFF
--- a/plugins/operator.yaml
+++ b/plugins/operator.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: operator
 spec:
-  version: v0.0.3
+  version: v0.0.4
   homepage: https://github.com/operator-framework/kubectl-operator
   shortDescription: Manage operators with Operator Lifecycle Manager
   description: |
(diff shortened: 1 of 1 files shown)


ahmetb:
The word "operator" is extremely overloaded already and doesn't describe what it fetches. It's very unlikely we will accept this name. 

Please revisit the [Naming Guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/) or watch this: https://youtu.be/8UYcmugKlUw

ahmetb:
That said since this is part of the operator framework, we can consider making an exception. 
/cc @corneliusweig

joelanford:
@ahmetb I guessed there might be some concern with the name, but wasn't sure how to else to name it given all of the underlying functions it handles. Suggestions welcome if `operator` is too generic.

corneliusweig:
I think we can argue that `operator` can be an exception here, as the project is incubating.

But first I'd like to evaluate the alternatives. `operator` is a rather long name and does not say what it does, other than that it is about operators. A verb term usually works much better, but this is also difficult in this case.
So what do you think about `olm` (=Operator Lifecycle Manager) instead? It's shorter+easier to type.
Another option could be `opmgr` (=Operator Manager).
WDYT? 🤔

joelanford:
I still prefer `operator`, I think. I did consider `olm` but that would result in overall more typing for users who would generally be using the `install/upgrade/uninstall` subcommands more often than the catalog subcommands.

Right now the base subcommands make typical usage look like this:
```sh
kubectl operator install
kubectl operator upgrade
kubectl operator uninstall
kubectl operator catalog add
kubectl operator catalog remove
```

If we made it `olm`, I think we'd have to move `operator` to a subcommand of `olm`. E.g.:
```sh
kubectl olm operator install
kubectl olm operator upgrade
kubectl olm operator uninstall
kubectl olm catalog add
kubectl olm catalog remove
```

plugins/operator.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: operator
+spec:
+  version: v0.0.3
+  homepage: https://github.com/operator-framework/kubectl-operator
+  shortDescription: Manage Kubernetes Operators with Operator Lifecycle Manager
```
corneliusweig, line 8:
```suggestion
  shortDescription: Manage operators with Operator Lifecycle Manager
```
Everything above 50 chars is usually truncated.

plugins/operator.yaml
```diff
@@ -0,0 +1,46 @@
...
+  description: |
+    This plugin is a package manager for Operators in your cluster. It
+    simplifies adding and removing Operator catalogs, and it has familiar
+    commands for installing, uninstalling, and listing available and
+    installed Operators.
+  caveats: |
+    * This plugin requires Operator Lifecycle Manager to be installed in your
+      cluster. See the installation instructions at https://olm.operatorframework.io/docs/getting-started/
```
corneliusweig, line 16:
```suggestion
      cluster. See the installation instructions at
      https://olm.operatorframework.io/docs/getting-started/
```

plugins/operator.yaml
```diff
@@ -0,0 +1,46 @@
...
+  name: operator
+spec:
+  version: v0.0.3
+  homepage: https://github.com/operator-framework/kubectl-operator
+  shortDescription: Manage Kubernetes Operators with Operator Lifecycle Manager
+  description: |
+    This plugin is a package manager for Operators in your cluster. It
+    simplifies adding and removing Operator catalogs, and it has familiar
+    commands for installing, uninstalling, and listing available and
+    installed Operators.
```
corneliusweig, line 13: IIUC, the plugin in particular facilitates the usage of operators from https://operatorhub.io/. Might be worth mentioning.

dmesser: @corneliusweig While that is true operatorhub.io is merely just a public index. Users of that toolchain have the ability to create their own indexes and the `kubectl operator` plugin allows to use those as well. So there could be a pointer somewhere to operatorhub.io as an example for a catalog to add to your cluster, but otherwise they are reasonably decoupled.

ahmetb:
The precedent in this area is that we accept project names as the plugin names, rather than the generic verb/noun. For example:

* `kubectl cilium` could've been named `kubectl network-policy`
* `kubectl datadog` could've been named `kubectl monitoring`

That's why I’m currently +1 for `kubectl olm`. I recommend opening this up to debate in Operator Framework community. If it’s acceptable to rename (especially because the project is very new), it might be good to do so. `olm` also is a good brand name and is short/unique, which is what we prefer.

If there’s a significant pushback from Operator Framework side, we can reconsider our position and might allow `operator` name, but for now I am hesitant as well.

joelanford:
@ahmetb @corneliusweig There would definitely be confusion around `olm install`, since that's already a [subcommand of the `operator-sdk`](https://sdk.operatorframework.io/docs/new-cli/operator-sdk_olm_install/) project that installs OLM itself.

ahmetb:
So I’m trying to understand why was this plugin developed if Operator SDK already has an official CLI to bootstrap and manage OLM. Could the functionality of managing individual operators/catalogs be rolled into the `operator-sdk` tool? I’m trying to see where the logical divide is.

joelanford:
@ahmetb Sure!

`operator-sdk` is a development tool. It is focused on the operator development lifecycle, starting from scaffolding out operator code and ending with OLM integrations like generating operator packages (called bundles) and verifying that a candidate bundle can run in a cluster via OLM without having to deal with catalogs. Its focus is the source code of operator projects and building deployable artifacts. OLM is necessary for that last developer use case of testing out a bundle, hence the subcommand to install OLM in the `operator-sdk` tool.

`kubectl operator` is meant for cluster administrators. This persona is not touching operator source code (at least not often). Instead, someone using this tool would be making operators available for installation on an OLM-enabled cluster  (i.e. `kubectl operator catalog add`) and installing and upgrading operators (i.e. `kubectl operator install|upgrade`).

There's obviously going to be some overlap between these personas, but there isn't a whole lot. Operator developers will probably use `kubectl operator` every now and then, but cluster administrators are unlikely to use `operator-sdk`.

joelanford:
Thanks @corneliusweig and @ahmetb!

Hopefully if `kubectl` ever has a native `operator` subcommand it will be because the Operator Framework community helped make that happen! :smile: :wink:

PR #768, opened 2020-08-21
Upgrade cert-manager plugin to v1.0.0-beta.0

hzhou97:
Upgrade cert-manager plugin to v1.0.0-beta.0

### DIFF
--- a/plugins/cert-manager.yaml
+++ b/plugins/cert-manager.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: cert-manager
 spec:
-  version: v0.16.0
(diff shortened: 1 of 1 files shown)


hzhou97:
Hi, I thought it was a good way to get more users to use the plugin to get more feedback.
But if that is not conventional or if krew prefers to only host stable versions of plugins, this upgrade is not strictly necessary either.

ahmetb:
Yeah, krew prefers to host only stable versions.
You can however host your own plugin index if you’d like to have a non-stable channel for your users.

hzhou97:
Ok I see, thanks. I'll close this PR then.

PR #783, opened 2020-08-27
Add easy-rollback plugin

Dentrax:
This PR aims to add [easy-rollback](https://github.com/Trendyol/kubectl-easy-rollback) plugin to Krew index.

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

### DIFF
--- a/plugins/easy-rollback.yaml
+++ b/plugins/easy-rollback.yaml
@@ -1,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: easy-rollback
+spec:
+  version: v1.0.19
+  homepage: https://github.com/Trendyol/kubectl-easy-rollback
+  shortDescription: Rollback your images.
+  description: |
+    Show previously deployed image versions and easily rollback to selected one.
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/Trendyol/kubectl-easy-rollback/releases/download/v1.0.19/kubectl-easy-rollback_1.0.19_Darwin_x86_64.tar.gz
+      sha256: a9e1597b3f6f243fe53a71428f1b0f590b21be2c2afe53484496ba03db2c0f5b
+      bin: kubectl-easy-rollback
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/Trendyol/kubectl-easy-rollback/releases/download/v1.0.19/kubectl-easy-rollback_1.0.19_Linux_x86_64.tar.gz
+      sha256: b93694598c04a2379615ca73b477840ce419bac75d0ae50a5e65bc89a54a2107
+      bin: kubectl-easy-rollback
+    - selector:
+        matchLabels:
+          os: windows
+          arch: amd64
+      uri: https://github.com/Trendyol/kubectl-easy-rollback/releases/download/v1.0.19/kubectl-easy-rollback_1.0.19_Windows_x86_64.tar.gz
+      sha256: 2a81ef45045525cbac899a563a0599406f2cf88b25452fa69cf993fdaf7e32ee
+      bin: kubectl-easy-rollback.exe


plugins/easy-rollback.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: Plugin
+metadata:
+  name: easy-rollback
+spec:
+  version: v1.0.19
+  homepage: https://github.com/Trendyol/kubectl-easy-rollback
+  shortDescription: Show previously deployed image versions and easily rollback to selected one.
+  description: |
+    This plugin shows previously deployed image versions of a specified Kubernetes API deployment object
+    in the command-line. Then you have to select one of these image version, in order to successfully rollback.
```
ahmetb, line 11: wrap at 60 chars

plugins/easy-rollback.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: easy-rollback
+spec:
+  version: v1.0.19
+  homepage: https://github.com/Trendyol/kubectl-easy-rollback
+  shortDescription: Show previously deployed image versions and easily rollback to selected one.
```
ahmetb, line 8: this is too long, it will be truncated.
choose a shorter description.

plugins/easy-rollback.yaml
```diff
@@ -0,0 +1,35 @@
...
+  description: |
+    This plugin shows previously deployed image versions of a specified Kubernetes API deployment object
+    in the command-line. Then you have to select one of these image version, in order to successfully rollback.
+  caveats: |
+    * Currently only supported `--kubeconfig, --context, --namespace` from global flags namespace.
```
ahmetb, line 13: This isn't really a big caveat. Users can see this in --help.

ahmetb:
/approve
/assign @corneliusweig 

Also, please consider setting up [Krew release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) after merged. It helps you skip manually making updates to your Krew manifests on each new version and send a pull request. It’s a GitHub workflow bot that can run every time you push a new tag. Those PRs are automatically approved.

developer-guy:
> /approve
> 
> /assign @corneliusweig 
> 
> 
> 
> Also, please consider setting up [Krew release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) after merged. It helps you skip manually making updates to your Krew manifests on each new version and send a pull request. It’s a GitHub workflow bot that can run every time you push a new tag. Those PRs are automatically approved.
> 
> 

Thank you for your suggestion 🙏

plugins/easy-rollback.yaml
Dentrax: Done 👍

Dentrax:
@corneliusweig We've just fixed @ahmetb 's reviews, can you please look at #c038b2b?

ahmetb:
Please avoid unnecessary pings. We'll get to it, I assume this has no urgency.

corneliusweig (commented):
If I understand correctly, this plugin does pretty much exactly what the following commands do:
```shell
kubectl rollout status deployment/xx
# and
kubectl rollout undo deployment/xx --to-revision=N
```
Please correct me if I'm wrong. For such tasks, I think it is better to rely on standard kubectl functionality instead of building a dedicated go binary. So I'm currently -1 for merging this. Please explain why your plugin is different and what problem it solves better than `kubectl`.

---

There are also some things you could improve in your plugin:

1) You did not include any auth providers. That means you cannot authenticate against any of the cloud providers. Please have the following blank import somewhere in your `main.go`:

   ```go
   ```

2) I'd recommend to stick to standard k8s flags. You introduced a new flag `--kubeconfig` which is non-standard.
https://github.com/Trendyol/kubectl-easy-rollback/blob/73e6b084b630b8daa858e7aa0c139f86cb90acb2/client/k8s.go#L31-L45
Instead, I'd recommend to use one of the functions from https://github.com/kubernetes/kubernetes/blob/7d1daa09383d32fe7a325d6275d9bd51cba45ee5/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L65-L74 which do all the kubeconfig and context resolution for you.

developer-guy:
as a user perspective, the user might not know image version when issued "kubectl rollout history" command, this command shows only the revision number of the deployment, the user needs to issue one more command to get the image version "kubectl rollout history --revision <revision_number>, in our plugin we reduced this step, firstly we showed image versions that previously deployed, then we handed the decision to the user for selecting the image that user wants to rollback, this is the difference between builtin rollout command and our plugin.

corneliusweig:
Hm.. I see how you are trying to make the user experience around rollbacks better. However, I have some concerns.
* First, the plugin is a heavy-weight wrapper that tries to solve a core `kubectl` functionality (`kubectl rollout`) in a slightly different way.
* It is heavy-weight, because it talks to the api-server directly, instead of calling `kubectl` (this could be done from go too). This makes the implementation complex & hard to maintain, and results in quite a large binary.
* Finally, the plugin follows the imperative command idiom, where resources are directly manipulated. Instead it is better practice to put resources under version control and use `kubectl apply -f`.

Altogether, I'm -1 for merging this into the index.

As krew-index matures, it has become more frequent that plugins get rejected. Although this is a set-back, please don't let this discourage you from implementing and submitting your next big idea. And this even does not stop you from distributing your plugin via krew, as we support custom indices now ([1](https://krew.sigs.k8s.io/docs/user-guide/custom-indexes/) & [2](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/)).

Feel free to ping @.ahmetb if you don't agree with this resolution.

ahmetb:
I agree, this one is in the gray area. On  one hand, it's a more human-friendly wrapper on kubectl rollout; on the other hand, it's a quite opinionated wrapper that is unlikely to work for a broad user audience. 

Thankfully, you can still distribute with Krew using custom indexes now, so I'll also recommend starting from there.

PR #791, opened 2020-09-01
Add plugin: edit-status

ulucinar:
This PR proposes a new kubectl plugin named `edit-status` to be indexed in the krew's default plugin index. The plugin can be used to edit `/status` subresources.

### DIFF
--- a/plugins/edit-status.yaml
+++ b/plugins/edit-status.yaml
@@ -1,0 +1,47 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: edit-status
+spec:
+  version: v0.1.0
+  homepage: https://github.com/ulucinar/kubectl-edit-status
+  shortDescription: Edit /status subresources
+  description: |
+    This plugin allows editing /status subresource interactively via a text editor.
+    The editor to be used can be specified via the KUBE_EDITOR or EDITOR
+    environment variables and the default editor is vi.
+  caveats: |
+    * For resources that are not in default namespace, currently you must
+      specify -n/--namespace explicitly (namespace of the current kubeconfig context
+      is not yet used).
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/ulucinar/kubectl-edit-status/releases/download/v0.1.0/kubectl-edit-status_v0.1.0_darwin_amd64.tar.gz
+      sha256: 5dfef266c8db616be3296ebaab5e7cdef85e447c89c7a4588e9f2e065b11c856
+      files:
+        - from: "*"
+          to: "."
+      bin: kubectl-edit_status
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/ulucinar/kubectl-edit-status/releases/download/v0.1.0/kubectl-edit-status_v0.1.0_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


corneliusweig:
Isn't the status field auto-populated? Why is it useful to put a custom message there, if it will be overwritten in the next status update?

ahmetb:
I think depending on the controller implementation, some status conditions might persist between updates, if the controller is smart enough to recognize and not touch the statuses it manages.

ulucinar:
> Isn't the status field auto-populated? Why is it useful to put a custom message there, if it will be overwritten in the next status update?

Hi Cornelius, Ahmet,
Thanks for the review.
The intended use case is updating `status.conditions` (or any `status` field defined) for a custom resource, not just its `status.state`. If a CRD declares `/status` subresource, it is currently not possible to update its `status` fields directly via `kubectl`.

plugins/edit-status.yaml
```diff
@@ -0,0 +1,47 @@
...
+          to: "."
+      bin: kubectl-edit_status
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/ulucinar/kubectl-edit-status/releases/download/v0.1.0/kubectl-edit-status_v0.1.0_linux_amd64.tar.gz
+      sha256: 274c52bcab995f8c3b506cd7b34e44b1bf2b459375497c3460cd16e64ae606fb
+      files:
+        - from: "*"
```
ahmetb, line 35: Can you delete files: ? This is the default behavior anyway.

corneliusweig:
> The intended use case is updating status.conditions (or any status field defined) for a custom resource, not just its `status.state`.

I see, that makes sense. Thanks for explaining. Could you mention in your plugin description that this is targeted at custom resources?

Also, wouldn't it make sense to refuse updating of **standard** API resources, because the API server manages their status? (I'm not sure if your plugin already does that).

plugins/edit-status.yaml
ulucinar: Done

ulucinar:
> > The intended use case is updating status.conditions (or any status field defined) for a custom resource, not just its `status.state`.
> 
> I see, that makes sense. Thanks for explaining. Could you mention in your plugin description that this is targeted at custom resources?

Thank you for the comments. I added a short explanation for the intended use case in the "description" field of the manifest.

> Also, wouldn't it make sense to refuse updating of **standard** API resources, because the API server manages their status? (I'm not sure if your plugin already does that).

No, as of now the plugin does not check if the resource being edited has the `/status` subresource or the API group it belongs to. And I'm not sure if we need to distinguish between built-in API objects and custom resources. For instance, my observation is that if you try to edit (any of the) `status` fields of an object that actually has `/status` subresource, `kubectl` [1] exits with 0 and reports that the object has been modified, whereas the object has actually not been edited.

If a user edits a `status` field (like `status.state`) that will be overridden by a controller (built-in or not), I think it's similar to editing the `spec.replicas` of an apps/v1/ReplicaSet who's controlled by an apps/v1/Deployment. 

What do you you think? I think, I could enumerate `CustomResourceDefinition`s and `APIService`s and allow only editing CRs/aggregated APIs belonging to them, but I'm not sure if that would be worth the effort and more importantly if that would be behaving similar to kubectl.

[1] Observation with `kubectl` 1.19.0.

ahmetb:
I think I agree with the charter of this plugin, it does something that is not doable through kubectl and status fields can be used to store operational data for humans.

I'll go ahead and approve this but @ulucinar hope work on getting user feedback and refining the plugin over time.
/lgtm
/approve

PR #793, opened 2020-09-01
Add reap plugin

micnncim:
Adding the plugin [`kubectl-reap`](https://github.com/micnncim/kubectl-reap) that deletes unused Kubernetes resources, since some users want to install the plugin via Krew (https://github.com/micnncim/kubectl-reap/issues/28).

  - Though looks the name `reap` doesn't follow the naming guide, I think it makes sense since we use the plugin by like `kubectl reap pods` and `kubectl reap configmaps`.
  ```console
  $ kubectl krew install --manifest=plugins/reap.yaml
  Installing plugin: reap
  Installed plugin: reap
  \
   | Use this plugin:
   |      kubectl reap
   | Documentation:
   |      https://github.com/micnncim/kubectl-reap
   | Caveats:
   | \
   |  | This plugin never deletes resources in the `kube-system` namespace.
   | /
  /
  WARNING: You installed plugin "reap" from the krew-index plugin repository.
     These plugins are not audited for security by the Krew maintainers.
     Run them at your own risk.

  $ kubectl krew info reap
  NAME: reap
  URI: https://github.com/micnncim/kubectl-reap/releases/download/v0.10.0/kubectl-reap_0.10.0_darwin_amd64.tar.gz
  SHA256: 9c32ba97d2049167756c4d818fcd8b07bf68e92652a73fd927709a727b52e5b9
  VERSION: v0.10.0
  HOMEPAGE: https://github.com/micnncim/kubectl-reap
  DESCRIPTION:
  This plugin deletes unused Kubernetes resources. It supports Pods,
  ConfigMaps, Secrets, PersistentVolumes, PersistentVolumeClaims,
  Jobs, PodDisruptionBudgets, and HorizontalPodAutoscalers.
  
  CAVEATS:
  \
   | This plugin never deletes resources in the `kube-system` namespace.
  /

  ```


plugins/prune.yaml
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  name: prune
+spec:
+  version: v0.2.0
+  homepage: https://github.com/micnncim/kubectl-prune
+  shortDescription: Prunes unused Kubernetes resources.
+  description: |
+    This plugin prunes unused Kubernetes resources.
+  caveats: |
+    This plugin doesn't delete resources in the `kube-system` namespace.
```
corneliusweig, line 12:
```suggestion
    This plugin never deletes resources in the `kube-system` namespace.
```

plugins/prune.yaml
```diff
@@ -0,0 +1,43 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/micnncim/kubectl-prune/releases/download/v0.2.0/kubectl-prune_0.2.0_darwin_amd64.tar.gz
+    sha256: ab14f9cc7eeaec799a4679455fd371b4c59f454cae8d1af7a516e82aad8e8cc0
+    files:
+    - from: "*"
+      to: "."
```
corneliusweig, line 22: nit: this is the default. You can drop these.

plugins/prune.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: prune
+spec:
+  version: v0.2.0
+  homepage: https://github.com/micnncim/kubectl-prune
+  shortDescription: Prunes unused Kubernetes resources.
+  description: |
+    This plugin prunes unused Kubernetes resources.
```
corneliusweig, line 10: You could consider listing the resource types that can be reaped.

corneliusweig (commented):
This looks like a fine plugin! I see that volumes are not on this list. Maybe that could also be a use-case.

One nit (for your next release): instead of importing the auth plugins individually, you can also do
```go
```

Thanks for contributing!
/approve

ahmetb:
Several problems I notice with the plugin:

1. We already have plugins:
   - [prune-unused](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/prune-unused.yaml)
   - [unused-volumes](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/unused-volumes.yaml)

    If there's already a plugin that does the approximately same thing (and if the value add to the user is minimal), we err on not accepting the new plugin.

2. The word "prune" is too generic, hence violates the naming guidelines.

    For example, one could imagine this plugin "looks at all kubernetes manifests in the current directory and removes everything else on the namespace/cluster". However, this plugin does not do that.

    Similarly, [when `prune-unused` plugin was proposed](https://github.com/kubernetes-sigs/krew-index/pull/162) it was proposed as "prune", and due to the same issue, we had to get it renamed.

micnncim:
@corneliusweig Thank you for your review! I'll fix what you've pointed out.

micnncim:
@ahmetb Thank you for your suggestion.

> 1. We already have plugins:

About this point, it seems `prune-unused` supports only ConfigMaps and Secrets and `unused-volumes` supports only PVCs. My plugin currently supports ConfigMaps, Secrets, and Pods. And also I plan to support other resources like PVC, PDB, HPA, Ingress, and so on. If I support them, would you consider approving this plugin?

> 2. The word "prune" is too generic, hence violates the naming guidelines.

I see. Then, how about another name like `rm-unused`, `snip{-unused}`, `drop{-unused}`or `reap{-unused}`?

corneliusweig:
I think that having a single plugin which can prune all resource types is preferable to having multiple plugins for specific resources. So I'd be supportive of adding this if you can support a superset of resources wrt the existing plugins.

As to naming, I agree that `prune` is generic and also commonly used for that task. However, a generic name also makes sense if many resource types are supported. So maybe use a synonym of `prune` that is less common? I do like `kubectl reap`.

ahmetb:
The real danger with using a verb like `prune` is that some day kubectl might implement it, and it will overshadow your plugin. I think I can also be supportive if this plugin becomes a superset. However, this is a slippery slope, and we don't wanna set a precedent by doing so.

corneliusweig:
This makes sense. As a community we have to decide what to do when a good plugin could be superseded by a better alternative. I think this plugin is not yet there because it does not strictly more than the existing ones, but if @micnncim keeps up his work this is only a question of time.

So I'd suggest we put this on hold for now to give us more time to reason about this case and the general problem outlined above. Right now, it seems that a precondition for merging this plugin is that it handles more resources than the existing alternatives.

/hold

micnncim:
@ahmetb @corneliusweig 

Hi, I've updated `kubectl-prune` (currently [`v0.5.0`](https://github.com/micnncim/kubectl-prune/releases/tag/v0.5.0). This plugin currently supports:

- ConfigMaps (not used in any Pods)
- Secrets (not used in any Pods or ServiceAccounts)
- Pods (whose status is not Running)
- PersistentVolumeClaim (not used in any Pods)
- PodDisruptionBudgets (not targeting any Pods)
- HorizontalPodAutoscalers (not targeting any resources)

I guess this plugin has become a superset of the existing ones, so I'd like to discuss whether I can get approval for this plugin with you again.

And about the naming issue, I'm willing to change this plugin's name if you suggest a proper one. I prefer a simple name like `kubectl-reap` as @corneliusweig suggested though I know `kubectl` might implement such simple names. I'd like to find a name which is simple but unlikely to collide with `kubectl`'s ones.

ahmetb:
I am ok with both prune and reap. That said, reap is less likely to collide with kubectl in the future IMO. The choice is yours.

micnncim:
Thank you! Then, if @corneliusweig also agrees, I'll rename this plugin to `kubectl-reap`. I'd like to confirm that again since remaining will require my time and effort.

micnncim:
And would I get an approval after the name is fixed and I update the manifest?

corneliusweig:
@micnncim I'm definitely supportive of having that plugin. I think `reap` is a neat and safe naming option.

There are still some outstanding comments on the manifest that you could look into. Otherwise, this is ready to get merged. :rocket:

micnncim:
@corneliusweig Thank you!

I looked into some words again and then came up with `tidy`. `tidy` sounds better to me than `reap` since it sounds `reap` means cutting and then gaining but this plugin *gain* nothing. I'm not a native English speaker so I'm not sure my thought makes sense. What do you think of it?

ahmetb:
When I see a plugin named "tidy", deleting resources would be the last thing I imagine it does. :) That's why we tend to clarify verbs with nouns in the naming guide.

micnncim:
Understand it 👍 Then, I'll use `reap` as we've discussed.

micnncim:
I've renamed the plugin name `kubectl-reap` and updated the PR body and the manifest.

corneliusweig:
Congrats, this is a nice one! 🚀

PR #797, opened 2020-09-01
Bump versions in CI scripts

corneliusweig:
- [x] Check that this PR fails initially and complains about a missing license file

Screenshot from 2020-09-02 00-53-06

- [ ] Remove bad commit and wait for green CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,14 +7,14 @@ jobs:
     name: Build and Tags
     runs-on: ubuntu-latest
     env:
-      KREW_VERSION: v0.3.4
+      KREW_VERSION: v0.4.0
       GO111MODULE: on
     steps:
 
-    - name: Set up Go 1.13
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,9 @@ jobs:
       run : |
         set -euo pipefail
         export GOBIN=$HOME/bin
-        go get sigs.k8s.io/krew/cmd/validate-krew-manifest@${KREW_VERSION}
+        # TODO(corneliusweig): pin validator version after next release
```
ahmetb, line 47: what's the reason for this, I'm not sure I get it.

is it because there's no license validator in the tag? if so, can we go-get with a particular commit instead of HEAD?

corneliusweig: Yes, the license check in the validator was merged after releasing v0.4.0, so we need to use some newer revision.

PR #817, opened 2020-09-16
new plugin: kubectl grab-resources

dougsland:
kubectl grab-resources command collects all resources from
the cluster based on label input and export to a single YAML file.

Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>

### DIFF
--- a/plugins/grab-resources.yaml
+++ b/plugins/grab-resources.yaml
@@ -1,0 +1,29 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: grab-resources
+spec:
+  version: "v0.1.0"
+  platforms:
+  - uri: https://github.com/dougsland/kubectl-grab-resources/archive/kubectl-grab-resources-v0.1.0.tar.gz
+    sha256: f837f517897891407ebc01127826c0afe61a86ea382474a8e1e4a7cbb12fe074
+    bin: kubectl-grab_resources
+    files:
+    - from: "./*/kubectl-grab_resources"
+      to: .
+    - from: "./*/LICENSE"
+      to: .
+    selector:
+      matchExpressions:
+      -   {key: os, operator: In, values: [darwin, linux]}
+  caveats: |
+    This plugin needs the following programs:
+    * kubectl neat plugin
+    Examples:
+    Generates the YAML file from app=kibana label
+      kubectl grab-resources -l app=kibana -o resources-label-app_kibana-2020-09-25-19:58:00.yaml
+  description: |
+    kubectl grab-resources command collects all resources from the
+    cluster based on label input and export to YAML file.
+  shortDescription: Export resources based on label input.
+  homepage: https://github.com/dougsland/kubectl-grab-resources


ahmetb:
Please review the naming guidelines. This name is quite generic.

dougsland:
> Please review the naming guidelines. This name is quite generic.

Sure, thanks for the heads up. Just renamed the project. Hope that works now.

plugins/pack-deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+      -   {key: os, operator: In, values: [darwin, linux]}
+  caveats: |
+    This plugin needs the following programs:
+    * kubectl neat plugin
+    Examples:
+    To generate the YAML file from a kibana deployment running in the cluster use:
+      kubectl pack-deploy kibana
+  description: |
+    Detects resources required from running deployments, statefulsets,
+    daemonsets and export it with all dependencies in a single YAML.
```
corneliusweig, line 27: From this description it is very hard to guess what the plugin is actually doing. IIUC, what it does is looking for resources whose names contain the given identifier. I think matching by name is a fragile way to export an application with multiple resources. You have to make it very clear how it works and avoid false promises.

Also, consider that the recommended way to select/group resources in k8s is via labels on the resources. Your plugin should default to that logic and only use the name matching as an explicit option.

plugins/pack-deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+  shortDescription: Export running deployments, statefulsets, daemonsets with all dependencies in a single YAML
```
corneliusweig, line 28: Better stay below ~50chars, everything above will get truncated.

Also, this description is a bit of a false promise. Try to capture the gist of your logic like

> Export all resources with a matching name/label

plugins/pack-deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - from: "./*/kubectl-pack_deploy"
+      to: .
+    - from: "./*/LICENSE"
+      to: .
+    selector:
+      matchExpressions:
+      -   {key: os, operator: In, values: [darwin, linux]}
+  caveats: |
+    This plugin needs the following programs:
+    * kubectl neat plugin
```
corneliusweig, line 21: @itaysk may find this interesting :)

plugins/pack-deploy.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pack-deploy
```
corneliusweig, line 4: I'm thinking if `grab-by-name` would be a better name?

plugins/pack-deploy.yaml
dougsland: hey @corneliusweig thanks for the review. I am aware that we should move to labels. I just learned about kubectl plugins recently  and watched all videos in youtube from @ahmetb  and started doing a plugin (an here we are now) :).
Next version, we will move this to approach you mentioned.

plugins/pack-deploy.yaml
dougsland: Sure, I will reduce to below ~50 chars. 

> Also, this description is a bit of a false promise. Try to capture the gist of your logic like
Not sure if I follow that. gist logic ?

plugins/pack-deploy.yaml
dougsland: Oh yeah, really liked kubectl neat. kudos to @itaysk

plugins/pack-deploy.yaml
dougsland: grab-deploy-by-name?

corneliusweig: The problem with deploy is that it sounds like deployment. Your plugin has a much broader scope than just deployments, so I would avoid to talk about that.

plugins/pack-deploy.yaml
corneliusweig: Gist: in the sense of "essential bits"

plugins/pack-deploy.yaml
dougsland: @corneliusweig 
/me nods. Okay, my last try:  How about grab-yaml-by-name or pack-by-name? 
Sounds reasonable for you to move forward with the current approach or you want before the merge to use labels?

plugins/pack-deploy.yaml
itaysk: awesome! 😃

plugins/pack-deploy.yaml
ahmetb: Unless you already have a solid idea and implementation in mind, I think this should be in krew-index yet. 

You can distribute this using custom indexes if you like.

corneliusweig: @ahmetb I'm a bit confused what you wanted to say. Maybe you forgot 'not'?

 Changing the mechanics of a plugin later on is much harder and may annoy your users. I suggest to make the decision now whether you want to use labels or names for selecting resources. I think that labels are much safer and better aligned with the paradigms of k8s, but ultimately it is your call.

ahmetb:
> /me nods. Okay, my last try: How about grab-yaml-by-name or pack-by-name?
> Sounds reasonable for you to move forward with the current approach or you want before the merge to use labels?

@dougsland I think in general, our problem is with the usability of the plugin. We try to accept plugins that are actually useful to users (this is a curated repository). If this is your first plugin, maybe try to distribute it with [custom indexes](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/).

dougsland:
Hey @ahmetb I am working to polish the plugin and send a new PR.

ahmetb:
> Hey @ahmetb I am working to polish the plugin and send a new PR.

I'd like to reiterate my comment again: You don't have to be on krew-index repository to distribute with Krew. Do not let our restrictions limit you. Start outside krew-index and if you think your plugin fits at a later point, you can re-submit.

dougsland:
@ahmetb @corneliusweig do you mind to re review this effort please?

corneliusweig:
Doesn't that work as well?

```sh
kubectl get all -l app=kibana -o yaml | kubectl neat
```

---

Although your command solves the same task a bit more ergonomically, users will likely prefer the `kubectl` variant, because they know what it's doing and they don't have to learn a new command.

That's the reason why trivial one-liners are not suited as plugins. The value that users get from it is not enough to warrant  learning a new plugin.

dougsland:
> Doesn't that work as well?
> 
> ```shell
> kubectl get all -l app=kibana -o yaml | kubectl neat
> ```

No, it's not the same. It won't collect for example: ingressroute or crd.

> Although your command solves the same task a bit more ergonomically, users will likely prefer the `kubectl` variant, because
> they know what it's doing and they don't have to learn a new command.

From my side, I belive it's hard to predict or assume what users will like or not. It's like use curl into the k8s http API vs kubectl. If one makes the life easier why not use?

> That's the reason why trivial one-liners are not suited as plugins. The value that users get from it is not enough to warrant
> learning a new plugin.

I don't believe that's the case but I will close the PR.

ahmetb:
FWIW we have a plugin named get-all that actually does get those resources I think.

PR #857, opened 2020-10-09
Add kubectl-config-registry plugin

mumoshu:
As explained in README and the plugin description, this plugins allows you to switch between KUBECONFIGs and locate specific KUBECONFIG. The former is useful for managing multiple development, test, and stating clusters and the latter is recommended for operating on production clusters.

As I'm submitting a new plugin, I've confirmed that:

BTW, thank you so much for developing and sharing awesome `kubectx` @ahmetb. As stated in README, `kubeconf` is initially derived from `kubectx`, intending to be used in combination 😃

### DIFF
--- a/plugins/config-registry.yaml
+++ b/plugins/config-registry.yaml
@@ -1,0 +1,54 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: config-registry
+spec:
+  homepage: https://github.com/mumoshu/config-registry
+  shortDescription: Switch between registered kubeconfigs
+  version: v0.2.1
+  description: |
+    Switch between kubeconfigs stored in your local filesystem.
+    Similar to "ctx" and "ns", but for switching the whole kubeconfig.
+
+    The recommended workflow is as follows:
+
+    - Run `init` to set up the registry at `~/.kube/registry`.
+      The registry is where all the registered kubeconfigs will be stored.
+      On `init`, the default kubeconfig is imported as `default`.
+    - Register more kubeconfigs with `import`. `import PATH NAME` registers the
+      kubeconfig at PATH as NAME
+    - Run `ls` to list all the registered configs
+    - Run `use NAME` to switch to the kubeconfig by name
+  caveats: |
+    If fzf is installed on your machine, you can interactively choose
+    between the entries using the arrow keys, or by fuzzy searching
+    as you type.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/mumoshu/config-registry/releases/download/v0.2.1/config-registry_v0.2.1_darwin_amd64.tar.gz
+    sha256: d35481b4d7255c19a1225959cad1b77dd7f525b37d3f02bef4b8f36a16a3af28
+    bin: config-registry
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/mumoshu/config-registry/releases/download/v0.2.1/config-registry_v0.2.1_linux_amd64.tar.gz
+    sha256: 30ac043d20203bc4941454913247ab2b799267c120a43e3401aebe2bdce45b0c
+    bin: config-registry
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm64
+    uri: https://github.com/mumoshu/config-registry/releases/download/v0.2.1/config-registry_v0.2.1_linux_arm64.tar.gz
+    sha256: 8fdeefcc3004b1f41303abbd81647355135c30322b68ecce4dbb58620092be0f
+    bin: config-registry
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/mumoshu/config-registry/releases/download/v0.2.1/config-registry_v0.2.1_windows_amd64.zip
+    sha256: 6b197d0297bd6ef0218d8f2fecb6980f5abd81031909d85ebc52c4e125e56be1
+    bin: config-registry.exe


ahmetb:
/assign @corneliusweig 
For naming considerations.

mumoshu:
FYI I had considered the following options:

- `kubectl kubeconf`: unique but lengthy
- `kubectl config`: conflicts with the `kubectl config` command
- `kubectl cfg`: unique but I personally prefer `conf` (no strong opinion though
-` kubectl conf`: unique

corneliusweig:
He @mumoshu,
thanks for submitting your plugin. Before going into naming considerations, I want to understand the plugin a little better. IIUC, this provides pretty much the same functionality as `ctx`/`kubectx` with a slightly different syntax and better switching between config files. Can you clarify in what ways `conf` is different from `ctx`?
I'm currently thinking, if this plugin is `ctx` plus some convenience, why are you trying to publish a new plugin instead of bringing your improved UX into `kubectx`? That would immediately benefit the huge user-basis that `kubectx` already has.

---

As to naming, I think that `conf` could be mixed up with `config` (both are just short forms of 'configuration'), and I don't even know of the top of my head if `conf` is even an alias for `config` in kubectl (but it's [looking good](https://github.com/kubernetes/kubectl/blob/e95e378e5972064b177a8e71eac2803f55c8c5df/pkg/cmd/config/config.go#L39-L52)).
Besides, conf is a very generic name. Such generic names are usually reserved for plugins with an already large user basis. `ctx` and `ns` are pretty much the only ones where we made that exception.

mumoshu:
@corneliusweig Hey! Thanks for reviewing.

Re: the former part of your comment, `kubeconf` is dedicated to manage kubeconfig files without modification, where `ctx` and `ns` are dedicated to modifying a single kubeconfig file in-place. That's the primary difference, as you may have already noticed.

I just developed a separate tool because it was feasible to complement `ctx` and `ns` without bloating the kubectx codebase. For instance, you can modify specific kubeconfig without affecting another kubeconfig at all with `KUBECONFIG=$(kubeconf locate your-prod-conf) kubectl ctx`.

If there's anyone requesting merging kubeconf into kubectx, I'd try to do so. But I don't have justification to do so.

As I've proposed in README, `kubeconf` especially shines when you have multiple clusters across multiple environments, AND you want a standard workflow that prevents you from running wrong operations on wrong clusters(like removing applications from the production cluster!)

Also, I now understand that `ctx` and `ns` naming were only exceptions.

How about `kubectl confman` then? Does it still sound odd to you?

corneliusweig:
Ok, after having a look at your code I have a better picture of what you built.

IIUC, you built something like a kubeconfig registry where you can check in kubeconfig files and __overwrite__ the default kubeconfig with one from the registry.

This is unsafe, because you don't know if the current kubeconfig was actually checked in. Also, when `use`ing a new config, you just overwrite the registered version on top of the default one. Any local changes like certificate update or context changes to the (old) current config will be lost.

I think your plugin is a valid idea, but you should not compare your plugin to `kubectx`. `kubectx` is a convenience tool to simplify a common task. Your plugin takes over control of the kubeconfig files and stores them all in a central place (from a security point of view, this may not be the best idea btw). So I'd recommend to re-imagine your plugin as a registry for kubeconfig files.

For that you need to make your plugin safer to use though. Losing modifications or overwriting untracked kubeconfig files should be impossible to do by accident. Maybe use hard links for this (https://stackoverflow.com/questions/16800044/create-hardlink-with-golang)? And before touching a config, you should look at its content hash and see if it is known. Makes sense?

WDYT?

mumoshu:
@corneliusweig Thanks a lot for the feedback! I've enhanced it to use hard links instead.

Also, it now checks if `~/.kube/config` has a hard link saved in the registry before switch, so that it never overwrites an unregistered kubeconfig. It seemed enough for preventing accidental overwrites so I didn't use content hash for now.

Also, the plugin is now named `config-registry`. This should be clear enough to denote that it provides the registry of kubeconfig. Also, it should be consistent with an existing plugin that is named `config-cleanup`, which is also related kubeconfigs.

WDYT? Would there be anything else I can do to move this forward?

plugins/config-registry.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: config-registry
+spec:
+  homepage: https://github.com/mumoshu/kubeconf
+  shortDescription: Switch between your kubeconfigs
```
corneliusweig, line 7:
```suggestion
  shortDescription: Switch between registered kubeconfigs
```

plugins/config-registry.yaml
```diff
@@ -0,0 +1,45 @@
...
+  version: v0.2.0
+  description: |
+    Also known as "kubeconf", a utility to switch between kubeconfigs stored in your local filesystem.
+    Similar to "ctx" and "ns", but for switching the whole kubeconfig.
```
corneliusweig, line 11: Please wrap around at ~80 chars, so that it prints nicely on wide and narrow terminals.

I'd recommend to describe the workflow here, so that users understand what is happening:

1) init sets up the registry at `~/.kube/registry` where all registered kubeconfigs will be stored.
2) switch between registered kubeconfigs
3) register additional kubeconfig files.

plugins/config-registry.yaml
```diff
@@ -0,0 +1,45 @@
...
+  caveats: |
+    If fzf is installed on your machine, you can interactively choose
+    between the entries using the arrow keys, or by fuzzy searching
+    as you type.
+    See https://github.com/mumoshu/kubeconf for customization and details.
```
corneliusweig, line 16:
```suggestion
```
No need for this, as the homepage will be shown by default.

corneliusweig (commented):
I think this plugin has reasonable UX now. Please consider renaming your repository to `kubectl-config-registry` or `config-registry` to avoid confusion (speaking out of experience).

mumoshu:
@corneliusweig I've renamed the repository to `config-registry` and fixed the plugin YAML accordingly. Now, it looks much easier to use and understand from a user's perspective. Thanks again for your review 😃

corneliusweig:
Alright, thanks for contributing!
/lgtm
/approve

PR #865, opened 2020-10-13
[plugin] kubectl configrep

loomsen:
This PR introduces `kubectl configrep` plugin. It let's you find configmaps in a namespace or across the cluster, that match a certain pattern. Like `grep -l` for configmaps.


loomsen:
I signed it.

loomsen:
I signed it.

ahmetb:
Thanks for your submission. I'm seeing little utility for this plugin.
It seems to be a small script wrapper for something rather simple such as:

```
for c in $(kubectl get configmaps --all-namespaces -o=name); do
    kubectl get "$c" --jsonpath='$.spec.*~' | grep KEYWORD
done
```

Usually we tend to not accept these sort of plugins, especially if they don't have an obvious use case.

loomsen:
OK no worries.

ahmetb:

PR #871, opened 2020-10-17
Add kubectl allctx plugin

onatm:
This plugin allows you to run `kubectl` `get` command on all the contexts found in `.kube/config`.

**Checklist**
- [x] Read the plugin naming guide
- [x] Installed plugin locally using `kubectl krew install --manifest=plugins/all.yaml`

### DIFF
--- a/plugins/allctx.yaml
+++ b/plugins/allctx.yaml
@@ -1,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: allctx
+spec:
+  homepage: https://github.com/onatm/kubectl-allctx
+  shortDescription: Run commands on contexts in your kubeconfig
+  version: "v1.3.0"
+  description: |
+    This command allows you to run kubectl commands on multiple contexts.
+  platforms:
+    - uri: https://github.com/onatm/kubectl-allctx/releases/download/v1.3.0/bundle.tar.gz
+      sha256: 8b9000864ca72401b57928953d548b85d2bf7c227682c80c0c75db256370d212
+      bin: ./kubectl-allctx
+      files:
+        - from: ./kubectl-allctx
+          to: .
+        - from: ./LICENSE
+          to: .
+      selector:
+        matchExpressions:
+          - key: os
+            operator: In
+            values: ["darwin", "linux"]


onatm:
I am aware `all` is an extremely generic name. However, I could not think of a better name, maybe `get-all` @ahmetb?

plugins/all.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: all
```
corneliusweig, line 4: This is too generic. `forall` might be ok though.

plugins/all.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  homepage: https://github.com/onatm/kall
+  shortDescription: Run 'kubectl get' on all the contexts found in your .kube/config
+  version: "v1.0.0"
+  description: |
+    This command allows you to run get command on multiple contexts' resources.
```
corneliusweig, line 10: Why only `get`? I imagine that all commands could also be useful.

plugins/all.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: all
+spec:
+  homepage: https://github.com/onatm/kall
+  shortDescription: Run 'kubectl get' on all the contexts found in your .kube/config
```
corneliusweig, line 7: Try to stay below 50 chars, otherwise your message will get truncated.

plugins/all.yaml
```diff
@@ -0,0 +1,35 @@
...
+  version: "v1.0.0"
+  description: |
+    This command allows you to run get command on multiple contexts' resources.
+
+    Documentation:
+
+      See https://github.com/onatm/kall
```
corneliusweig, line 14: No need to include that here, as it will be shown by `krew info` anyway.

plugins/all.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+    Usage:
+      # get all pods from all contexts
+      $ kubectl all get pods
+
+      # get all pods that have 'app' in their names from all contexts
+      $ kubectl all get pods --like app
```
corneliusweig, line 21: Examples are better suited for `<plugin> --help`

corneliusweig (commented):
This is pretty borderline for me. Usually we require plugins to be more substantial than bash one-liners.

However, you separate each output into a group with the context name shown on top. So this _might_ be ok.

Let's wait for a second opinion from @chriskim06 or @ahmetb.

plugins/all.yaml
chriskim06: Maybe `Run 'kubectl get' on contexts in your kubeconfig`?

plugins/all.yaml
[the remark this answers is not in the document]
onatm: sounds good 👍

plugins/all.yaml
onatm: I'll shorten it and make it non `get` specific.

plugins/all.yaml
onatm: I thought making it `get` only would be safer but I believe explicitly calling `all` command already states that commands would run for all contexts. I'll create a new release to make it work for all other commands.

onatm:
I made changes based on your comments @corneliusweig @chriskim06 

It became actually better than what I initially imagined by allowing all the commands!

ahmetb:
allctx? 

get-all already exists. "All" would be a no-go for now.

onatm:
I've changed it to forall but forgot to edit PR's title. Do you think allctx is better @ahmetb ?

corneliusweig:
`allctx` is definitely better than `forall`, because it gets to the point what `all` means.

corneliusweig:
One last thought: if you decide on a name, consider also renaming your repository. Having one consistent name for a project is actually a good thing (speaking out of experience).

onatm:
> `allctx` is definitely better than `forall`, because it gets to the point what `all` means.

Fair enough 👍 

> One last thought: if you decide on a name, consider also renaming your repository. Having one consistent name for a project is actually a good thing (speaking out of experience).

What do you recommend for the name? allctx or kubectl-allctx?

ahmetb:
Doesn't matter a lot as long as users can find it with a web search query. You can take a look at other plugin manifests to see what's the convention.

onatm:
I've done the changes as you suggested @ahmetb @corneliusweig. Thank you for your guidance!

corneliusweig:
Nice! Thanks for contributing this handy tool.
/lgtm
/approve

PR #872, opened 2020-10-19
Add halloween plugin

scraly:
This plugin allows you to add randomly a Halloween emoji just before the resources name.
It's a special Halloween mode :-).

### DIFF
--- a/plugins/halloween.yaml
+++ b/plugins/halloween.yaml
@@ -1,0 +1,21 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: halloween
+spec:
+  homepage: https://github.com/scraly/kubectl-halloween
+  shortDescription: Display Hallowen emoji randomly before resources name
+  version: v0.0.2
+  description: |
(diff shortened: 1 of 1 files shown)


ahmetb:
I think custom indexes are a better place to distribute this plugin. 

https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

corneliusweig:
Haha, this is funny :)

However, I agree that krew-index needs to keep its focus on plugins that increase productivity instead of scaring coworkers (^_^);
With custom indexes you can still easily install the plugin.

/close

scraly:
Thanks for your feedbacks.
I know the plugin will be refused (at 99% ) ^^ but I've tried.
I will create my own index :).

FYI, this plugin have been created during the creation of a new talk about Kubernetes plugin creation and distribution through Krew :-)

scraly:
Custom krew-index created: 

https://github.com/scraly/krew-index

Thanks :)

PR #881, opened 2020-10-21
Update tree manifest (remove defaults)

ahmetb:
This manifest was used as an example in
https://krew.sigs.k8s.io/docs/developer-guide/example-manifests/
and is likely the leading cause why people still develop new
plugin manifests with `files:[{from: *, to: .]` mode.

/assign @corneliusweig


ahmetb:

-----

This pull request **does not** seem to be a straightforward version bump. I'll have a human review this.

>file plugins/tree.yaml is not a straightforward version bump: could not find the old/new version spec in the diff

PR #892, opened 2020-10-29
Addition of kopy plugin

TejaBeta:
This PR includes an addition of new kubectl plugin `kopy`.

A simple plugin to copy resources from a particular namespace from one context to another context. 

Tested: installed the new manifest manually and verified that it worked for all the platforms.

### DIFF
--- a/plugins/kopy.yaml
+++ b/plugins/kopy.yaml
@@ -1,0 +1,51 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kopy
+spec:
+  version: "v0.0.1"
+  homepage: https://github.com/TejaBeta/kopy
+  shortDescription: Plugin to copy resources from one context to another
+  description: |
+      kopy is a kubectl plugin or a cli to copy K8s resources from one 
+      context to another context. Only requirement here is to have both 
+      the contexts inside your kube config and appropriate accesses to 
+      the clusters.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/TejaBeta/kopy/releases/download/v0.0.1/kopy_0.0.1_Darwin_x86_64.tar.gz
+    sha256: 32e9a6f924f4144e5baa3fa265c69311b37ef9a574ea41c485653cebeac40c96
+    files:
+      - from: "./LICENSE"
+        to: "."
+      - from: "./kopy"
+        to: "."
+    bin: "kopy"
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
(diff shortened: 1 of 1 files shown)


ahmetb:
Do you mind explaining how's this different than:

    kubectl --context=a get RES/NAME | kubectl eksporter | kubectl apply --context=b -f-

TejaBeta:
`kopy` copies all the resources within a namespace that includes `svc`, `configmaps`, `ingress`, `secrets`, `deployments` and other resources with a single command, however future versions are gone have other improvements with specific resource copy needs. 

While as you need to run this  `kubectl --context=a get RES/NAME | kubectl eksporter | kubectl apply --context=b -f-` cmd may be three or four times to get everything copied which is lengthy. And it's more like `kopy` is not imitating `eksporter`, as `kopy` has its own functionality with everything integrated within it, and while still there is requirement of slight manipulation of resources is part of the project.

ahmetb:
I ask this because, again, we don't want to accept plugins if multiple plugins/builtin commands can be put together.

We currently have a `get-all` plugin that addresses your "all the resources within a namespace" point. 
So, you can replace `get RES/NAME` in my example with `get-all --all-namespaces`, then it ends up doing what your plugin does.

TejaBeta:
Cool, I see your point. 

Kinda `gray-area` for me as well to decide at the moment.

corneliusweig:
I'm also currently -1 here. How I currently see it is that `eksporter` or `neat` are great primitives to remove the clutter from a manifest so that it can be immediately applied in another namespace. What makes these commands great is that they are composable with `kubectl get` and `kubectl apply`. When I copy resources I want to take a look at what's happening first. With these plugins that's as easy as piping the output into `less`. Plus I don't need to learn how  resources are selected with `eksporter`, because this is done with standard `kubectl`.

`kopy` tries to do everything at once: select resources, remove clutter from manifests, and apply in the new namespace. That makes it somewhat rigid, because our users may have a similar but not identical requirement. Don't get me wrong---if copying resources from one namespace to another is what you need to do this can be a great solution. However, I don't think this action is common enough to warrant a dedicated plugin.

So if you want to share your work via `krew`, there is a great alternative now. You can create a custom index and distribute your plugin through that ([1](https://krew.sigs.k8s.io/docs/user-guide/custom-indexes/), [2](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/)). WDYT?

TejaBeta:
Cool! Thank you for your input. I agree with what you said about the rigidity of `kopy` when compared to other existing plugins. Also it completely makes sense from many aspects when compared to other plugins. And I think the comment does gives me great insights on what my next plugin should have or should look like.

corneliusweig:
You have a great mindset! Looking forward to your next ideas.

PR #895, opened 2020-11-06
Add rabbitmq

ansd:
See https://github.com/rabbitmq/cluster-operator/issues/440

### DIFF
--- a/plugins/rabbitmq.yaml
+++ b/plugins/rabbitmq.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: rabbitmq
+spec:
+  homepage: https://github.com/rabbitmq/cluster-operator
+  shortDescription: Manage RabbitMQ clusters
+  version: v0.48.0
+  description: |
+    A plugin providing utilities to operate RabbitMQ clusters deployed
+    by the RabbitMQ cluster operator.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/rabbitmq/cluster-operator/archive/0.48.0.tar.gz
+    sha256: 82d09f2bd1a6744549b871e9823b7008f40eca9d7e6c0fd16d028f2ea520c0a9
+    bin: kubectl-rabbitmq
+    files:
+    - from: cluster-operator-*/bin/kubectl-rabbitmq
+      to: .
+    - from: cluster-operator-*/LICENSE.txt
+      to: .
--- a/plugins/rabbitmq.yaml
+++ b/plugins/rabbitmq.yaml
@@ -9,6 +9,9 @@ spec:
   description: |
     A plugin providing utilities to operate RabbitMQ clusters deployed
     by the RabbitMQ cluster operator.
+  caveats: |
+    This plugin requires the RabbitMQ cluster operator to be installed.
+    To install the RabbitMQ cluster operator run `kubectl rabbitmq install-cluster-operator`.
   platforms:
   - selector:
       matchExpressions:


ansd:
Signed the CLA.

plugins/rabbitmq.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Plugin
+metadata:
+  name: rabbitmq
+spec:
+  homepage: https://github.com/rabbitmq/cluster-operator
+  shortDescription: Manage RabbitMQ clusters
+  version: v0.48.0
+  description: |
+    A plugin providing utilities to operate RabbitMQ clusters deployed
+    by the RabbitMQ cluster operator.
```
corneliusweig, line 11: I see that you offer to install the cluster operator via this tool. Maybe mention this as a post-install setup step here or in the `caveats` section. Usually we don't recommend to print basic usage instructions, but this is different, because the plugin only becomes useful once the operator is installed.

plugins/rabbitmq.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: rabbitmq
+spec:
+  homepage: https://github.com/rabbitmq/cluster-operator
+  shortDescription: Manage RabbitMQ clusters
```
corneliusweig, line 7:
```suggestion
  shortDescription: Manage RabbitMQ operator
```
IIUC, this plugin is about interacting with the operator and not with RabbitMQ itself.

plugins/rabbitmq.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: rabbitmq
```
corneliusweig, line 4: This name makes sense and seems justified given your membership in the project. :+1:

corneliusweig (commented):
It's great to see notable projects adopting `krew` as a distribution platform!
I have minor suggestions about the manifest, but this is looking great already.

---

I see that you plan to adopt `krew-release-bot`. Today I learnt about a great tool to make this zero-friction: https://rajatjindal.com/tools/krew-release-bot-helper/

plugins/rabbitmq.yaml
ansd: I added a `caveats` section.

plugins/rabbitmq.yaml
ansd: This plugin has one command `install-cluster-operator` which installs the operator.
Most other commands are used to interact with the RabbitMQ clusters (the custom resources deployed by the operator) directly. 
For example `manage` opens the Management UI for a RabbitMQ cluster, `debug` enables debug logging for a RabbitMQ cluster, `enable-all-feature-flags` enables feature flags for a RabbitMQ cluster.
Other commands, e.g. `create`, `list`, and `delete` interact more with the operator but might arguably also be part of "managing" RabbitMQ clusters? 🤔
I would prefer `Manage RabbitMQ clusters` but would also be happy to rename to `Manage RabbitMQ operator`.

ansd:
Thank you @corneliusweig for reviewing and for this tool 🙂.
I really like the ease of use of krew for both users and developers.
I'll try out https://rajatjindal.com/tools/krew-release-bot-helper/.

plugins/rabbitmq.yaml
corneliusweig: Ok, thanks for explaining.
My main concern about `cluster` is that it is a heavily overloaded term. In particular, the k8s folks usually think about the *kubernetes* cluster as a whole and less about a rabbitmq cluster. Given the circumstances, it looks this ambiguity is inevitable.

corneliusweig (commented):
/lgtm
/approve
Thanks for contributing!

PR #926, opened 2020-11-18
Add kubectl applythis plugin

turkenh:
A simple kubectl plugin that allows to create/apply Kubernetes resources directly from editor.

More details:

https://github.com/turkenh/kubectl-applythis/blob/main/README.md

### DIFF
--- a/plugins/applythis.yaml
+++ b/plugins/applythis.yaml
@@ -1,0 +1,25 @@
+
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: applythis
+spec:
+  homepage: https://github.com/turkenh/kubectl-applythis
+  shortDescription: Directly create/apply yaml files without writing into file.
+  version: v0.1.0
+  description: |
+    A simple kubectl plugin that allows to create/apply Kubernetes resources directly from editor.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/turkenh/kubectl-applythis/releases/download/v0.1.0/kubectl-applythis-v0.1.0.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
@turkenh it seems that this is basically just `paste | kubectl apply -f-` with a chance to edits.
Typically we don't accept plugins that shorten an already simple command pipeline, so I'll mark this as "falls into gray area".

This plugin could be implemented with a bash function like:


```
file=$(mktemp)
"$EDITOR" "$file"
kubectl apply -f "$file"
rm "$file"
```

which is also another reason that it might not warrant a plugin in krew-index. If we end up rejecting it, you can still distribute it by hosting your custom plugin index.

/assign @corneliusweig 
for additional opinions.

turkenh:
@ahmetb I see, sounds good. However, I want to add that those bash steps might not be obvious for anyone that want to achieve this. So, I still see some advantage with having this a shared plugin.

I wanted to share this as a plugin after I found myself keep creating temporary files to create/apply kubernetes resources :)

corneliusweig:
Hi @turkenh ,
thanks for submitting your plugin to `krew-index`. Although this is a handy trick if you apply modified manifests a lot, I think it is not suited for `krew-index`.
For one, modifying manifests on-the-fly is not great for reproducibility and version control. It is generally easier to reason about your cluster if you only apply checked-in versions.
Besides, I would see this feature more in the realm of editors and not as much a kubectl plugin. For example, you could also write this as a vim command `:w !kubectl apply -f -`.

I think your util is best suited for collections of shell aliases and would fit nicely there. If you prefer to publish as kubectl plugin, I'd recommend to submit to a custom index such as https://github.com/ishantanu/awesome-kubectl-plugins#use-awesome-kubectl-plugins-as-a-custom-index-via-krew.

---

Super-nit: if you stick with a plugin, try to come up with a name that also signals the modification part. For example, `modapply` sounds better to me.

/close

PR #937, opened 2020-11-27
Add kubectl-env plugin

almariah:
Add kubectl-env plugin

https://github.com/almariah/kubectl-env

### DIFF
--- a/plugins/env.yaml
+++ b/plugins/env.yaml
@@ -1,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: env
+spec:
+  version: v1.0.0
+  homepage: https://github.com/almariah/kubectl-env
+  shortDescription: Allow templating kustomize manifests using environment variables
+  description: |
+    The kubectl-env plugin will allow templating kustomize manifests using environment variables.
+  platforms:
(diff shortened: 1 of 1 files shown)


ahmetb:
Please read the naming guide. This plug-in doesn't do what I would expect it to do despite the name "env". That's a sign you've chosen a generic yet vague name. Please consider renaming. 
/hold

corneliusweig:
Thanks for considering publication as a `krew` plugin.

As far as I can tell, there is significant overlap with established commands, like `envsubst` (see https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html). I think users will have more confidence in established and vetted utils, so there is no user value in adding this command right now.
Feel free to explain how your command is different then I'll take another look.

Although this might feel like a set-back, please don't let this discourage you from implementing and submitting your next big idea. And this does not even stop you from distributing your plugin via krew, as we support custom indices now ([1](https://krew.sigs.k8s.io/docs/user-guide/custom-indexes/), [2](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/)). So you can instead submit this manifest to another index such as https://github.com/ishantanu/awesome-kubectl-plugins#use-awesome-kubectl-plugins-as-a-custom-index-via-krew.

/close

PR #956, opened 2020-12-11
Create kconnect.yaml

seankelly001:


ahmetb:
Plugin name violates nearly every rule we have in the naming guide. It’s reserving a generic word and not articulate about what the plugin does. Please read the naming guide, look at other plugins, and consider renaming your plugin. 

Alternatively, you can still distribute your plugin via Krew on your custom inde: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index. That way you don’t have to rename.

PR #968, opened 2020-12-23
add janitor plugin v0.1.0

dastergon:
This PR introduces the `janitor` plugin. A `kubectl` plugin to assist in finding Kubernetes objects in a problematic state.

## Actions taken

PTAL

### DIFF
--- a/plugins/janitor.yaml
+++ b/plugins/janitor.yaml
@@ -1,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: janitor
+spec:
+  version: v0.1.0
+  homepage: https://github.com/dastergon/kubectl-janitor
+  shortDescription: Find objects in a problematic state
+  description: |
+    This plugin assists in finding
+    Kubernetes objects in a problematic state.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/dastergon/kubectl-janitor/releases/download/v0.1.0/kubectl-janitor_v0.1.0_darwin_amd64.tar.gz
+    sha256: b991c1c80c58c4563b5718babb8345ac9d7bfe98e08dcbf633a7a7882f447826
+    bin: kubectl-janitor
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/dastergon/kubectl-janitor/releases/download/v0.1.0/kubectl-janitor_v0.1.0_linux_amd64.tar.gz
+    sha256: a2d1d34e4173569cff0b6804c93376c535e9a33e939e44b8da3111d48eb3a6ab
+    bin: kubectl-janitor
+  - selector:
+      matchLabels:
+        os: windows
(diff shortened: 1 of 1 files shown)


plugins/janitor.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: janitor
+spec:
+  version: v0.1.0
+  homepage: https://github.com/dastergon/kubectl-janitor
+  shortDescription: Find objects in a problematic state
```
corneliusweig, line 8:
```suggestion
  shortDescription: Lists objects in a problematic state
```

plugins/janitor.yaml
```diff
@@ -0,0 +1,34 @@
...
+  description: |
+    This plugin assists in finding
+    Kubernetes objects in a problematic state.
```
corneliusweig, line 11: Maybe add

> It currently supports Job, Pod, PersistentVolume, and PersistentVolumeClaim resources.

corneliusweig (commented):
This looks like a useful plugin for diagnosing typical failure modes. The manifest looks pretty clear and the name is acceptable.

However, there's a pretty plugin that solves the same problem statement in krew already: https://github.com/bergerx/kubectl-status
So I'm not yet sure whether we should merge this plugin to the central krew index (you also have the option now to run private indexes).

Maybe you can explain where your plugin is more useful than `kubectl-status`? It would also make sense to mention your unique selling points in the `description`.

dastergon:
Hi @corneliusweig, Thanks a lot for the feedback! 

I wasn't aware of [`kubectl-status`]( https://github.com/bergerx/kubectl-status).  At first sight, the plugin seems similar to `janitor` but, I believe both tools have a different focus.
 
According to `kubectl-status`' [README file](https://github.com/bergerx/kubectl-status/blame/master/README.md#L104), the output is aimed to be consumed from humans only and not be that parse-friendly.  In contrast, `janitor` is less opinionated with its output, is parse-friendly, and has no color coding. Also, it extends cases where `--field-selector` is limited.  The output follows the same output k8s users expect from a command such as `kubectl get`. 

Unlike `kubectl-status`, the primary goal of `janitor` is not to replace `kubectl describe` or show everything in one compact page. Instead, extend kubectl with some command patterns around finding resources in a problematic state. The idea is that during troubleshooting the user can quickly find some of the problems (i.e., unbound claims, available volumes, broken jobs, etc.). Then, the user can explore further using the native kubectl commands such as `describe` or `event`.  Consequently, the user won't lose time trying to find all the different kubectl one-liners and options for `--field-selector`, `jsonpath`, `jq`, `uniq`, etc. 

I wouldn't say that the plugins are mutually exclusive, but one could complement the other. However, I wouldn't consider merging them since they have different goals.

corneliusweig:
Ok thanks for clarifying, this makes sense to me. I think we can go ahead and merge, as the manifest looks good as well.

/assign chriskim06
For a second opinion

chriskim06:
@dastergon thanks for the contribution!

/lgtm
/approve

PR #996, opened 2021-01-23
Add kubeSize (capacity) as a plugin for krew to install

akrzos:
Add [kubeSize](https://github.com/akrzos/kubeSize) a kubectl capacity and sizing tool.

Output from test install:
```
$ kubectl krew install --manifest=deploy/krew/capacity.yml
Installing plugin: capacity
Installed plugin: capacity
\
 | Use this plugin:
 | 	kubectl capacity
 | Documentation:
 | 	https://github.com/akrzos/kubeSize
/
$ kubectl capacity c
NODES                     PODS                                      CPU (cores)                                   MEMORY (GiB)
Total Ready Unready Unsch Capacity Allocatable Total Non-Term Avail Capacity    Allocatable Requests Limits Avail Capacity     Allocatable Requests Limits Avail
1     1     0       0     110      110         9     9        101   4.0         4.0         0.8      0.1    3.1   1.9          1.9         0.2      0.4    1.8
```

### DIFF
--- a/plugins/capacity.yaml
+++ b/plugins/capacity.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capacity
+spec:
+  version: "v0.1.2"
+  homepage: https://github.com/akrzos/kubeSize
+  shortDescription: "Get cluster size and capacity"
+  description: |
+    Show counts of nodes and pods. Aggregates max-pod, cpu, and memory capacity.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/akrzos/kubeSize/releases/download/v0.1.2/kubeSize_0.1.2_macOS_x86_64.tar.gz
+    sha256: afb8853d1113f80343bc35aea99637a851e99dda121fb45ad351c5d1040dee5b
+    bin: kubectl-capacity
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/akrzos/kubeSize/releases/download/v0.1.2/kubeSize_0.1.2_Linux_x86_64.tar.gz
+    sha256: d1e680f66c66ca7f9664bf2efcdcdf4e9580b565de45152c2cba2f208b457df6
+    bin: kubectl-capacity
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/akrzos/kubeSize/releases/download/v0.1.2/kubeSize_0.1.2_Windows_x86_64.tar.gz
+    sha256: 743c86196b826e63cdc44bd05709ba93538637b0a5698ca481bdcc90bf8daaae
+    bin: kubectl-capacity.exe
+


ahmetb:
I think the word `capacity` does not go well with our Naming Guide linked above. We have similar plugins that deal with capacity/utilization analysis, please take a look at their names.

plugins/capacity.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capacity
```
corneliusweig, line 4: Maybe?
```suggestion
  name: cluster-stat
```

plugins/capacity.yaml
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  version: "v0.1.1"
+  homepage: https://github.com/akrzos/kubeSize
+  shortDescription: "Get cluster size and capacity"
+  description: |
+    Exposes size and capacity data for Kubernetes clusters
```
corneliusweig, line 10: It's not clear to me what size and capacity refers to. Number of cores? Memory limit? Number of nodes?
Please be a bit more specific about what numbers you show.

akrzos: Sure I will improve the description.  The main idea was to help provide quick sensible numbers that can determine if you could fit a pod/application based on its schedule-able requirements on the cluster or worker node space.

akrzos:
> I think the word `capacity` does not go well with our Naming Guide linked above. We have similar plugins that deal with capacity/utilization analysis, please take a look at their names.

I see a few plugins which overlap somewhat in functionality and output however they lack any introspection on max-pods and non-terminated pod counts.  My plugin doesn't seem to have any sort of name conflict so the only guideline I think it could be subject to would be the "Use Verbs/Resource Types".  Would `view-capacity` be a name you would agree with and allow my plugin to be in the krew index?

plugins/capacity.yaml
corneliusweig: I think that a plugin that determines if you can fit an application with certain requirements into a cluster is a very good idea. However, the approach you take can be improved.

At the moment, your plugin shows certain capacity metrics in the cluster. So the users need to look at their own manifests to compare these numbers and check if it fits or not. But your plugin could do this check for the user!

So imagine you had a plugin called `will-it-fit`. You run `$ kubectl will-it-fit my-deployment.yaml`, and the plugin says `yes` or `no`. Isn't that simple? This will be much more useful and a much better fit than having yet another capacity/cluster-stats plugin.

corneliusweig:
`view-capacity` is certainly better, but there are still two problems:

- there is still significant overlap with existing plugins.
- It is unclear what "capacity" means.

akrzos:
> `view-capacity` is certainly better, but there are still two problems:
> 
> * there is still significant overlap with existing plugins.
> * It is unclear what "capacity" means.

Of course the term as broad as capacity means a lot of things in a Kubernetes cluster.  In terms of this plugin, capacity means how much available "space" you have according to pod count, schedulable cpu, and schedulable memory. (Of course there are many other metrics and/or problems in the 'can it fit' capacity space for kubernetes) This plugin exposes all three of those  metrics is several groupings to help you determine if you could fit something.

For Example, you could examine this at the "cluster" level:
```console
$ kubectl capacity c
NODES                     PODS                                      CPU (cores)                                   MEMORY (GiB)
Total Ready Unready Unsch Capacity Allocatable Total Non-Term Avail Capacity    Allocatable Requests Limits Avail Capacity     Allocatable Requests Limits Avail
3     3     0       0     330      330         13    13       317   12.0        12.0        1.1      0.3    10.9  5.8          5.8         0.3      0.5    5.5
```
We can see this cluster has space for 317 pods, 10.9 schedulable cpu cores, and 5.5GiB of schedulable memory.

However as you might have already guessed, not all nodes can host a workload, so maybe we need to examine just the space on nodes with a specific role.  Thus the need for the node-role capacity command as:
```console
$ kubectl capacity nr
ROLE   NODES                     PODS                                      CPU (cores)                                   MEMORY (GiB)
       Total Ready Unready Unsch Capacity Allocatable Total Non-Term Avail Capacity    Allocatable Requests Limits Avail Capacity     Allocatable Requests Limits Avail
<none> 2     2     0       0     220      220         7     7        213   8.0         8.0         0.4      0.2    7.6   3.9          3.9         0.2      0.4    3.6
master 1     1     0       0     110      110         6     6        104   4.0         4.0         0.7      0.1    3.4   1.9          1.9         0.0      0.0    1.9
```
Since likely we are not placing our workload on master nodes, we can examine the available pods, cpu, and memory on the "\<none\>" role nodes in this case.(This was a kind cluster which didn't label the roles as worker or compute etc)  So we can actually fit 213 pods, 7.6 cpu cores, and 3.6GiB of memory instead of what we saw in the cluster command.

Lastly, maybe your workload is one large pod so you need to look at available resources on the nodes themselves and determine if the workload is schedulable.  You can use the node command to help here:
```console
$ kubectl capacity no
NAME                STATUS ROLES  PODS                                      CPU (cores)                                   MEMORY (GiB)
                                  Capacity Allocatable Total Non-Term Avail Capacity    Allocatable Requests Limits Avail Capacity     Allocatable Requests Limits Avail
3node-control-plane Ready  master 110      110         6     6        104   4.0         4.0         0.7      0.1    3.4   1.9          1.9         0.0      0.0    1.9
3node-worker        Ready  <none> 110      110         3     3        107   4.0         4.0         0.2      0.1    3.8   1.9          1.9         0.1      0.2    1.8
3node-worker2       Ready  <none> 110      110         4     4        106   4.0         4.0         0.2      0.1    3.8   1.9          1.9         0.1      0.2    1.8
```

This is just one aspect that the plugin strives to solve.  I will continue to build on this plugin to add more "capacity" like problems it can help solve.  The namespace sub-command already helps me quickly gather a bunch of namespace "size" requirements by showing me non-terminated pod counts, requests.cpu and requests.memory per namespace.  I intend on eventually answering more capacity questions with this plugin as well with more sub-commands.  I also have a desire to understand a cluster's "size", I consider this the number of nodes, namespaces, pods, containers, deployments, configmaps, secrets, and objects in general. I will add a sub-command to address that.

I hope that helps you understand my goals with this plugin and how it is different than the existing plugins.  Is it the intention of krew plugins to only have one plugin be available in an attempt to solve a specific problem or will krew allow more than one plugin per potential problem/functionality space?

ahmetb:
@corneliusweig please re-review when you have time.

corneliusweig:
Thanks for your detailed response!

However, my main concern still stands. This plugin has very high overlap with existing plugins (https://github.com/davidB/kubectl-view-allocations, https://github.com/robscott/kube-capacity, https://github.com/etopeter/kubectl-view-utilization). I was trying to understand where your plugin stands out, but AFAICT the differences are pretty subtle. So taking the viewpoint of a random user who looks for such a plugin: it does not help if they have 4 plugins for the same task, right?

So considering our users, I think it does not add value if we add this plugin.

---

You seem to have more ideas how to improve this plugin. If you are at a point where this plugin does strictly more than the existing ones, let's revisit this.

/cc @ahmetb @chriskim06 
For a second opinion.

ahmetb:
1. **Naming:** we previously rejected a plugin named `capacity` (https://github.com/robscott/kube-capacity), because we don't want to give people first-comers advantage on a word that they might not be able to deliver on.

2. **Functionality:** I thought only difference between this plugin vs `view-allocations`+`resouce-capacity` is that it shows **pod counts** per namespace/node/cluster-wide (in addition to CPU/memory). But the `view-allocations` plugin actually already shows pod counts as well (and it's also groupable per node/namespace/cluster-wide etc). I'm attaching a screenshot for reference.

   As far as I'm seeing, I don't think this plugin offer any superior functionality that the `view-allocations` plugin doesn't have. 

**view-allocations:**

kubectl view-allocations
kubectl view-allocations


**Proposed plugin:**

proposed plugin
proposed plugin

ahmetb:
I'm closing this since there's not enough value-add with existing plugins we have.

Thanks for your submission.

/close

PR #997, opened 2021-01-24
Add azad-proxy to krew-index

simongottschlag:
More info: https://github.com/xenitab/azad-kube-proxy

### DIFF
--- a/plugins/azad-proxy.yaml
+++ b/plugins/azad-proxy.yaml
@@ -41,10 +41,18 @@ spec:
         os: windows
         arch: amd64
   version: "v0.0.3"
+  homepage: https://github.com/xenitab/azad-kube-proxy
+  caveats: |
+    This plugin is used in combination with the azad-kube-proxy:
+    - https://github.com/xenitab/azad-kube-proxy
+
+    There shouldn't be a use case where you can use the plugin without the proxy.
+    If you haven't been asked by a cluster admin to install this plugin - it may
+    not be the one you need.
   shortDescription: Generate and handle authentication for azad-kube-proxy
   description: |
-    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to generate
-    the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
-    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD
-    groups for users with more than 200 of them.
-  homepage: https://github.com/xenitab/azad-kube-proxy
+    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy)
+    to make it easier to generate the kubeconfig and handles the authentication.
+    The proxy itself enables you to use Azure AD to authenticate to any Kubernetes
+    API, decouples you from the Cloud providers FQDNs and solves issues related 
+    to Azure AD groups for users with more than 200 of them.
--- a/plugins/azad-proxy.yaml
+++ b/plugins/azad-proxy.yaml
@@ -43,8 +43,8 @@ spec:
   version: "v0.0.3"
   shortDescription: Generate and handle authentication for azad-kube-proxy
   description: |
-    Azure AD Kubernetes API Proxy: azad-kube-proxy
-    A plugin that works together with the proxy to make it easier to generate the kubeconfig and handles the authentication.
-    The proxy itself enables you to use Azure AD to authenticate to any Kubernetes API, decouples you from the Cloud
-    providers FQDNs and solves issues related to Azure AD groups for users with more than 200 of them.
+    A plugin that works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to 
+    generate the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
+    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD groups for
+    users with more than 200 of them.
   homepage: https://github.com/xenitab/azad-kube-proxy
--- a/plugins/azad-proxy.yaml
+++ b/plugins/azad-proxy.yaml
@@ -43,8 +43,8 @@ spec:
   version: "v0.0.3"
   shortDescription: Generate and handle authentication for azad-kube-proxy
   description: |
-    A plugin that works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to 
-    generate the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
-    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD groups for
-    users with more than 200 of them.
+    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to generate
+    the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
+    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD
+    groups for users with more than 200 of them.
   homepage: https://github.com/xenitab/azad-kube-proxy
--- a/plugins/azad-proxy.yaml
+++ b/plugins/azad-proxy.yaml
@@ -1,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: azad-proxy
+spec:
+  platforms:
+  - uri: https://github.com/XenitAB/azad-kube-proxy/releases/download/v0.0.3/kubectl-azad_proxy-darwin-amd64.zip
+    sha256: 51ae0098cbbb0b5da01f5b086c02474a8b74012e5f373c2dbeea36ad659c65e2
+    bin: kubectl-azad_proxy
+    files:
+    - from: kubectl-azad_proxy
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  - uri: https://github.com/XenitAB/azad-kube-proxy/releases/download/v0.0.3/kubectl-azad_proxy-linux-amd64.zip
+    sha256: 323182ba7960e3b587c5d7b6971e0aaad29c0d5d1acb4d276e20c7f7c5c3ab98
+    bin: kubectl-azad_proxy
+    files:
+    - from: kubectl-azad_proxy
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  - uri: https://github.com/XenitAB/azad-kube-proxy/releases/download/v0.0.3/kubectl-azad_proxy-windows-amd64.zip
+    sha256: 5556aa4a6b679fe8e33639fd9df008290f2a3b9f0b965dc2fcdde0c913721233
+    bin: kubectl-azad_proxy.exe
+    files:
+    - from: kubectl-azad_proxy.exe
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v0.0.3"
+  shortDescription: Plugin to generate and handle authentication for azad-kube-proxy
+  description: |
+    Azure AD Kubernetes API Proxy: azad-kube-proxy
+    This is a plugin to make authentication and rotation of tokens from Azure AD simpler
+    when using azad-kube-proxy.
+  homepage: https://github.com/xenitab/azad-kube-proxy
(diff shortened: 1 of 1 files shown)


plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - from: kubectl-azad_proxy.exe
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v0.0.3"
+  shortDescription: Plugin to generate and handle authentication for azad-kube-proxy
```
ahmetb, line 44: "Plugin to" redundant. It's clear this is a plugin :)

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,49 @@
...
+  description: |
+    Azure AD Kubernetes API Proxy: azad-kube-proxy
```
ahmetb, line 46: This line seems redundant. Maybe use this opportunity to explain what your proxy achieves (I can't tell from this description, I probably should be)

ahmetb, line 46: Probably say ActiveDirectory here?

ahmetb:
Out of curiosity: is azad a common acronym for Azure ActiveDirectory? I typically see "aad" being used.

simongottschlag:
> Out of curiosity: is azad a common acronym for Azure ActiveDirectory? I typically see "aad" being used.

Hi!

No, the most common abbriviation is AAD. There are a few other services with names like AAD Proxy and I don't want to cause any confusion or irritation with Microsoft. aad-kube-proxy would have been the best name, but went with this to try and leave aad-proxy to Microsoft.

plugins/azad-proxy.yaml
simongottschlag: @ahmetb fixed!

plugins/azad-proxy.yaml
simongottschlag: @ahmetb does it look better now? :+1:

simongottschlag: @ahmetb I'm using Azure AD in the documentation and I think that's the most common way to refer to it. At least what I'm used to.

OK to keep it?

ahmetb:
One more question: do users use this plugin one-off or on a continuing basis (e.g. perhaps to launch the proxy)?

/cc @corneliusweig for another pair of eyes on naming/fit review.

simongottschlag:
> One more question: do users use this plugin one-off or on a continuing basis (e.g. perhaps to launch the proxy)?
> 
> /cc @corneliusweig for another pair of eyes on naming/fit review.

@ahmetb The users will use it when connecting to new cluster:
```shell
kubectl azad-proxy generate --cluster-name dev-cluster --proxy-url https://dev.example.com --resource https://dev.example.com
```

This, in the background, will create a kubeconfig that uses `kubectl azad-proxy login` all the time:
```YAML
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: [...]
    server: https://dev.example.com
  name: dev-cluster
contexts:
- context:
    cluster: dev-cluster
    user: dev-cluster
  name: dev-cluster
current-context: dev-cluster
kind: Config
preferences: {}
users:
- name: dev-cluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - azad-proxy
      - login
      command: kubectl
      env:
      - name: CLUSTER_NAME
        value: dev-cluster
      - name: RESOURCE
        value: https://dev.example.com
      - name: TOKEN_CACHE
        value: ~/.kube/azad-proxy.json
      - name: EXCLUDE_AZURE_CLI_AUTH
        value: "false"
      - name: EXCLUDE_ENVIRONMENT_AUTH
        value: "false"
      - name: EXCLUDE_MSI_AUTH
        value: "false"
      provideClusterInfo: false
```

simongottschlag:
> One more question: do users use this plugin one-off or on a continuing basis (e.g. perhaps to launch the proxy)?
> 
> /cc @corneliusweig for another pair of eyes on naming/fit review.

@ahmetb I just pushed some updated documentation as well if you want some more. It's still a fresh project, but won't be fun using without this plugin. :)

You can find the latest docs here: https://github.com/XenitAB/azad-kube-proxy

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+kind: Plugin
+metadata:
+  name: azad-proxy
+spec:
+  platforms:
+  - uri: https://github.com/XenitAB/azad-kube-proxy/releases/download/v0.0.3/kubectl-azad_proxy-darwin-amd64.zip
+    sha256: 51ae0098cbbb0b5da01f5b086c02474a8b74012e5f373c2dbeea36ad659c65e2
+    bin: kubectl-azad_proxy
+    files:
+    - from: kubectl-azad_proxy
```
corneliusweig, line 11: jfyi, no need to change: The binary name can be anything here---krew users will always get the plugin executable as given by the `metadata.name` field.

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+        os: windows
+        arch: amd64
+  version: "v0.0.3"
+  shortDescription: Generate and handle authentication for azad-kube-proxy
+  description: |
+    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to generate
+    the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
+    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD
+    groups for users with more than 200 of them.
+  homepage: https://github.com/xenitab/azad-kube-proxy
```
corneliusweig, line 50: Maybe end this line with `\n`?

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v0.0.3"
+  shortDescription: Generate and handle authentication for azad-kube-proxy
+  description: |
+    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to generate
+    the kubeconfig and handles the authentication. The proxy itself enables you to use Azure AD to authenticate
+    to any Kubernetes API, decouples you from the Cloud providers FQDNs and solves issues related to Azure AD
+    groups for users with more than 200 of them.
```
corneliusweig, line 49: Please wrap this message at ~80chars so that it prints nicely on wide an narrow terminals.

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: azad-proxy
```
corneliusweig, line 4: If the plugin remains focused on handling authn via Azure AD, maybe call it `azad-auth` instead?

corneliusweig (commented):
Hey, thanks for your submission. This looks like a useful plugin when using the `azad-kube-proxy`. My understanding is that every authentication calls `azad-proxy login`, so even if the plugin is only called once to set up authentication to a cluster, it is used all the time. So this definitely qualifies as plugin to me 👍 

I think your name is ok-ish, but I'm not familiar with how tightly the term `azad` is tied to Microsoft. If this association is strong, it might result in trouble. Otherwise, `azad-auth` seems even more fitting for the current feature set. Besides there may be naming confusion with other `azad` tools, like `azad-kube-proxy`.

plugins/azad-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: "v0.0.3"
+  shortDescription: Generate and handle authentication for azad-kube-proxy
+  description: |
+    Works together with the proxy (Azure AD Kubernetes API Proxy: azad-kube-proxy) to make it easier to generate
```
corneliusweig, line 46: This sounds like users need to have `azad-kube-proxy` installed. So please include a link where they can download this executable, and add a `caveat` section about this.

simongottschlag: Hi @corneliusweig!

This plugin is only used in combination with the proxy (and both are located in the GitHub link).

I don't see people using the plugin and then going for the proxy - wouldn't do much for them.

In this case, I am pointing the users of the proxy here to easily get the plugin.

Do you still want me to add more? (Since the GitHub link is already there)

simongottschlag:
> Hey, thanks for your submission. This looks like a useful plugin when using the `azad-kube-proxy`. My understanding is that every authentication calls `azad-proxy login`, so even if the plugin is only called once to set up authentication to a cluster, it is used all the time. So this definitely qualifies as plugin to me 👍 
> 
> 
> 
> I think your name is ok-ish, but I'm not familiar with how tightly the term `azad` is tied to Microsoft. If this association is strong, it might result in trouble. Otherwise, `azad-auth` seems even more fitting for the current feature set. Besides there may be naming confusion with other `azad` tools, like `azad-kube-proxy`.

@corneliusweig I chose azad because it's not commonly used - aad is what people are using.

I was thinking of something more generic than azad-proxy, but then people may think they can use the plugin without the proxy (which maybe would be possible, but not a use case I'm going to be actively looking into).

With the above in mind, trying not to use the common name for AAD and not confuse people that this is a generic plugin - but one that requires something on the server side - do you think it's ok?

Thanks!

corneliusweig:
Thanks for clarifying this a bit. For example, I didn't see that this is the same as `azad-kube-proxy`.

Overall, it more and more looks to me, that `proxy` is not the ideal name here. To me it sounds like this should be `azad-auth`. After what you wrote, I think `azad` is better than `aad`.

I'm not fully sure I understand your comment about server-side requirements. If your plugin requires a server-side Azure plugin for authentication, you should mention this :). And please include where one can find setup instructions, so that users are not lost.

simongottschlag:
> Thanks for clarifying this a bit. For example, I didn't see that this is the same as `azad-kube-proxy`.
> 
> 
> 
> Overall, it more and more looks to me, that `proxy` is not the ideal name here. To me it sounds like this should be `azad-auth`. After what you wrote, I think `azad` is better than `aad`.
> 
> 
> 
> I'm not fully sure I understand your comment about server-side requirements. If your plugin requires a server-side Azure plugin for authentication, you should mention this :). And please include where one can find setup instructions, so that users are not lost.

Hi!

We shouldn't call it azad-auth, since it will confuse people that this may be related to pure Azure AD auth in some way. If we must have auth, then azad-proxy-auth would be the way to go.

Regarding server side:
The project I'm working on (azad-kube-proxy) makes it possible to provide Azure AD authentication to all Kubernetes clusters (AKS, GKE, EKS, on-prem etc). To make it easy for the end users (not the cluster admins) to use this proxy when it's up and running, I've created this plugin to take care of kubeconfig generation and token retrieval/cache/rotation.

With that said, you can't (at least not my plan) use this plugin without first having the proxy (server side) up and running.

I want to publish the plugin to Krew to make it easier for the cluster admins to tell the end users how to authenticate to their clusters.

This means that people wouldn't be looking into the krew-index and then try to use it without having the proxy first. It's more of an easy way to distribute the plugin and help with lifecycle management / updates for all those end users. Right now there are no end users, only me. 😊

I'm fully open to changing this to what you want - I just want you to understand why I've chosen the name before changing it.

Thanks!

plugins/azad-proxy.yaml
simongottschlag: @corneliusweig ah, good to know! I'm compiling it like this and hopefully the only way people are going to use it is through Krew :)

plugins/azad-proxy.yaml
simongottschlag: Done!

plugins/azad-proxy.yaml
simongottschlag: Done!

plugins/azad-proxy.yaml
simongottschlag: Caveat added! :+1:

plugins/azad-proxy.yaml
simongottschlag: I wrote a more detailed explination here regarding this: https://github.com/kubernetes-sigs/krew-index/pull/997#issuecomment-766459068

Do you still want me to change it? I would rather not since I do think this is the name it should be to match the project, but I will of course follow your recommendations if you still think it's neccesary.

Thanks!

simongottschlag:
Hi @corneliusweig!

Do I need to change the name? Anything else missing or needs changing?

Thanks!

ahmetb:
/lgtm
/approve
Thanks!

ahmetb:
/lgtm
/approve
Thanks!

simongottschlag:
> 
> 
> Also, please consider setting up [Krew release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) which helps you skip manually making updates to your Krew manifests on each new version and send a pull request. It’s a GitHub workflow bot that can run every time you push a new tag. Those PRs are automatically approved.

@ahmetb thank you! 😁🎉 The automation is already prepared and I'll try it with a release tomorrow!

PR #1000, opened 2021-01-27
Add artifacthub-repo.yml for ownership claim

### LINKED ISSUE #972: Krew index in Artifact Hub
Hi 👋 

We've just added support for Krew index repositories to Artifact Hub. This will allow publishers listing the kubectl plugins available in their repositories in Artifact Hub.

The default Krew index repository [has been added](https://artifacthub.io/packages/search?page=1&repo=krew-index) as an example. At the moment the publisher is the Artifact Hub organization, but it would be great if you could claim its ownership, so that the legit owner appears as the publisher. It's an easy and fully automated process documented [here](https://artifacthub.io/docs/topics/repositories/#ownership-claim). There is already a [Kubernetes organization](https://artifacthub.io/packages/search?page=1&org=kubernetes) registered in Artifact Hub that is publishing some packages, like the [nginx ingress controller](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx). Maybe you'd like to use that one if possible, whatever you think suits better.

Artifact Hub scans repositories periodically so any new plugins added or updated in the `krew-index` repository will be listed automatically. We've also added support for some AH [specific annotations](https://artifacthub.io/docs/topics/annotations/krew/) that can be used to improve users' experience in the UI.

Thank you very much!

chriskim06:
I got added to the kubernetes-sigs AH org. Once this file is added I'll make the ownership claim there under that org. After that I'll submit a follow-up PR adding the AH repositoryID to this file.

Fixes #972 

/assign @ahmetb 
/assign @corneliusweig 
/cc @tegioz

### DIFF
--- a/artifacthub-repo.yml
+++ b/artifacthub-repo.yml
@@ -1,0 +1,3 @@
+owners:
+  - name: chriskim06
+    email: chriskim06@gmail.com


ahmetb:
/hold
I have concerns:
1. This YAML gets distributed to all krew installations. It'll have a personal email address in it. I don't think this is good.
2. Depending on how our YAML scanning is impplemented, if this got picked up accidentally, it might brick existing Krew installations etc. I'm a bit uneasy about adding extra files to this repo (especially `.yaml`), although `.github/workflows` directory already has some.

Could there be another way of claiming ownership?

tegioz:
Yes, there is another way. You can invite me (tegioz) temporarily to the `kubernetes-sigs` AH organization so that I can transfer the repo to it. Once it's done I'll leave the organization.

You can also setup an account using a non-personal email. The owners section, including the emails, can be deleted once the ownership claim process has completed. But it'd still be committed.

If you leave the `artifacthub-repo.yaml` file in the repo with the `repositoryID` field set, all the packages available on it will display a [`Verified publisher`](https://artifacthub.io/docs/topics/repositories/#verified-publisher) label.

Screen Shot 2021-01-27 at 20 59 20

ahmetb:
> it will display a [`Verified publisher`](https://artifacthub.io/docs/topics/repositories/#verified-publisher) label.

So a problem with Krew-index regarding this is that these plugins we distribute are not vetted for security and functionality (with the exception of "krew" plugin itself).

So we should not have a "Verified Publisher" icon next to kubectl plugins coming from Krew, as it would be misleading to the users. Have we discussed this?

tegioz:
In Artifact Hub, the `Verified Publisher` label indicates that the entity publishing the repository owns or has control over it.

This definition is displayed in the UI in several places:

Screen Shot 2021-01-28 at 09 13 39

We added support for it because anyone can add any supported repository publicly available, like I did with the `Krew index` one from the Artifact Hub organization. This way users can quickly know when this is happening, potentially helping them decide between multiple options, when used in combination with the source repository url displayed.

But said that, you can leave it out if you think it isn't a good fit. IMHO, once you claim it, the entity publishing it will be the one owning the repository, so it'll be a perfect match for what this label represents.

Somehow this is similar to the [verified creator badge](https://docs.github.com/en/developers/github-marketplace/about-verified-creator-badges) in Github.

corneliusweig:
IMHO, a label with "verified" or "official" is exactly what we don't want. As already said above, krew plugins undergo a minimal vetting on the first submission for general fit as a plugin, but no security vetting. Plugins are untrusted community contributions, so making them look official would be bad.

tegioz:
You can leave it out then if it isn't a good fit, it's not a problem 👍

PR #1003, opened 2021-01-28
Add new plugin kubectl-mc, a multi cluster kubectl client

jonnylangefeld:
# Multi Cluster `kubectl`
A new plugin to send kubectl commands to multiple contexts in your kubeconfig at once.
More info here: https://github.com/jonnylangefeld/kubectl-mc

Example usage:
```bash
$ kubectl mc -r kind -- get pods -n kube-system

kind-kind
---------
NAME                                         READY   STATUS    RESTARTS   AGE
coredns-f9fd979d6-q7gnm                      1/1     Running   0          99m
coredns-f9fd979d6-zd4jn                      1/1     Running   0          99m
etcd-kind-control-plane                      1/1     Running   0          99m
kindnet-8qd8p                                1/1     Running   0          99m
kube-apiserver-kind-control-plane            1/1     Running   0          99m
kube-controller-manager-kind-control-plane   1/1     Running   0          99m
kube-proxy-nb55k                             1/1     Running   0          99m
kube-scheduler-kind-control-plane            1/1     Running   0          99m

kind-another-kind-cluster
-------------------------
NAME                                                         READY   STATUS    RESTARTS   AGE
coredns-f9fd979d6-l2xdb                                      1/1     Running   0          91s
coredns-f9fd979d6-m99fx                                      1/1     Running   0          91s
etcd-another-kind-cluster-control-plane                      1/1     Running   0          92s
kindnet-jlrqg                                                1/1     Running   0          91s
kube-apiserver-another-kind-cluster-control-plane            1/1     Running   0          92s
kube-controller-manager-another-kind-cluster-control-plane   1/1     Running   0          92s
kube-proxy-kq2tr                                             1/1     Running   0          91s
kube-scheduler-another-kind-cluster-control-plane            1/1     Running   0          92s
```

### DIFF
--- a/plugins/mc.yaml
+++ b/plugins/mc.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: mc
 spec:
-  version: v0.4.0
+  version: v0.5.0
   homepage: https://github.com/jonnylangefeld/kubectl-mc
   shortDescription: Run kubectl commands against multiple clusters at once
   description: |+2
@@ -30,32 +30,35 @@ spec:
     # run a debug container on every kind cluster in the context
     mc --regex kind -- run debug --image=markeijsermans/debug --command -- sleep infinity
 
-    # list all contexts with 'dev' in the name, but not '-ci-' in the name
+    # list all contexts with 'dev' in the name, but not '-test-' in the name
     mc -r dev -l -n '-test-'
 
     # list all pods with label 'app.kubernetes.io/name=audit' in the 'default' namespace from all clusters with 'gke' in the name, but not 'dev'
     # run max 5 processes in parallel and enable debug output
     mc -r gke -n 'dev' -p 5 -d -- get pods -n gatekeeper-system -l app.kubernetes.io/name=audit
+
+    # print the context and the pod names in kube-system using jq
+    mc -r kind -o json -- get pods -n kube-system | jq 'keys[] as $k | "\($k) \(.[$k] | .items[].metadata.name)"'
   platforms:
   - selector:
       matchLabels:
         os: darwin
         arch: amd64
-    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.4.0/kubectl-mc_v0.4.0_darwin_amd64.tar.gz
-    sha256: 541f8036a60ae620e5e1c0bfb47937d546199c215ad5bfabece3dbc22fc75b52
+    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.5.0/kubectl-mc_v0.5.0_darwin_amd64.tar.gz
+    sha256: b059a56b9db41bbeaf5c40d83c2f73384b085ae94ec747c97f4d01dca4ce0f52
     bin: kubectl-mc
   - selector:
       matchLabels:
         os: linux
         arch: amd64
-    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.4.0/kubectl-mc_v0.4.0_linux_amd64.tar.gz
-    sha256: a207e37458136d6e1dccea98b40875204c9dcbc7e9f77e3592ff65f11769a2c2
+    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.5.0/kubectl-mc_v0.5.0_linux_amd64.tar.gz
+    sha256: 3fa3f94fd38cc6d112fe121e706e870ac39364ccf2d68bf1c8b79693f8012e12
     bin: kubectl-mc
   - selector:
       matchLabels:
         os: windows
         arch: amd64
-    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.4.0/kubectl-mc_v0.4.0_windows_amd64.tar.gz
-    sha256: 36b0b2287fbf767398ce12d19f1ef776614cdbd62b7094690954b99c1904079d
+    uri: https://github.com/jonnylangefeld/kubectl-mc/releases/download/v0.5.0/kubectl-mc_v0.5.0_windows_amd64.tar.gz
+    sha256: 4b4ae8aa68faf4268cc478869932492eebf56f9f619ace5feb88431c33ed1ef6
     bin: kubectl-mc.exe
 
--- a/plugins/mc.yaml
+++ b/plugins/mc.yaml
@@ -1,0 +1,61 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: mc
+spec:
+  version: v0.4.0
+  homepage: https://github.com/jonnylangefeld/kubectl-mc
+  shortDescription: Run kubectl commands against multiple clusters at once
+  description: |+2
+    A Multi Cluster kubectl Client
+
+    If you work at a company or organization that maintains multiple Kubernetes
+    clusters it is fairly common to connect to multiple different kubernetes
+    clusters throughout your day. And sometimes you want to execute a command
+    against multiple clusters at once. For instance to get the status of a
+    deployment across all `staging` clusters.
+
+    `kubectl-mc` (short for multi cluster) supports this workflow and significantly
+    reduces the return time by executing the necessary API requests in parallel go
+    routines.
+
+    Examples:
+
+    # list all kind contexts
+    mc -r kind -l
+
+    # list the pods in the kube-system namespace of all dev clusters
+    mc -r dev -- get pods -n kube-system
+
+    # run a debug container on every kind cluster in the context
+    mc --regex kind -- run debug --image=markeijsermans/debug --command -- sleep infinity
+
+    # list all contexts with 'dev' in the name, but not '-ci-' in the name
+    mc -r dev -l -n '-test-'
+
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for your submission. 
Did you import Auth plugins as listed in the Best Practices doc?

jonnylangefeld:
Hi @ahmetb, I assume you are talking about [this](https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/#auth-plugins)?
No, I have not. However, the addon doesn't use `client-go`.
Since the addon is supposed to support any `kubectl` command to send against multiple contexts at once, it just calls the `kubectl` binary via 
```go
exec.Command("kubectl", args...)
```
[source](https://github.com/jonnylangefeld/kubectl-mc/blob/02f6a75b1e3ce9da746fa034760f122934296b0f/main.go#L188).

That means the addon doesn't communicate with the API server directly. `kubectl` itself handles all the authentication.

corneliusweig:
Thanks for submitting your plugin. It looks like there already is a plugin with nearly identical feature set: https://github.com/onatm/kubectl-allctx

I think that implementing this via Go has some technical advantages (speed, testing, extensibility). However, bash plugins like `allctx` also provide greater trust (because you can see what will be executed).

Overall, there seems to be no compelling reason to add `mc`, given that there is already `allctx`. Krew index already has many useful plugins, so it's getting more and more common that we have to say no due to significant overlap with existing plugins. You can still distribute your plugin via `krew` though, because we now support custom indexes ([1](https://krew.sigs.k8s.io/docs/user-guide/custom-indexes/), [2](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/)). There is also https://github.com/ishantanu/awesome-kubectl-plugins which is a popular alternative index.

If you think I overlooked something, please let me know.

/close

jonnylangefeld:
Thanks for the review @corneliusweig! I respect your decision, however I do want to make a few more points for why I think `mc` adds unmatched features over `allctx`.

At work I often have the use case to run `kubectl` on certain clusters. For instance on all staging clusters. Or on all gke clusters, but not on the on-prem clusters.
That's why I added the feature of selecting contexts based on regex to `mc`. You can also exclude contexts based on regex (for instance run a command on all dev clusters, but not the ones including the word 'test' ([example](https://github.com/jonnylangefeld/kubectl-mc/blob/master/main.go#L53-L54)).
As the name suggests, with `allctx` you only have the possibility to run commands across all contexts ([src](https://github.com/onatm/kubectl-allctx/blob/main/kubectl-allctx#L36)).

In addition to that, my main reason for implementing `mc` in go rather than bash was the ability to execute all `kubectl` commands in parallel via go routines ([src](https://github.com/jonnylangefeld/kubectl-mc/blob/master/main.go#L122)). So every command is executed in parallel through its own process (There is an optional flag to limit the maximum amount of parallel go routines as well). `allctx` only successively executes the `kubectl` sub-commands one by one ([src](https://github.com/onatm/kubectl-allctx/blob/main/kubectl-allctx#L81-L91)).

Last but not least, I'm also planning to add a features that parses the output for cases where `-o yaml` or `-o json` is given (not every `kubectl` command allows for yaml or json output). That would allow us to get for instance a single json object as output across multiple clusters, that then can be piped into `jq` or used in any other way. (I'm not done with that feature yet, but will add it soon).

Regarding the trustworthiness of a bash script over a go binary, I would say since it's both open source, it's the same level of trust. And don't we already trust go binaries like `kubectl` and `krew`? 😊 

Given these points I'd ask for you to reconsider, but of course it's up to you!

ahmetb:
Yeah I think it's worth reconsidering if it is notably different. 
/reopen

jonnylangefeld:
I added the feature I was talking about in my last post: You can now parse outputs with `-o yaml` or `-o json` across contexts for further investigation with `jq` or `yq` or just to store the output somewhere.
Hope now the feature set is rich enough to add it despite the existing `allctx` 😊

plugins/mc.yaml
```diff
@@ -0,0 +1,64 @@
...
+  homepage: https://github.com/jonnylangefeld/kubectl-mc
+  shortDescription: Run kubectl commands against multiple clusters at once
+  description: |+2
+    A Multi Cluster kubectl Client
+
+    If you work at a company or organization that maintains multiple Kubernetes
+    clusters it is fairly common to connect to multiple different kubernetes
+    clusters throughout your day. And sometimes you want to execute a command
+    against multiple clusters at once. For instance to get the status of a
+    deployment across all `staging` clusters.
```
corneliusweig, line 16: More users will read this, if the text is short. A short description takes longer to write, but it is worth it!

How about:

```suggestion
    This plugin helps to work with multiple kubernetes clusters at once. For instance
    to get the status of a deployment across all `staging` clusters.
```

plugins/mc.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+    # list all contexts with 'dev' in the name, but not '-test-' in the name
+    mc -r dev -l -n '-test-'
+
+    # list all pods with label 'app.kubernetes.io/name=audit' in the 'default' namespace from all clusters with 'gke' in the name, but not 'dev'
+    # run max 5 processes in parallel and enable debug output
+    mc -r gke -n 'dev' -p 5 -d -- get pods -n gatekeeper-system -l app.kubernetes.io/name=audit
+
+    # print the context and the pod names in kube-system using jq
+    mc -r kind -o json -- get pods -n kube-system | jq 'keys[] as $k | "\($k) \(.[$k] | .items[].metadata.name)"'
```
corneliusweig, line 41: This should go into your `kubectl mc --help` output. Try to work with a single example, and give advice where to find more help.

plugins/mc.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: mc
+spec:
+  version: v0.5.0
+  homepage: https://github.com/jonnylangefeld/kubectl-mc
+  shortDescription: Run kubectl commands against multiple clusters at once
+  description: |+2
+    A Multi Cluster kubectl Client
```
corneliusweig, line 10: A heading should not be necessary. `krew` already takes care of this in its output.

corneliusweig (commented):
Sorry for :snail:-ing so hard!

I'm not completely opposed to having this plugin, because it is useful. However, I think that the overlap with `allctx` is very large. Did you know that `allctx` can also select contexts by pattern matching (the plugin name might not be optimal though).

Let's hear one more opinion on this. I already left a few comments how to improve the manifest.

/assign @chriskim06

plugins/mc.yaml
jonnylangefeld: 👍🏼 
They are already part of `kubectl mc --help`, so I just removed them from here.

plugins/mc.yaml
jonnylangefeld: removed 👍🏼

jonnylangefeld:
Regarding pattern matching in `allctx`: [In the code](https://github.com/onatm/kubectl-allctx/blob/main/kubectl-allctx#L81) I only see how it iterates over all contexts in the kubeconfig, unfiltered.
It only [takes a `--like` flag](https://github.com/onatm/kubectl-allctx/blob/main/kubectl-allctx#L61-L70) as input, but that's used to filter the [output](https://github.com/onatm/kubectl-allctx/blob/main/kubectl-allctx#L87), not the contexts (like give me all pods with 'grafana' in their name from all contexts). A resource selection like that is probably better done through labels.
Happy to get another opinion, I implemented your feedback.

corneliusweig (approved):
Hey @jonnylangefeld, I'm sorry, you are right. `allctx` indeed only operates on only all contexts. So `mc` indeed solves a different use case. Thanks for being insistent here!

I left a final nit about the manifest, otherwise this looks good!

chriskim06:
I agree this seems to add useful functionality on top of what `allctx` does. After Cornelius' last comment is addressed I think this will be good to go.

plugins/mc.yaml
```diff
@@ -0,0 +1,37 @@
...
+  version: v0.5.0
+  homepage: https://github.com/jonnylangefeld/kubectl-mc
+  shortDescription: Run kubectl commands against multiple clusters at once
+  description: |+2
+    This plugin helps to work with multiple kubernetes clusters at once, filtered by
+    a regular expression. For instance to get the status of a deployment across all
+    `staging` clusters run the following command:
+    mc -r staging -- get deployments -n kube-system
+
+    For more examples run kubectl mc --help
```
ahmetb, line 15: this line might also be redundant (I assume `kubectl mc` already shows help).

we also tell users how to use this plugin after installation:

```
Installed plugin: tree
\
 | Use this plugin:
 | 	kubectl tree
 | Documentation:
 | 	https://github.com/ahmetb/kubectl-tree
[...]
```

jonnylangefeld: you're right it does! I'll remove this line

jonnylangefeld:
Alright, thank you for your guys' feedback, I addressed it in the last commit.
Seems like this is ready to be merged then?

ahmetb:
/lgtm
/approve

Thanks!

jonnylangefeld:
Already done 😉 
https://github.com/jonnylangefeld/kubectl-mc/blob/master/.github/workflows/release.yml#L28-L29
Will see if it works on my next release 😄 
Thank you for merging!

PR #1021, opened 2021-02-09
added plugin pod-inspect

jpriebe:


ahmetb:
/hold
You have manually removed the pull request template which asks you to read the naming guidelines.

jpriebe:
@ahmetb - I'm not sure what I did wrong here.  Yes, I removed the template, but that appeared to be a comment that doesn't need to stay in the request once it's filed.  I *did* read the naming guidelines.  Did I violate the guidelines?  I've already had to rename my plugin twice to try to get something that works with kubectl and that (I thought) met the guidelines.

ahmetb:
> Punctuation
> Plugin names must be all lowercase and separate words with hyphens. Don’t use camelCase, PascalCase, or snake_case; use kebab-case.

https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/

jpriebe:
When I first read the naming guidelines, I was still planning to use "pod-inspect".  I then tried "pod-inspect", and it didn't work (I put kubectl-pod-inspect in my path, but kubectl wouldn't recognize it), so I renamed it "pod_inspect", forgetting item #1 in the naming guidelines.  My apologies.

I understand that krew does some symlink magic, but if I use kebab-case, then my plugin won't work outside of a krew context (without asking my users to do some extra work to rename the executable or symlink it or whatever).

Do you have any guidance so that a developer can build a pluging that easily works both inside and outside of krew?

chriskim06:
You can leave the binary name snake cased, it's just the plugin name should be kebab cased. There's a few plugins like this, [cert-manager](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/cert-manager.yaml#L21) is one that I can immediately think of.

jpriebe:
@chriskim06 - great idea.  I'm refactoring now.  One thing I noticed: the go snippet to detect whether running as a plugin on this page: https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/ doesn't seem to actually do anything different when the command is run directly vs as a plugin.

I'd really like to detect that so that I can output appropriate usage for the person who happens to run my plugin directly.  I find for kubectl plugin beginners, this dichotomy is kind of confusing.  Especially with the "-"/"_" naming issues.

If you've got any other techniques for accurately detecting `kubectl-pod_inspect` vs `kubectl pod-inspect` invocations, I'd love to see them.  I'll spend a little time experimenting myself.  Maybe I can detect that my process was `execve`'d by a process called with a different command line?

chriskim06:
hmm I'm not sure, afaik none of our plugins do this currently. I would think most people would invoke your plugin via kubectl, so it wouldn't be an issue.

ahmetb:
If you install your plugin with Krew, that snippet will contain the name of the symlink (kubectl-foo_bar) we create for your plugin in os.Args[0].

So if you distribute your binary without kubectl- prefix, you can easily tell the difference. Does that help?

jpriebe:
Updated pull request with renamed plugin.  I found a couple of thing:

- there doesn't seem to be a way to reliably distinguish between a user calling your plugin directly or calling it via kubectl; because of this, I opted to craft my usage text to reflect the common case of calling it through kubectl
- cobra *really* does not like putting "kubectl " in front of your Use: string.  It will create some highly inaccurate usage strings from that, and you have to jump through a lot of hoops to override the template.  Not sure if there's a better approach.

jpriebe:
@ahmetb - that's an option I didn't think about - I was fixated on calling my binary "kubectl-SOMETHING".  I don't think it's terrible the way I've structured it, but maybe at some point in the future, I'll change the binary name so I can tell the difference.  I've spent WAY too much time refactoring this thing, and I'm not inclined to change anything needlessly at this point.

plugins/pod-inspect.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pod-inspect
+spec:
+  version: v0.1.7
+  homepage: https://github.com/jpriebe/kubectl-pod-inspect
+  shortDescription: Get all of a pod's details at a glance
+  description: |
+    Provides detailed information about a pod, including its containers' statuses, pod events, and logs from non-ready containers.
```
ahmetb, line 10: please wrap at 80 chars, looks better on user terminals that way.

jpriebe: fixed.

ahmetb:
/lgtm
/approve
/hold
cc: @corneliusweig any objections?

corneliusweig:
This looks like a fine plugin! Thanks for building and sharing! :tada: 
/hold cancel

PR #1022, opened 2021-02-10
Add Reliably plugin

mperrien:
This plugin makes the [Reliably CLI](https://github.com/reliablyhq/cli) directly accessible from kubectl.

The Reliably CLI is a tool that scans Kubernetes manifests for potential reliability issues.

Signed-off-by: Marc Perrien <marc@chaosiq.io>

### DIFF
--- a/plugins/reliably.yaml
+++ b/plugins/reliably.yaml
@@ -1,28 +1,37 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
-  name: "reliably"
+  name: reliably
 spec:
-  version: v0.0.2
-  homepage: https://github.com/reliablyhd/kubectl-reliably
+  version: v0.3.0
+  platforms:
+  - uri: https://github.com/reliablyhq/cli/releases/download/v0.3.0/reliably_v0.3.0_darwin_amd64.tar.gz
+    sha256: b704ac7c018a4446d397acc05cc1d25caee14584b27c76465f39a60cd5529d09
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    bin: reliably
+    files:
+    - from: "reliably"
+      to: "."
+    - from: LICENSE
+      to: .
+  - uri: https://github.com/reliablyhq/cli/releases/download/v0.3.0/reliably_v0.3.0_linux_amd64.tar.gz
+    sha256: 3f43d20d52215220ec5d439f09e119e544fb678ef75a01a3d441bef685790d53
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    bin: reliably
+    files:
+    - from: "reliably"
+      to: "."
+    - from: LICENSE
+      to: .
+  homepage: https://github.com/reliablyhq/cli
   shortDescription: "Surfaces reliability issues in Kubernetes."
   description: |
     Surfaces reliability issues in your Kubernetes configuration,
     using the Reliably CLI.
-  platforms:
-  - selector:
-      matchExpressions:
-      - key: "os"
-        operator: "In"
-        values:
-        - darwin
-        - linux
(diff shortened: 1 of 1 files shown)


mperrien:
CLA is signed

ahmetb:
Thanks for the submission. 
This plugin seems to be a light shim around "reliably" CLI. This can be ok and we have plugins like that. 

However, if your cli is a single binary written in Go, you can actually turn it into a multi-purpose binary. It can act like a kubectl plugin if it is called as such. Have you considered this? 

What benefit do you give your users as a kubectl plugin here if you already have a standalone cli tool?

mperrien:
Hello Ahmet, thanks for your response.

While we could probably add a reliably subcommand to enable the CLI as a kubectl plugin, the goal here was to make it discoverable to people who only work through kubectl plugins to get access to our features. This is why we thought of relying on the Krew package manager.

Does it make sense?

ahmetb:
@mperrien I don't think I conveyed my question clearly. Here's another attempt.

It seems the actual value is in  `reliably` CLI. But this plugin does not ship `reliably` CLI. It ships a bash script, that installs `reliably` CLI [1], then the next time, it just launches reliably CLI [2].

As it can be seen in [2] your script is literally just an alias to reliably CLI. And you're reinventing the "installation" step which Krew is designed for yourself in a bash script.

So my question is, why don’t you just ship the actual `reliably` CLI binary as a kubectl like many other plugins do? See `cert-manager` or `tail` plugins as examples of some plugins that distribute themselves as both plugins as well as standalone tools.

[1] https://github.com/reliablyhq/kubectl-reliably/blob/bde140db8abaf0f9554812bc0a3cb32a36e30f0a/reliably.sh#L32-L120
[2] https://github.com/reliablyhq/kubectl-reliably/blob/bde140db8abaf0f9554812bc0a3cb32a36e30f0a/reliably.sh#L121

mperrien:
@ahmetb I think I finally understood...

I pushed an updated manifest which ships the Go binary as the kubectl plugin.

Thank you!

plugins/reliably.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - from: "reliably"
+      to: "."
+    - from: LICENSE
+      to: .
+  homepage: https://github.com/reliablyhq/cli
+  shortDescription: "Surfaces reliability issues in Kubernetes."
+  description: |
+    Surfaces reliability issues in your Kubernetes configuration,
+    using the Reliably CLI.
+
```
ahmetb, line 37: nit: if you can remove this last empty line (has spaces though), it would be easier for you to automate releases in the future (as they would get auto-approved). https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/

plugins/reliably.yaml
```diff
@@ -0,0 +1,37 @@
...
+        os: linux
+        arch: amd64
+    bin: reliably
+    files:
+    - from: "reliably"
+      to: "."
+    - from: LICENSE
+      to: .
+  homepage: https://github.com/reliablyhq/cli
+  shortDescription: "Surfaces reliability issues in Kubernetes."
```
ahmetb, line 33: no trailing "."

plugins/reliably.yaml
```diff
@@ -0,0 +1,37 @@
...
+      matchLabels:
+        os: linux
+        arch: amd64
+    bin: reliably
+    files:
+    - from: "reliably"
+      to: "."
+    - from: LICENSE
+      to: .
+  homepage: https://github.com/reliablyhq/cli
```
ahmetb, line 32: I think if you link to https://reliably.com/docs/ it can be better, as there's no documentation/examples in your repo to get started.

We'd have to map your name to repo elsewhere in code for https://krew.sigs.k8s.io/plugins to be able to link to your repo, but that's ok and we already do it for other plugins.

ahmetb (changes requested):

mperrien:
Thank you Ahmet, I made the requested changes

PR #1079, opened 2021-03-02
add plugin colour

meltingrobot:

### DIFF
--- a/plugins/colour.yaml
+++ b/plugins/colour.yaml
@@ -1,0 +1,22 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: colour
+spec:
+  version: "v1.0.0"
+  platforms:
+  - uri: https://github.com/meltingrobot/kubernetes-colour/archive/1.0.0.tar.gz
+    sha256: 57260145394924096b2050579a495db3c1b0a94b7aa86c2e45df1de7f5ec9ec7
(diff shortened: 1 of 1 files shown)


ahmetb:
/kind gray-area

I've jokingly demoed this idea on stage several times.
https://youtu.be/PvWr_Au-934?t=1021
https://youtu.be/83ITOTsXsHU?t=515

I don’t think the functionality is enough to warrant a place in krew-index, that said you can use [custom indexes](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/) to distribute your plugin.

I'll let other maintainers chime in.
/hold

ahmetb:
@corneliusweig thoughts? planning to close.

corneliusweig:
I agree that colouring output does not add value. On the contrary, it even breaks other plugins, which _do_ use colors.

On the other side, I think there are playful people working with kubernetes that will like colourful output. So I'd encourage you to publish your plugin in a custom registry ([1](https://krew.sigs.k8s.io/docs/user-guide/custom-indexes/), [2](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/)). For example, https://github.com/ishantanu/awesome-kubectl-plugins#use-awesome-kubectl-plugins-as-a-custom-index-via-krew is such an alternative registry.

Thanks for sending your plugin for review though. Don't hesitate to send your next idea!

/close

PR #1088, opened 2021-03-05
Adds kubectl-ccsecret plugin file

rewanth1997:
Requesting to add `ccsecret` plugin to krew index.

Link to source code: https://github.com/rewanth1997/kubectl-ccsecret

```
wget https://raw.githubusercontent.com/rewanth1997/kubectl-ccsecret/master/deploy/krew/plugin.yaml
kubectl krew install --manifest=./plugin.yaml
```

Repo link: https://github.com/rewanth1997/kubectl-ccsecret

Please review the code. I'm happy to make the necessary changes if any.

Regards,
Rewanth


rewanth1997:
@corneliusweig @ahmetb Feel free to suggest a better name to improve the discoverability of this plugin.

rewanth1997:
@luksa Similar to last time (https://github.com/kubernetes-sigs/krew-index/pull/254), I developed one more plugin. Please review it and provide your valuable inputs.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+    files:
+    - from: "*"
+      to: "."
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: v1.0.0
+  homepage: https://github.com/rewanth1997/kubectl-ccsecret
+  caveats: |
```
corneliusweig, line 39: Please reserve `caveats` for things like dependencies or prerequisites.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+    Usage:
+      kubectl ccsecret <COMMAND> <FLAGS>
```
corneliusweig, line 41: This is shown by default by `krew info`

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    Documentation:
+      https://github.com/rewanth1997/kubectl-ccsecret
```
corneliusweig, line 44: There's already the `homepage` field.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+    Examples:
+
+    Create generic secret with --from-literal:
+    $ kubectl ccsecret generic my-secret --from-literal key1 --from-literal key2
+    Enter value for key1: ******
+    Enter value for key2: ******
+
+    Create docker-registry secret with --docker-password:
+    $ kubectl ccsecret docker-password my-secret --docker-password
+    Enter value for docker password: ******
```
corneliusweig, line 62: One example should be enough. I don't think it adds much to have two almost identical ones.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    Documentation:
+      https://github.com/rewanth1997/kubectl-ccsecret
+  shortDescription: Create kubectl secrets by taking sensitive input from console
+  description: |
+    kubectl-ccsecret allows users to create kubectl secrets by taking sensitive
+    input from the console to prevent information leakages via terminal history,
+    shoulder surfing, etc
+    
+    More on https://github.com/rewanth1997/kubectl-ccsecret
```
corneliusweig, line 51: The `homepage` field is shown when calling `krew info`. No need to repeat yourself.

corneliusweig (commented):
I think this plugin is original and useful.

I would usually argue that creating resources from CLI is bad practice, but secrets may be an exception. Let's also hear another opinion from @chriskim06.

Regarding naming: `ccsecret` is fine regarding our naming rules. However, I find it not very intuitive or memorable. WDYT about something like `blind-secret`?

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+  name: ccsecret
+spec:
+  platforms:
+  - uri: https://github.com/rewanth1997/kubectl-ccsecret/releases/download/v1.0.0/kubectl-ccsecret-Darwin-x86_64.tar.gz
+    sha256: e22ab5c443dd1ae00fdc8e1a68f142edbde642bbc1c84862569a9c510b81b373
+    bin: kubectl-ccsecret
+    files:
+    - from: "*"
+      to: "."
```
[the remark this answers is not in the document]
rewanth1997: Done

plugins/ccsecret.yaml
rewanth1997: Noted and fixed

plugins/ccsecret.yaml
rewanth1997: Done

plugins/ccsecret.yaml
rewanth1997: Removed it.

plugins/ccsecret.yaml
rewanth1997: Sure, makes sense.

plugins/ccsecret.yaml
rewanth1997: Yeah, done.

rewanth1997:
@corneliusweig Thanks for the quick review and I'm glad you liked the idea. Most of the issues you pointed in the manifest file are related to redundancies and I fixed them all.

I spent more time brainstorming for an apt name than I spent building this plugin but couldn't come up with any handy name.

`blind-secret` is a good idea but since we are creating secrets, I think we should have `create` keyword in the plugin name. I thought of `create-hidden-secret`, `create-console-secret` but these are too lengthy to type, so I changed the name to `ccsecret`, a short form of `create console secret`. I think maybe we can make it `create-blind-secret` or `cbsecret` or `create-bsecret`. I don't know, this is very confusing.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,41 @@
...
+        arch: amd64
+  version: v1.0.0
+  homepage: https://github.com/rewanth1997/kubectl-ccsecret
+  shortDescription: Create kubectl secrets by taking sensitive input from console
+  description: |
+    kubectl-ccsecret allows users to create kubectl secrets by taking sensitive
+    input from the console to prevent information leakages via terminal history,
+    shoulder surfing, etc
+
+    Examples:
```
corneliusweig, line 36: nit
```suggestion
    Example:
```

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+    Create generic secret with --from-literal:
+    $ kubectl ccsecret generic my-secret --from-literal key1 --from-literal key2
+    Enter value for key1: ******
+    Enter value for key2: ******
```
corneliusweig, line 41:
```suggestion
    $ kubectl ccsecret generic my-secret --from-literal key
    Enter value for key: ******
```
I think the second item does not add much. But you can also leave it there.

plugins/ccsecret.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - uri: https://github.com/rewanth1997/kubectl-ccsecret/releases/download/v1.0.0/kubectl-ccsecret-Windows-x86_64.zip
+    sha256: 5d5469b792354010b4c9260d181a3ef8446b744f16481bcf447b4c81f7929865
+    bin: kubectl-ccsecret.exe
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  version: v1.0.0
+  homepage: https://github.com/rewanth1997/kubectl-ccsecret
+  shortDescription: Create kubectl secrets by taking sensitive input from console
```
corneliusweig, line 30: No need to mention `kubectl`, this is redundant. Maybe keep the description a bit more abstract, like so:
```suggestion
  shortDescription: Create secrets with improved privacy
```

corneliusweig (commented):
The problem with `ccsecret` is that the name does not transport the idea behind the plugin. So I'd recommend to spend more time on a good plugin name. You could also be more poetic and call it `whisper-secret` or so.

Btw, I like a lot that you simply wrap `kubectl` with your plugin. I wish more plugins were designed in that way.

---

Idea for the next version: add an option that requires to type secrets twice, so that you can spot typos.

plugins/ccsecret.yaml
rewanth1997: Fixed.

plugins/ccsecret.yaml
rewanth1997: True. Done.

plugins/ccsecret.yaml
rewanth1997: Yeah, this sounds more rhetorical. Done.

rewanth1997:
@corneliusweig I agree, even I didn't like the `ccsecret` name. `whisper-secret` sounds promising and aptly conveys the idea of the plugin IMO. I really liked it 😄 Maybe we can wait for one more day to see if others have any recommendations. @ahmetb @luksa

Thanks for the compliment on the implementation, I'm glad you had a glance at the code. The modular coding style and structure make it easy to integrate other flags (if any)  as well to the list of `whisper-secret`, whenever required.

---
In the next version, we can add an option like `--check/-c` that allows the user to enter the secret value twice to check typos. This will be a good addition.

rewanth1997:
Since we are considering security POV, we will set the default setting to prompt the input box twice. In the next release, we can add a flag like (`--ignore-typo/-i`) that allows the user to accept the typo risk and prompts the input field only once.

ahmetb:
Honestly the name is ok with me, I might've called it `mksecret` or something like that otherwise. Right now the `ccsecret` name indeed doesn’t convey the meaning of the plugin well ("create secrets with secure input prompt"), and therefore it'll be less discoverable, but it is acceptable.

rewanth1997:
Thanks @ahmetb. `mksecret` is also good but it doesn't convey the meaning of the plugin well. I liked @corneliusweig recommendation `whisper-secret` which kind of conveys the meaning of the plugin but it's a bit long. Do you have any other handy names coming to your mind that are short and also might convey the meaning of the plugin? @ahmetb @corneliusweig

ahmetb:
I just tested this plugin and noted it doesn't work on darwin, but you provided binary release. So maybe there's need for more polishing here.

I recommended `mksecret` because it's like `make-secret` but shorter. It might be just me :) You can call it `type-secret` or `input-secret`.

ahmetb:
/hold cancel
/lgtm
/approve

PR #1105, opened 2021-03-12
Adding build to krew-index

mikeroySoft:
This plugin installs [buildkit-cli-for-kubectl](https://github.com/vmware-tanzu/buildkit-cli-for-kubectl) which allows you to build OCI container images directly within a Kubernetes cluster.  It replaces the `docker build` command to let you quickly and easily build your single and multi-architecture container images.

### DIFF
--- a/plugins/buildkit-cli.yaml
+++ b/plugins/buildkit-cli.yaml
@@ -1,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: build
+spec:
+  version: v0.1.2
+  homepage: https://github.com/vmware-tanzu/buildkit-cli-for-kubectl
+  shortDescription: Build OCI and Docker images with your kubernetes cluster
+  description: |
+    BuildKit CLI for kubectl is a tool for building OCI and Docker images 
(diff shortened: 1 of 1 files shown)


ahmetb:
The name `build` is probably going to be an issue. See [naming guide]*https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/#be-specific) "Be specific" point. If we assign the name `build`, we can't ever assign it to another plugin. As maintainers, we need to think about "what if a kaniko CLI comes up and also asks for `build` name?".

If you’d like to keep this name, an option might be to distribute by maintaining [your own plugin index](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/). Otherwise, I recommend you to go with a name like `buildkit`.
/kind gray-area

mikeroySoft:
Thanks @ahmetb, we'll evaluate the options and go from there.

dhiltgen:
> The name `build` is probably going to be an issue

Totally get the concern.

We're aiming for a "drop in" replacement for `docker build` via `kubectl build`, preserving the UX and general developer experience.  It sounds like we should hold off on the krew distribution model for now and revisit in the future.

mikeroySoft:
Thanks folks, the discussion is appreciated.

PR #1132, opened 2021-03-22
Add external-forward plugin

int128:
I'd like to add a new plugin https://github.com/int128/kubectl-external-forward. It is a kubectl plugin to connect to an external host via socat pod on a cluster.

Close https://github.com/kubernetes-sigs/krew-index/pull/1170.


plugins/socat.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: socat
+spec:
+  homepage: https://github.com/int128/kubectl-socat
+  shortDescription: connect to external host via socat pod
+  description: This is a kubectl plugin to connect to an external host via socat Pod on a cluster
```
ahmetb, line 8: Delete until "connect", redundant.

plugins/socat.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: socat
+spec:
+  homepage: https://github.com/int128/kubectl-socat
+  shortDescription: connect to external host via socat pod
```
ahmetb, line 7: Capitalize C

ahmetb (changes requested):

ahmetb:
Do you think  it is worth clarifying how this plugin is different/better than `kubectl port-forward`?

int128:
It would be nice to change the name to like `kubectl external-forward` because this plugin provides the port-forwarding to an external host. `kubectl port-forward` forwards to only internal pod.

corneliusweig:
Thanks for submitting another helpful tool!
From first looks, it seems to be pretty similar to socks5-proxy (https://github.com/yokawasa/kubectl-plugin-socks5-proxy). Can you clarify in what regard it is different? Maybe also update your `description` with this explanation to help our users choose the right alternative.

ahmetb:
@int128 I think I get this but it would be great to enhance your repo (and `description` field) to explicitly state how it's different than `kubectl port-forward`. Even after seeing the diagram in your repo, I'm somewhat left puzzled about its value add.

int128:
Thank you for your review. Your review helped to enhance the explanation in the repository. Finally I revised the description to clarify the value of this plugin as follows:

>    This allows you to connect from your laptop to a host outside of a cluster.
>    It is useful when a host such as database is in a private network and unreachable from your laptop.
>    While `kubectl port-forward` connects to a pod inside a cluster, this connects to a host outside a cluster.
>
>    This plugin runs a TCP proxy pod on a cluster and set up a port-forwarder on your laptop.
>    You just run your application with localhost configuration as if a database is running on local.
>    No proxy configuration such as HTTP_PROXY or SOCKS is needed.

I decided to rename this project because I'm thinking using a proxy other than socat.

corneliusweig:
Sorry for the long silence, this fell through the cracks :pray:

I think I understand the purpose of your plugin now, let me rephrase it: Your plugin sets up a jump-host in the cluster to be able to connect to any valid address within the cluster network.

Although `external-forward` is ok-ish, what about calling your plugin `jumphost`. This would immediately ring a bell for what it can be useful. WDYT?

plugins/external-forward.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: external-forward
+spec:
+  homepage: https://github.com/int128/kubectl-external-forward
+  shortDescription: Connect to external host via cluster
+  description: |
```
corneliusweig, line 8: Please remember to wrap around at ~80 chars, so that it prints nicely on wide and narrow terminals.

corneliusweig:
@int128 gentle ping. Do you intend to take this up? IIRC we could merge almost as-is, but the naming will be final.

int128:
Thank you for your review.
Let me close this request because this is not still stable. 
Actually I have introduced this tool in our organization and it still has some problems.
For example, a garbage pod is remaining after disconnected.
I'd like to try yet another approach such as Kubernetes controller and the name would be different.

PR #1133, opened 2021-03-22
add pod-viewer plugin

HamzaZo:
Adding a new plugin called `pod-viewer`
`pod-viewer` provides you with a full view of Kubernetes pod with information such as conditions, events, logs, etc.


HamzaZo:
I signed it

ahmetb:
/assign @corneliusweig 
IIRC there are quite a few plugins already like this.

HamzaZo:
@ahmetb  Until now I didn't see a plugin that gives the whole output, that's why I started developing the `pod-viewer`. However, if you suggest any enhancement, I'm all ears

ahmetb:
We tend to not accept plugin if there's no sufficient value add. If all you're doing is showing describe+logs, I think it falls into that territory. 

PR #1136, opened 2021-03-23
Add kpexec

ssup2:
Hi. I want to register my plugin called 'kpexec' . Please review this. Thanks.
- https://github.com/ssup2/kpexec

### DIFF
--- a/plugins/kpexec.yaml
+++ b/plugins/kpexec.yaml
@@ -7,20 +7,19 @@ spec:
   shortDescription: Execute process with privileges in a pod
   description: |
     kpexec is a kubernetes cli that runs commands in a container with high
-    privileges. kpexec helps execute pod debugging smoothly by executing commands
-    with high privileges regardless of container privileges. Also, kpexec has a
-    tools mode that adds useful debugging tools in the debugged container. The
-    tools mode is useful when there necessary debugging tools are missing in the
-    target container. kpexec runs a highly privileged container on the same node as
+    privileges. It runs a highly privileged container on the same node as
     the target container and joins into the namespaces of the target container
-    (IPC, UTS, PID, net, mount).
+    (IPC, UTS, PID, net, mount). This is useful for debugging where you often
+    need to execute commands with high privileges. Also, kpexec has a tools
+    mode, which adds useful debugging tools into the debugged container. The
+    tools mode is useful when there necessary debugging tools are missing in the
+    target container.
 
-    kubectl-exec runs the command with the same privileges as the
+    In contrast, kubectl-exec runs the command with the same privileges as the
     container. For example, if a container does not have network privileges,
-    the command executed by kubectl-exec also has no network privileges. Also,
-    kubectl-exec does not provide an option to force command to run with high
-    privileges. This makes debugging the pod difficult. If you use kpexec instead
-    of kubectl-exec, you can easily perform debugging.
+    the command executed by kubectl-exec also has no network privileges.
+    This makes debugging the pod difficult. If you use kpexec instead
+    of kubectl-exec, you can easily get high privileges for debugging.
   caveats: |
     kpexec requires the privileges to run privileged pods with hostPID.
   version: v0.2.7
--- a/plugins/kpexec.yaml
+++ b/plugins/kpexec.yaml
@@ -1,0 +1,62 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kpexec
+spec:
+  homepage: https://github.com/ssup2/kpexec
+  shortDescription: Execute process with privileges in a pod
+  description: |
+    kpexec is a kubernetes cli that runs commands in a container with high
+    privileges. kpexec helps execute pod debugging smoothly by executing commands
+    with high privileges regardless of container privileges. Also, kpexec has a
+    tools mode that adds useful debugging tools in the debugged container. The
+    tools mode is useful when there necessary debugging tools are missing in the
+    target container. kpexec runs a highly privileged container on the same node as
+    the target container and joins into the namespaces of the target container
+    (IPC, UTS, PID, net, mount).
+
+    kubectl-exec runs the command with the same privileges as the
+    container. For example, if a container does not have network privileges,
+    the command executed by kubectl-exec also has no network privileges. Also,
+    kubectl-exec does not provide an option to force command to run with high
+    privileges. This makes debugging the pod difficult. If you use kpexec instead
+    of kubectl-exec, you can easily perform debugging.
+  caveats: |
+    kpexec requires the privileges to run privileged pods with hostPID.
+  version: v0.2.7
+  platforms:
+  - bin: kubectl-pexec
+    uri: https://github.com/ssup2/kpexec/releases/download/v0.2.7/kubectl_pexec_0.2.7_Linux_amd64.tar.gz
+    sha256: b8428d93b0648f15b901cd39472b508a164bc41a91de8a3b1ae162deab9f9449
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  - bin: kubectl-pexec
+    uri: https://github.com/ssup2/kpexec/releases/download/v0.2.7/kubectl_pexec_0.2.7_Linux_arm64.tar.gz
+    sha256: b41c2272672af7d11244d759185fcd63a71d9dd6b2438aa8eeadfc7f081f8d58
+    selector:
+      matchLabels:
+        os: linux
+        arch: arm64
+  - bin: kubectl-pexec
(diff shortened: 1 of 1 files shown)


plugins/kpexec.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kpexec
+spec:
+  homepage: https://github.com/ssup2/kpexec
+  shortDescription: Execute process with privilege in a pod
```
ahmetb, line 7: privileges

plugins/kpexec.yaml
```diff
@@ -0,0 +1,45 @@
...
+  description: Execute process with privilege in a pod
```
ahmetb, line 8: Please go into detail of how it works here (80 character wrap multi line string)

ahmetb:
I quickly noticed from your source code you don't import client-go auth plugins. https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/ssup2/kpexec%24%40master+auth&patternType=literal Please consider doing so.

ssup2:
Thanks, your feedback. Now, kpexec only supports kubeconfig file and I don't have a plan to support client-go auth plugins. Is it necessary to register??

ahmetb:
I'm not entirely sure what "supporting only kubeconfig file" means. If my cluster uses the gcp auth method (configured in kubeconfig file), you won't be able to authenticate any API call you make to my cluster from client-go (unless you're shelling out to "kubectl exec" or "kubectl run")?

ssup2:
kpexec's client-go uses local kubeconfig file (default ~/.kube/config) and kpexec also uses kubectl command.
So to use kpexec, kubectl and kubeconfig must be set properly.

I tested works well with AWS eks and my local cluster.

ahmetb:
Alright I'll test this but this manifest is currently not installing.

ahmetb:
Yep I was right. 

```
$ kpexec --cnsenter-gc
Failed to run cnsenter pod's garbage collector : failed to set clientset : no Auth Provider found for name "gcp"
```

Please familiarize yourself with https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/#auth-plugins.
/hold

ssup2:
Oops.. Thanks. I will fix the manifest and auth plugin. Thanks :)

ssup2:
@ahmetb I fixed manifest and apply auth plugin to kpexec. Thanks :)
commit - https://github.com/ssup2/kpexec/commit/fe37691aefbce066df23c43b8697af39d3b2ef39

ssup2:
@corneliusweig Hello. Are there any feedbacks?

plugins/kpexec.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  homepage: https://github.com/ssup2/kpexec
+  shortDescription: Execute process with privileges in a pod
+  description: |
+    kpexec is a kubernetes cli that runs commands in a container with high
+    privileges. kubectl-exec runs the command with the same privileges as the
+    container. For example, if a container does not have network privileges, the
+    command executed by kubectl-exec also has no network privileges. Also,
+    kubectl-exec does not provide an option to force command to run with high
+    privileges. This makes debugging the pod difficult.
```
corneliusweig, line 14: I'd move the `kubectl exec` discussion after you have introduced `kpexec`.

plugins/kpexec.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    kpexec helps execute pod debugging smoothly by executing commands with high
+    privileges regardless of container privileges. Also, kpexec supports tools
```
corneliusweig, line 17: Maybe add something like this:

`kpexec` runs a highly privileged container on the same node as the target container, and joins into the namespaces of the target container (IPC, UTS, PID, net, user, mount).

corneliusweig (commented):
Thanks for submitting this nice debugging tool!

This looks similar to other plugins out there:
- https://github.com/aylei/kubectl-debug
- https://github.com/verb/kubectl-debug (now an alpha `kubectl` command)
- https://github.com/kvaps/kubectl-node-shell

In general I think this plugin is very useful, and we should merge it. However, long-term I expect this command to be superseded by `kubectl debug`. In fact, our `debug` plugin is now an alpha command in `kubectl` and we will remove the `debug` plugin at some point.

What your plugin does, should also be available in plain `kubectl debug`. So how would you feel about making a KEP and upstream your command into `kubectl` (or find out if one exists already)? Overall, this seems to be worthwhile, because running privileged pods in a cluster has some trust issues. If this were part of `kubectl`, these trust problems would be less problematic.

ssup2:
@corneliusweig Thanks for your feedback :) I applied all of your feedback. Thanks.

https://github.com/kubernetes/kubernetes/issues/30656
As you see, running a privileged process in the pod is a reluctant issue. 
So I am not confident in putting kpexec into kubectl.

I don't care if kubectl-debug will replace kpexec in the future.

plugins/kpexec.yaml
```diff
@@ -0,0 +1,62 @@
...
+    target container. kpexec runs a highly privileged container on the same node as
+    the target container and joins into the namespaces of the target container
+    (IPC, UTS, PID, net, mount).
+
+    kubectl-exec runs the command with the same privileges as the
+    container. For example, if a container does not have network privileges,
+    the command executed by kubectl-exec also has no network privileges. Also,
+    kubectl-exec does not provide an option to force command to run with high
+    privileges. This makes debugging the pod difficult. If you use kpexec instead
+    of kubectl-exec, you can easily perform debugging.
```
corneliusweig, line 23:
```suggestion
    In contrast, kubectl-exec runs the command with the same privileges as the
    container. For example, if a container does not have network privileges,
    the command executed by kubectl-exec also has no network privileges. 
    This makes debugging the pod difficult. If you use kpexec instead
    of kubectl-exec, you can easily get high privileges for debugging.
```

plugins/kpexec.yaml
```diff
@@ -0,0 +1,62 @@
...
+  shortDescription: Execute process with privileges in a pod
+  description: |
+    kpexec is a kubernetes cli that runs commands in a container with high
+    privileges. kpexec helps execute pod debugging smoothly by executing commands
+    with high privileges regardless of container privileges. Also, kpexec has a
+    tools mode that adds useful debugging tools in the debugged container. The
+    tools mode is useful when there necessary debugging tools are missing in the
+    target container. kpexec runs a highly privileged container on the same node as
+    the target container and joins into the namespaces of the target container
+    (IPC, UTS, PID, net, mount).
```
corneliusweig, line 16:
```suggestion
    kpexec is a kubernetes cli that runs commands in a container with high
    privileges. It runs a highly privileged container on the same node as
    the target container and joins into the namespaces of the target container
    (IPC, UTS, PID, net, mount). This is useful for debugging where you often
    need to execute commands with high privileges. Also, kpexec has a tools
    mode, which adds useful debugging tools into the debugged container. The
    tools mode is useful when there necessary debugging tools are missing in the
    target container. 
```

corneliusweig (commented):
Overall LGTM. I have some wording suggestions to make the text easier to read & understand.

Why do you use the name `kpexec` by the way? Your binary is called `kubectl-pexec`, and `pexec` would make more sense to me overall (-> "privileged exec"). You won't be able to change the plugin name, so please consider this carefully. I think that `pexec` would be the better plugin name (shorter, easier to remember), but `kpexec` is also acceptable.

---

One afterthought: your plugin relies on the docker containers `ssup2/cnsenter` and `ssup2/cnsenter-tools`. They get pulled by their `latest` tag, which means you don't know which version of the container a user is _actually_ using. I think you would do yourself a favour by pinning the versions of those containers. Then, you can determine the actual versions by the version of the `kpexec` plugin.

---

Btw, if you make more changes to the manifest, can you just add a commit instead of amending? this will make it easier to review.

corneliusweig (commented):
/lgtm
/approve
/hold in case you want to change the name

ssup2:
@corneliusweig I updated the description.

I agree with your opinion about the name. But I want to keep the name. Initially, the project name of this plugin was kubectl-pexec. However, while developing, I found out that the kubectl plugin cannot use autocompletion. (kubectl (tab) pexec (tab) not working) So I changed the name to kpexec and directed it to the Kubernetes cli project, not the kubectl plugin. For this reason, I want to keep the name.

https://github.com/ssup2/kpexec/commit/7b2d6a7bb8b30342151bfddace304882f63bfeb1
I recently patched about kpexec use version of cnsenter pod. kpexec uses the same cnseter image as its version. For example kpexec version is v0.2.7, kpexec uses the 0.2.7 version of cnseter pod.

corneliusweig:
ssup2/kpexec@7b2d6a7.. whoops, I missed that :)
Thanks for sharing your hard work! This really is a nice plugin!
/hold cancel

ssup2:
@corneliusweig @ahmetb I really appreciate your reviews. Thanks :)

PR #1141, opened 2021-03-24
Submit a plugin for scanning container/image

carbonetes:

### DIFF
--- a/plugins/carbonetes-scan.yaml
+++ b/plugins/carbonetes-scan.yaml
@@ -1,0 +1,30 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: carbonetes-scan
+spec:
+  homepage: https://github.com/carbonetes/kubectl-carbonetes-scan
+  shortDescription: Seamlessly integrates complete container analysis directly into your Kubernetes cluster.
+  version: v1.0.0
+  description: |
+    Provides complete container analysis and
+    policy evaluation as a fully managed service.
+    Carbonetes analyzes your container images for native
+    code vulnerabilities, software composition analysis (SCA),
+    license types, malware, secrets, and bill of materials.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/carbonetes/kubectl-carbonetes-scan/archive/refs/tags/v1.0.0.tar.gz
+    sha256: ee9152b7920cb8524ecc3ae8fe96721a266c48da759ba3d802b58f77cd3ad2c2
+    bin: kubectl-carbonetes-scan
+    files:
+    - from: kubectl-carbonetes-scan-*/kubectl-carbonetes-scan
+      to: .
+    - from: kubectl-carbonetes-scan-*/LICENSE
+      to: .


carbonetes:
I've signed it. Thanks.

plugins/carbonetes-scan.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: carbonetes-scan
+spec:
+  homepage: https://github.com/carbonetes/kubectl-carbonetes-scan
+  shortDescription: Seamlessly integrates complete container analysis directly into your Kubernetes cluster.
```
ahmetb, line 7: shortening by half would be ideal, lots of redundant/subjective words in there.

plugins/carbonetes-scan.yaml
```diff
@@ -0,0 +1,30 @@
...
+  version: v1.0.0
+  description: |
```
ahmetb, line 9: by examining your source code I can tell there's a huge caveat in this plugin: This plugin requires Docker on your machine to work.

Make use of `caveats:` section and check if `docker` is installed, otherwise error properly please.

ahmetb: also, what's the reason why you don't distribute the `carbonetes` binary natively for linux/mac and just let users use that? what's the point of docker-run in this case?

plugins/carbonetes-scan.yaml
carbonetes: Thanks for the review. We'll make a change to this description.

plugins/carbonetes-scan.yaml
carbonetes: The reason why we used docker-run is that we already have an existing docker image on the docker hub, which we were currently using on our plugins on other software development tools. We want to maintain these plugins simultaneously when we have some new features, so we've decided to use the same image for creating other plugins.

ahmetb:
So it seems this is a very light shim on top of a `docker run` command. Typically we reject plugins that are just light shims on top of a heavyweight process (such as those installing software, or bootstrapping programs).

kubectl plugins typically do meaningful work standalone, but in this case this plugin is parsing an image name from a pod and passing it to `docker run [carbonetes-image]` in a shell script. I'm not sure if that's enough value add, but let's debate.
/kind gray-area
/cc @corneliusweig

carbonetes:
> So it seems this is a very light shim on top of a `docker run` command. Typically we reject plugins that are just light shims on top of a heavyweight process (such as those installing software, or bootstrapping programs).
> 
> kubectl plugins typically do meaningful work standalone, but in this case this plugin is parsing an image name from a pod and passing it to `docker run [carbonetes-image]` in a shell script. I'm not sure if that's enough value add, but let's debate.
> /kind gray-area
> /cc @corneliusweig

I've already updated the manifest file and make use of the **caveats:** section. Also, we added checking if Docker is installed as you prescribed. I hope these changes add some weight for your approval.

corneliusweig:
Thanks for clarifying this and making the edits.

However, I'd still like to understand why this plugin needs to be run via `docker run`. The much preferred way would be to ship a monolithic binary that contains the logic.

One drawback of your current approach is that you don't know which version is actually being executed. You run `docker run carbonetes/carbonetes-scan:latest`. This is bad for two reasons:

- when running for the first time, you don't know which version will be pulled
- if the image was pulled before, it won't be pulled again

The bottom line is that you never know what version really runs when a user has problems with our plugin.

---

If you insist that distribution via docker containers is the only viable approach for you, I recommend to:

- specify the exact version tag in your script, ideally with the version sha:  `docker run yourimage@sha256:12345fe`
- or use the `--pull=always` option of docker >=19.09 (not so nice because it downloads the image every time!)

carbonetes:
> Thanks for clarifying this and making the edits.
> 
> However, I'd still like to understand why this plugin needs to be run via `docker run`. The much preferred way would be to ship a monolithic binary that contains the logic.
> 
> One drawback of your current approach is that you don't know which version is actually being executed. You run `docker run carbonetes/carbonetes-scan:latest`. This is bad for two reasons:
> 
> * when running for the first time, you don't know which version will be pulled
> * if the image was pulled before, it won't be pulled again
> 
> The bottom line is that you never know what version really runs when a user has problems with our plugin.
> 
> If you insist that distribution via docker containers is the only viable approach for you, I recommend to:
> 
> * specify the exact version tag in your script, ideally with the version sha:  `docker run yourimage@sha256:12345fe`
> * or use the `--pull=always` option of docker >=19.09 (not so nice because it downloads the image every time!)

Hi @corneliusweig,

Since you want kubectl plugins to do significant work independently, we have to make changes to our code. We'll remove the **docker run** if that makes our plugin incorrect.

Our apologies for insisting to get approval although in the first place you've already mentioned that there's something wrong with our code.

The next thing is what we'll be going to do with this PR? Do we just close this or you need to reject this PR?

Or we'll just need to update our code for this PR?

ahmetb:
I think the recommended action would be to distribute your plugin directly as a binary (like other plugins do) as opposed to a script that eventually runs "Docker run"￼.


If that doesn't work for you, we can close this PR, but can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository (instead of krew-index).

carbonetes:
Hi, @ahmetb and @corneliusweig,

We've now removed the **docker run** on our code as you prescribed.

Please review again our code.

Thanks.

plugins/carbonetes-scan.yaml
```diff
@@ -5,15 +5,17 @@ metadata:
...
     Provides complete container analysis and
     policy evaluation as a fully managed service.
     Carbonetes analyzes your container images for native
     code vulnerabilities, software composition analysis (SCA),
     license types, malware, secrets, and bill of materials.
   caveats: |
-    Docker needs to be installed on your machine to use this plugin.
+    You need to have a valid credentials at Carbonetes to execute this plugin.
+    Doesn't have any credentials yet?
+    Register here: https://carbonetes.com
```
corneliusweig, line 18: 🤔 This seems to be a new requirement. Is that really necessary?
I think no existing plugin requires an account for a third-party site. I was looking if we explicitly require that plugins are open source, but could not find it. Let's hear more opinions.
/cc @ahmetb @chriskim06

corneliusweig (commented):
Do I understand correctly, that the analysis works by sending a bunch of details to an analysis server?
https://github.com/carbonetes/kubectl-carbonetes-scan/blob/66d21034c9a305c5a2dcefc396bbb8a1cd9cfd19/kubectl-carbonetes-scan#L485

This needs to be mentioned in `caveats` section, because the default assumption is that the only communication is between the k8s control plane and the plugin. But not third parties.

ahmetb:
@corneliusweig I think quite a few plugins by AquaSecurity or ReplicatedHQ follow the same model (sending your data to a remote server or requiring an account).

corneliusweig:
> @corneliusweig I think quite a few plugins by AquaSecurity or ReplicatedHQ follow the same model (sending your data to a remote server or requiring an account).

IIUC that's not the case. The closest I could find was in `support-bundle`:

> This archive can be manually inspected or uploaded to https://vendor.replicated.com for automated analysis.

So no automatic upload, and still a choice for manual inspection. There's also aquasecurity/starboard, which builds on trivy---but AFAICT this operates locally.

---

@carbonetes In any case, before we can continue evaluating your submission, you have to sign the Kubernetes CLA. 😉

carbonetes:
I signed it

carbonetes:
Hi,

Any updates on this?

Thanks.

plugins/carbonetes-scan.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: Plugin
+metadata:
+  name: carbonetes-scan
+spec:
+  homepage: https://github.com/carbonetes/kubectl-carbonetes-scan
+  shortDescription: Integrates container analysis directly into your cluster.
+  version: v1.1.0
+  description: |
+    Provides complete container analysis and
+    policy evaluation as a fully managed service.
```
corneliusweig, line 11: What kind of policies is that? Do users have to specify them?

plugins/carbonetes-scan.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: carbonetes-scan
+spec:
+  homepage: https://github.com/carbonetes/kubectl-carbonetes-scan
+  shortDescription: Integrates container analysis directly into your cluster.
+  version: v1.1.0
+  description: |
```
corneliusweig, line 9: Please wrap around at ~80 chars, so that it prints nicely on wide and narrow terminals.

plugins/carbonetes-scan.yaml
corneliusweig: Ok, I think an account as prerequisite is fine. However, please make it  clear that the evaluated manifests are sent to a third party.

corneliusweig (commented):
Sorry, this slipped through the cracks. 🐌 

I think we need to be open for commercial plugins, or  plugins that require accounts with a third party. However, please state clearly that your plugin sends manifests to a third party service.

/cc @ahmetb

ahmetb:
> However, please state clearly that your plugin sends manifests to a third party service.

We have several replicatedhq plugins also work in a similar fashion. I am not sure if this warrants an explicit statement, but it's always good to have.

PR #1156, opened 2021-03-29
add kubectl multinet plugin

s1061123:
This PR is to add kubectl-multinet plugin, which shows pod network interface status, defined in multi-net-spec(*1).

*1: https://github.com/k8snetworkplumbingwg/multi-net-spec

   -> Verified with `kubectl krew install --manifest=multinet.yaml`

### DIFF
--- a/plugins/multinet.yaml
+++ b/plugins/multinet.yaml
@@ -1,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: multinet
+spec:
+  version: v0.2.1
+  homepage: https://github.com/k8snetworkplumbingwg/kubectl-multinet
+  shortDescription: "Shows pods' network-status of multi-net-spec"
+  description: |
+    Shows pods' network-status annotation, defined in NPWG multi-net-spec,
+    https://github.com/k8snetworkplumbingwg/multi-net-spec.
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/k8snetworkplumbingwg/kubectl-multinet/releases/download/v0.2.1/kubectl-multinet_0.2.1_linux_amd64.tar.gz
+    sha256: 71f9a63ed36b68eec64ec9e36be09ec95f846412799f95b9b93e69b85bd4f8a3
+    bin: kubectl-multinet
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm64
+    uri: https://github.com/k8snetworkplumbingwg/kubectl-multinet/releases/download/v0.2.1/kubectl-multinet_0.2.1_linux_arm64.tar.gz
+    sha256: c7c1946983f24051cce985efaeb94c724668862831cd1c4dce61439502083090
+    bin: kubectl-multinet


plugins/multinet.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: multinet
+spec:
+  version: v0.2.1
+  homepage: https://github.com/k8snetworkplumbingwg/kubectl-multinet
+  shortDescription: "Shows pods' network-status annotation of NPWG multi-net-spec"
```
ahmetb, line 8: This might be tad bit too long drop 1-2 words to fit into "krew list" cmd.

ahmetb (commented):
Thanks it'll be reviewed.

plugins/multinet.yaml
```diff
@@ -0,0 +1,25 @@
...
+  description: |
+    Shows pods' network-status annotation, defined in NPWG multi-net-spec.
```
ahmetb, line 10: You can link to the spec link here if you want.

plugins/multinet.yaml
```diff
@@ -0,0 +1,25 @@
...
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
```
ahmetb, line 14: most plugins also provide a darwin build for macOS (many Kubernetes users use mac).

ahmetb:
I've tried this plugin in my cluster but could not get the outputs described in your repo. Could be because I don’t use the Multi-Network CRD/proposal?

So can you please elaborate if this plugin is a temporary one –or do you intend on keeping it on Krew even after the proposal is accepted to Kubernetes project (is that the intention)? Clarifying your intention/long-term plan can help us.

plugins/multinet.yaml
s1061123: Drop two words

plugins/multinet.yaml
s1061123: Done.

plugins/multinet.yaml
s1061123: Will do but it might be a bit later, not now.

s1061123 (commented):
@ahmetb thank you for your comment! Changed a diff. PTAL?

s1061123:
@ahmetb This netwrok annotation can be used with the some CNI which support multi-net-spec. This spec is kinds of de-facto standard, not formal specification, however, it is already supported [multus-cni](https://github.com/k8snetworkplumbingwg/multus-cni) and other CNIs.

This is not temporally one, I will try to update as long as (NPWG](https://github.com/k8snetworkplumbingwg/) is active.

ahmetb:
Great. Would you like to add some more info about how the users can get started to "description"?

Also when I tried this I ran it without args, and got no output. From UX perspective it would be better to steer users in the right direction (with more docs links and examples).

I can merge as is, but recommend elaborating more in "description".
/approve

plugins/multinet.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: multinet
+spec:
+  version: v0.2.1
+  homepage: https://github.com/k8snetworkplumbingwg/kubectl-multinet
+  shortDescription: "Shows pods' network-status of multi-net-spec"
+  description: |
+    Shows pods' network-status annotation, defined in NPWG multi-net-spec, https://github.com/k8snetworkplumbingwg/multi-net-spec.
```
corneliusweig, line 10: Please wrap around at ~80 chars, so that it prints nicely on wide and narrow terminals.

s1061123: fixed. thanks.

ahmetb:
/lgtm
/approve
Thanks!

PR #1157, opened 2021-03-29
Change plugin name from kpexec to pexec

ssup2:
Change plugin name from kpexec to pexec


ahmetb:
Why the change? Just so you know we don't do this normally. We might because the plugin is probably not installed by anyone but it will break existing users.

ssup2:
@ahmetb 
Sorry. This is my mistake. 
I expected worked with "kubectl pexec" command.
Is it impossible to change your name? If so, I will close this PR.

corneliusweig:
FWIW, I think this is the better name.

---

@ssup2 Since I advised you to choose this name but you didn't in the original PR, I'm curious why you changed your mind. In particular, I'd like to understand if I didn't communicate clearly. Maybe our docs need to be clearer around how the plugin will be invoked?

ssup2:
@corneliusweig When I pushed the first PR, I didn't recognize that the plugin name became the binary name. (name: aaa -> kubectl-aaa -> kubectl aaa). And after the first PR is merged, I realized that. This is my mistake. If I had known it, I would accept your advice.

In the developer guild, Is there a guide about the relationship between plugin name and binary name except under the scope naming rule? Before I write the manifest file, I read the manifest guide (https://krew.sigs.k8s.io/docs/developer-guide/plugin-manifest), But I can't find it.

corneliusweig:
@ssup2 Thanks for sharing this. I opened https://github.com/kubernetes-sigs/krew/pull/707 to make this clearer. Can you comment if this would have helped you to understand the naming mechanics better?

ssup2:
@corneliusweig Sure. Thanks.

PR #1175, opened 2021-04-06
Add kubectl-macro plugin to krew index

morningspace:
Hi,

I'd like to raise the request to add kubectl-macro plugin to krew index. It is a kubectl plugin to run `kubectl macro` (or `macro` for short) that wraps a set of kubectl calls into one command to be run repeatedly.

More details can be found from the [online documentation](https://morningspace.github.io/kubemacro/docs/).
The source code repository can be found [here](https://github.com/morningspace/kubemacro/).

There is also a website, called [KubeMacro Hub](https://morningspace.github.io/kubemacro-hub/), essentially a [GitHub repository](https://github.com/morningspace/kubemacro-hub/) too, which is designed as a central place for people to exchange the macros.

I've gone through the guidelines and verified the installation locally. Let me know if that makes sense and I'm happy to make any change per request.

Thanks in advance.

### DIFF
--- a/plugins/macro.yaml
+++ b/plugins/macro.yaml
@@ -4,7 +4,7 @@ metadata:
   name: macro
 spec:
   homepage: https://github.com/morningspace/kubemacro
-  shortDescription: "Wrap a set of kubectl calls into one command and run as a macro."
+  shortDescription: "Wrap multiple kubectl calls and run as a macro"
   version: "v0.1.0"
   description: |
     Run kubectl macro (macro for short) that wraps a set of 
--- a/plugins/macro.yaml
+++ b/plugins/macro.yaml
@@ -4,7 +4,7 @@ metadata:
   name: macro
 spec:
   homepage: https://github.com/morningspace/kubemacro
-  shortDescription: "Wrap multiple kubectl calls and run as a macro"
+  shortDescription: "Wrap multiple kubectl calls to run as a macro"
   version: "v0.1.0"
   description: |
     Run kubectl macro (macro for short) that wraps a set of 
--- a/plugins/macro.yaml
+++ b/plugins/macro.yaml
@@ -1,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: macro
+spec:
+  homepage: https://github.com/morningspace/kubemacro
+  shortDescription: "Wrap a set of kubectl calls into one command and run as a macro."
+  version: "v0.1.0"
+  description: |
+    Run kubectl macro (macro for short) that wraps a set of 
+    kubectl calls into one command that you can run from the
+    command line as many times as you want.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: "os"
+        operator: "In"
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/morningspace/kubemacro/archive/v0.1.0.tar.gz
+    sha256: 8b691712abf5a08e69775e13c10615922043cbecf941191c945327f2cef1b4f4
+    bin: kubectl-macro.sh
+    files:
+    - from: "/kubemacro-*/kubectl-macro.sh"
+      to: "."
+    - from: "/kubemacro-*/LICENSE"
+      to: "."


ahmetb:
Thanks for your submission. So this is basically like a plugin that has its own plugin ecosystem, correct￼?

plugins/macro.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: macro
+spec:
+  homepage: https://github.com/morningspace/kubemacro
+  shortDescription: "Wrap a set of kubectl calls into one command and run as a macro."
```
ahmetb, line 7: I think you need to make this shorter, it probably doesn't fit as it is now. Please eliminate redundant words.

morningspace:
Thanks @ahmetb for your prompt reply!

Ideally, it'd be nice if it could ultimately evolve to an ecosystem:-) For now, it's just simply a kubectl plugin as a runner of "macro", and a place for people to exchange macros when needed. Of course, if the macro is only for private use, it can be kept locally and doesn't have to be shared via the hub. The hub is just a convenient place for people who have common requirements and don't want to reinvent the wheels.

A bit more background of the idea of this plugin:

When work w/ clusters using kubectl, I found it's very often that some common tasks need more than one kubectl call to complete. To repeat the same steps one by one manually every time is tedious. Also, it doesn't have to be a full-fledged go program, or a kubectl plugin, just wrapped as a shell function is sufficient, which must be easy to copy and paste, easy to share. So, I use the term `macro` to express the idea: wrap up multiple kubectl calls into a shell function that can be run as a whole from command line in repeatable manner. It's kind of complement to the kubectl plugin mechanism. Of course, for more complicated scenario, to implement as kubectl plugin is a better option.

Hope that helps!

plugins/macro.yaml
morningspace: Sure, it is revised now, would it look better?

corneliusweig:
Hi @morningspace,

Thanks again for submitting your plugin to `krew-index`. This is indeed a very original plugin, and no other plugin really compares :).
However, we are undecided if `macro` is a good fit. Let me give you some background:
- Macros could apply to anything, not just `kubectl`. It is therefore not clear that they will remain focused on `kubectl`.
- Shell scripts are easy to inspect by users, but they often come with big caveats. They often are not portable between shell versions, not portable to other OSes, and testing is problematic. Krew therefore does not want to endorse shell-script plugins.
- Shell snippets are often better distributed as a library of shell functions or aliases. This especially applies to trivial functions of a few lines.

On the other hand, `macro` does fill a niche that `krew` intentionally does not want to fill: we require that plugins are non-trivial, and we have rejected a few trivial ones. The `macro` hub would be a better place for such micro-plugins. So there can be a nice interplay between krew/ macro + krew-index/ macro hub.

As you see, this is not an easy decision. We therefore want to delay this to a later point and re-evaluate `macro` in 3-6 months. If there is good traction behind `macro` and we see that the ecosystem works well, this will be a great addition. So try grow your idea and evangelize it to a broader community 🚀 

/hold

morningspace:
@corneliusweig Thanks for your great comments and I totally understand your points. Also thanks @ahmetb for the early review. Just one thing I'd like confirm:

So, it's not the "macros" but the "runner" that I'd like to contribute to `krew index`. It is wrapped as kubectl plugin and used to execute the macros. I am thinking `krew index` as an official place for people looking for variant kubectl plugins, so there could be potential users for this plugin and I can get feedback from them.

For macros, or specifically `kubectl macros`, definitely they should not belong to this index. Sorry for the newly-coined "word". There could be better naming, but just currently I'm trying to use this as the metaphor to explain my idea. It should not be a general macro, but a macro dedicated for making kubectl calls.

P.S.

Re: The reason I'm intentionally using shell to implement this idea is because:
1) People usually use kubectl from command line in shell env, and using shell to wrap kubectl calls into an executing block seems the most intuitive idea for most people.
2) More important, I'd like to clearly differentiate `kubectl macro` from `kubectl plugin` to avoid making people feel this is going to replace one by another. They have different focuses. With that, next time when someone submits a trivial shell-script plugin to krew-index, maybe we could suggest them to switch to [kubemacro-hub](https://github.com/morningspace/kubemacro-hub) :-D

I'm going to write a post on [my personal website](http://morningspace.github.io/) to evangelize this idea. But let me know if you have any better suggestion as I'm very interested about how people think about it from broader community. Thx!

Re: Shell snippets distributed as a library. Exactly. Actually I do have an idea to put multiple macros into one bundle (sort of macro bundle), but just hasn't been implemented yet :-)

corneliusweig:
Sorry for my late reply. Yes, you made it clear that this submission is about the macro _runner_, and not the macros run by plugin macro. Phew.. naming is hard! :wink:

morningspace:
@corneliusweig Not a problem at all and thanks for your reply. Yeah, naming is hard :-)

So, do you think if that's ok to get this _runner_ added to krew index? I thought it could get more visibility if it's available in krew-index, so that can get more feedback from community on this idea. Meanwhile, I was preparing some posts re: this topic on medium.com. My first post on medium.com: [A Quick Introduction to KubeMacro](https://morningspace.medium.com/a-quick-introduction-to-kubemacro-f1d878db17a3) :-) Let me know if you have any other good suggestion. Thanks!

PR #1176, opened 2021-04-06
feat: add new plugins translate and move2kube

HarikrishnanBalagopal:
`kubectl-move2kube` is a plugin made by Konveyor to help users migrate their apps to run on K8s.
`kubectl-translate` is a plugin with a subset of the features of `kubect-move2kube`, for simpler use cases.

The plugins can containerize applications and generate K8s yamls, Helm charts, CI/CD pipelines and Operators required to migrate an existing application to K8s.

The source is available at:
https://github.com/konveyor/move2kube

I tested the plugin installation locally on macos and linux/amd64 servers with installing locally and through the URL in the manifest.

I read the Naming Guide, it mentioned not adding `kube` to the plugin name but in this case `move2kube` is the name of the CLI tool.

Signed-off-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

### DIFF
--- a/plugins/move2kube.yaml
+++ b/plugins/move2kube.yaml
@@ -1,0 +1,31 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: move2kube
+spec:
+  version: v0.2.0-alpha.5
+  homepage: https://move2kube.konveyor.io/
+  shortDescription: Migrate your application to run on Kubernetes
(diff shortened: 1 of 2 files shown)


ahmetb:
Please submit them separately, as the evaluation would be independent.
/close

ahmetb:
I suspect we wouldn't be assigning you the name `translate` if it's something that only works with the other `move2kube` plugin.

PR #1179, opened 2021-04-06
feat: add a new translate plugin

HarikrishnanBalagopal:
Resubmitting the closed PR as 2 separate PRs (https://github.com/kubernetes-sigs/krew-index/pull/1176)

This is the `translate` plugin developed by Konveyor.
It helps users migrate their application to k8s.
It can containerize applications and generate the necessary K8s yamls, Helm charts, CI/CD pipelines, Operators, etc.
It can also convert existing k8s yamls to a different version.

Signed-off-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

### DIFF
--- a/plugins/translate.yaml
+++ b/plugins/translate.yaml
@@ -1,0 +1,32 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: translate
+spec:
+  version: v0.2.0-alpha.5
+  homepage: https://move2kube.konveyor.io/
+  shortDescription: Migrate your application to run on Kubernetes
+  description: |
+    Translate creates all the resources required for deploying your application to Kubernetes, including containerization and Kubernetes resources.
+    It supports translating from Docker Swarm/Docker Compose, Cloud Foundry apps and even other non-containerized applications.
+    This plugin contains a subset of the features of a more flexible CLI tool called Move2Kube https://github.com/konveyor/move2kube
+    For more documentation and support for this plugin and Move2Kube, visit https://move2kube.konveyor.io/
+  caveats: |
+    * Optional dependencies:
+      - docker
+      - operator-sdk
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/konveyor/move2kube/releases/download/v0.2.0-alpha.5/kubectl-translate-v0.2.0-alpha.5-darwin-amd64.tar.gz
+      sha256: f017ca365e296ed2ce7bc814f8b0572ff1ec051dda83216b836ef4dac3f0e684
+      bin: kubectl-translate/kubectl-translate
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/konveyor/move2kube/releases/download/v0.2.0-alpha.5/kubectl-translate-v0.2.0-alpha.5-linux-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
This sounds like a highly opinionated tool that has its own way of doing things and the description sounds hand-wavy (that I cannot explain how exactly it works and what specifically it does). That’s a potential problem with the style of the plugin. (We don't prefer all-in-one plugins.)

Second, per our naming guidelines, our question would be: What if someone else comes and says "I also have a `translate` plugin that migrates to Kubernetes in another way." You’re potentially trying to take advantage of being a first-comer that claims this broad name.

That’s why I think we’d recommend you to have a more specific name, such as `konveyor-translate` where users can tell what’s going on ("it's about konveyor project") as opposed to saying "I don't know what this `translate` does".

/kind gray-area

ashokponkumar:
> This sounds like a highly opinionated tool that has its own way of doing things and the description sounds hand-wavy (that I cannot explain how exactly it works and what specifically it does). That’s a potential problem with the style of the plugin. (We don't prefer all-in-one plugins.)
> 
> Second, per our naming guidelines, our question would be: What if someone else comes and says "I also have a `translate` plugin that migrates to Kubernetes in another way." You’re potentially trying to take advantage of being a first-comer that claims this broad name.
> 
> That’s why I think we’d recommend you to have a more specific name, such as `konveyor-translate` where users can tell what’s going on ("it's about konveyor project") as opposed to saying "I don't know what this `translate` does".
> 
> /kind gray-area

Thanks @ahmetb for the comments. 

The goal of this plugin is to help users create their artifacts to make their application run in their kubernetes cluster. As you might have noticed in [our website](https://move2kube.konveyor.io/), this includes artifacts such as kustomize yamls, Helm charts, CI/CD yamls etc. So the goal is singular, "create artifacts that will help accelerate the journey to kubernetes". But we can definitely improve the description based on your feedback.

The reason we went with "translate" is that, the "move2kube" tool has multiple sub-commands like collect, plan, translate, validate, version etc. But for a normal user, who is trying to get their application running on Kubernetes, they can live with just the single sub-command "translate", so we took that sub-command alone and made a plugin, since it will suffice most users requirements, without the users needing to type `kubectl move2kube translate` everytime. Also it does have the version conversion feature of "kubectl convert" which has been deprecated, so we thought it is a name which captures both the above features. We plan to also put out the entire `move2kube` tool as a plugin, but thought for our users, having sub-command within a sub-command is not an ideal experience. Please advise on what is the standard practise here.

[Konveyor](https://www.konveyor.io/) is the open source community that Move2Kube is part of.

ahmetb:
> But for a normal user, who is trying to get their application running on Kubernetes, they can live with just the single sub-command "translate"

Typically this is not a kubectl plugin.
Converting an application into Kubernetes ("translate") is a one-off operation per-application, so it is not really a part of the kubectl experience.

> As you might have noticed in [our website](https://move2kube.konveyor.io/), this includes artifacts such as kustomize yamls, Helm charts, CI/CD yamls etc.

Again, this sounds a lot like bootstrapping/one-off scenario, for which we don’t accept plugins for. We tend to think of kubectl plugins as tools to interact with Kubernetes on a regular basis. 

Imagine you have a tool (like Cloud-Native buildpacks `pack` CLI) that takes an application source code, converts it to a container image (by creating Dockerfile –or not), pushes the image and gives you Kubernetes Deployment manifest, that would not be a plugin that we’d be accepting here.

HarikrishnanBalagopal:
@ahmetb Thanks for looking at it. We can close this then.

PR #1202, opened 2021-04-13
Add new plugin resource-versions

chengshiwen:
`kubectl-resource-versions` is a plugin made by [Shiwen Cheng](https://github.com/chengshiwen) to help users find out the API resources along with the supported API versions on the server as shown below.:

image

The source is available at:
https://github.com/chengshiwen/kubectl-resource-versions

### DIFF
--- a/plugins/resource-versions.yaml
+++ b/plugins/resource-versions.yaml
@@ -1,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: resource-versions
+spec:
+  version: v0.1.0
+  platforms:
+  - bin: kubectl-resource-versions
+    uri: https://github.com/chengshiwen/kubectl-resource-versions/releases/download/v0.1.0/kubectl-resource-versions-darwin-amd64.tar.gz
+    sha256: 407d562fa3c34f49bc1b101294d06fec13639ff95c23bb2d071a5326f67de9af
+    files:
+      - from: "kubectl-resource-versions-darwin-amd64/*"
+        to: .
+    selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+  - bin: kubectl-resource-versions
+    uri: https://github.com/chengshiwen/kubectl-resource-versions/releases/download/v0.1.0/kubectl-resource-versions-linux-amd64.tar.gz
+    sha256: ecbbbd0892dd2e49f4b487ff92c95e2c322e3eb7b02c78b8354aef216404d643
+    files:
+      - from: "kubectl-resource-versions-linux-amd64/*"
+        to: .
+    selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+  - bin: kubectl-resource-versions.exe
+    uri: https://github.com/chengshiwen/kubectl-resource-versions/releases/download/v0.1.0/kubectl-resource-versions-windows-amd64.tar.gz
+    sha256: 062eee95e959de5d0c0230a51ab03a05d2f992d070bed5dd0d75311cfd1835c5
+    files:
+      - from: "kubectl-resource-versions-windows-amd64/*"
+        to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: Print the supported API resource versions on the server
+  homepage: https://github.com/chengshiwen/kubectl-resource-versions
+  caveats: |
+      Usage:
+        kubectl resource-versions [flags]
+  description: |
+      Print the supported API resource versions on the server
+
+      Have you ever wondered which api group and version should be used for a specified resource on a provided kubernetes cluster?
+      It is useful to find out the matrix of resources and api versions by using this kubectl plugin as shown below:
+
+      $ kubectl resource-versions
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks, good contribution! I think this is great. I personally would love to see this in `kubectl api-version` but it seems like it doesn't support that.

Can you please import `client-go/auth` plugins? Otherwise this won't work on many cloud clusters.

plugins/resource-versions.yaml
```diff
@@ -0,0 +1,49 @@
...
+    uri: https://github.com/chengshiwen/kubectl-resource-versions/releases/download/v0.1.0/kubectl-resource-versions-windows-amd64.tar.gz
+    sha256: 062eee95e959de5d0c0230a51ab03a05d2f992d070bed5dd0d75311cfd1835c5
+    files:
+      - from: "kubectl-resource-versions-windows-amd64/*"
+        to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: Print the supported API resource versions on the server
```
ahmetb, line 38: > Print available versions of each API

shorter & better?

plugins/resource-versions.yaml
```diff
@@ -0,0 +1,49 @@
...
+  homepage: https://github.com/chengshiwen/kubectl-resource-versions
+  caveats: |
```
ahmetb, line 40: we already show what you typed here after installation. please delete.

plugins/resource-versions.yaml
```diff
@@ -0,0 +1,49 @@
...
+      Usage:
+        kubectl resource-versions [flags]
+  description: |
+      Print the supported API resource versions on the server
+
+      Have you ever wondered which api group and version should be used for a specified resource on a provided kubernetes cluster?
+      It is useful to find out the matrix of resources and api versions by using this kubectl plugin as shown below:
+
+      $ kubectl resource-versions
```
ahmetb, line 49: This is not super helpful. You can simply say:

> This plugin prints the API resources available on the API server along with groups/versions that offer that API.

Again, usage string not necesary.
Also make sure to wrap at 80 chars.

ahmetb (changes requested):

chengshiwen:
> client-go/auth

Yes, I have

plugins/resource-versions.yaml
chengshiwen: Done

plugins/resource-versions.yaml
chengshiwen: Done

plugins/resource-versions.yaml
chengshiwen: Done

plugins/resource-versions.yaml
```diff
@@ -0,0 +1,40 @@
...
+    files:
+      - from: "kubectl-resource-versions-windows-amd64/*"
+        to: .
+    selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+  shortDescription: Print available versions of each API
+  homepage: https://github.com/chengshiwen/kubectl-resource-versions
+  description: This plugin prints the API resources available on the API server along with groups/versions that offer that API
```
corneliusweig, line 40: Please wrap around at ~80 chars, so that it prints nicely on wide and narrow terminals.

corneliusweig (approved):
Thanks for you submission! At first I though, "well, isn't this `api-version` command"? This plugin fills a hole in that command, so it's a good plugin. Naming-wise, I'd propose to stay closer to the original. So what do you think about `all-api-versions`? To me, resources has the connotation of a concrete instance, but apis are just the resource "type". So `resource-versions` is not ideal I think.

---

From an ecosystem perspective: Have you tried to add this upstream to `kubectl` as well? If you can get this into `kubectl` itself, your work could have a much bigger impact.

chengshiwen:
> Thanks for you submission! At first I though, "well, isn't this `api-version` command"? This plugin fills a hole in that command, so it's a good plugin. Naming-wise, I'd propose to stay closer to the original. So what do you think about `all-api-versions`? To me, resources has the connotation of a concrete instance, but apis are just the resource "type". So `resource-versions` is not ideal I think.
> 
> From an ecosystem perspective: Have you tried to add this upstream to `kubectl` as well? If you can get this into `kubectl` itself, your work could have a much bigger impact.

Thanks for your review. I will consider submitting the changes to https://github.com/kubernetes/kubectl. For the plugin name, I will reconsider it, thanks for the suggestion again!

plugins/resource-versions.yaml
chengshiwen: Done, the modified description is `Print the API resources along with the supported API versions on the server`

corneliusweig:
From krews side this is ready to go. However let's avoid renaming a plugin as it is bad user experience.
So if you consider renaming the plugin, please make your decision now. :)

chengshiwen:
> From krews side this is ready to go. However let's avoid renaming a plugin as it is bad user experience.
> So if you consider renaming the plugin, please make your decision now. :)

Thanks for your reply. My decision is to keep the current plugin name.

corneliusweig:
/lgtm
/approve
In case this gets added to mainline kubectl, please open an issue here to discuss next steps.

chengshiwen:
> In case this gets added to mainline kubectl, please open an issue here to discuss next steps.

OK

PR #1213, opened 2021-04-16
Remove debug plugin from krew-index

### LINKED ISSUE #949: Remove 'debug' plugin as builtin command graduates to beta in kubectl 1.20.0
`debug` is now a beta command in kubectl 1.20.0 giving the plugin in krew no option to be invoked from this point forward. We should probably delete it soon.

/cc @verb

verb:
This functionality has been available in kubectl for the past
4 releases, first as `kubectl alpha debug` and then `kubectl debug`.
It should be safe to remove at this point.

Fixes #949


ahmetb:
Enough time passed since #949 was proposed, should be ok to delete.
/lgtm
/approve

PR #1217, opened 2021-04-18
Request to merge for 'Container Connector' which is conn 

sysnet4admin:

### DIFF
--- a/plugins/conn.yaml
+++ b/plugins/conn.yaml
@@ -1,0 +1,23 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: conn
+spec:
+  platforms:
+  - sha256: 7aba7524bb10b355b322f8ad9b92d0ef4b4d63ac3480c17c59442c8be096e358
+    uri: https://github.com/sysnet4admin/k8s-plugins/releases/download/v1.0.1/k8s-plugins.zip
+    bin: conn.sh
+    files:
+    - from: "*/conn/*"
+      to: "."
+    selector:
+      matchExpressions:
+      - {key: os, operator: In, values: [darwin, linux]}
+  version: "v1.0.1"
+  homepage: https://github.com/sysnet4admin/k8s-plugins
+  shortDescription: Display pods's list per each namespace and connect to container 
+  description: |
+    Uses 'conn' could connect pod. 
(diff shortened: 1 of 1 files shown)


sysnet4admin:
Hi Reviewer, 

Hopefully have a good sunday.
 
I fixed the issues as below 
Thus if you do not mind, please let me re-run to verify?

krew-index CI / Build and Tags (pull_request) Failing after 35s — Build and Tags
1) I have changed LICENSE name from LICENSE.adoc to LICENSE. 

cla/linuxfoundation — sysnet4admin needs CLA
1) Contribute to CNCF as an individual 

As this guideline: [](https://krew.sigs.k8s.io/docs/developer-guide/release/new-plugin/)

If I have something wrong, please let me know. 

Thank you in advance. 
sysent4admin.

chriskim06:
Hey @sysnet4admin, it seems the archive that your latest manifest points to has an incorrect license file. Also you'll need to make sure your commits are under the email you used to sign the cncf cla (your latest commit seems incorrect). I think you might have to amend your previous commit, but it might also work if you just commit again correctly. Also please take a look through the [naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/), there's a section saying to avoid using abbreviations.

As far as your plugin submission goes, it seems like it's just a utility to get an interactive `sh`. I think we've decided against adding things like this in the past that don't offer a ton of functionality and are just wrappers around a few `kubectl` commands. @ahmetb @corneliusweig feel free to correct me if I'm wrong about that point.

sysnet4admin:
@chriskim06 
Yes you are totally right! So I changed LICENSE name again. 
But there is some unauthorized user issue...(which is same my email address rarely used..)

How could I ignore it?
Please let me advice. Thank you so much in advance!

chriskim06:
We tend to not accept plugins if there isn't a sufficient value add. From what I can see this is just getting namespace, pod, and container to pass as arguments to `kubectl exec -it ... -- /bin/sh`, so I think this falls into that category. 

However, you can still distribute your plugin via Krew using [custom indexes](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/). That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

/close

sysnet4admin:
Thank you for your advice Chirs!

PR #1259, opened 2021-05-02
Add promdump plugin

ihcsim:
This PR is a proposal to add the [promdump](https://github.com/ihcsim/promdump) plugin to krew-index.

promdump is a tool that can be used to dump and restore Prometheus data blocks. It
is different from the `promtool tsdb dump` command in such a way that its
output can be re-used in another Prometheus instance. And unlike the Promethues
TSDB `snapshot` API, promdump doesn't require Prometheus to be started with the
`--web.enable-admin-api` option. Instead of dumping the entire TSDB, promdump
offers the flexibility to filter persistent blocks by time range.

The README in the repos has more information on [Why This Tool](https://github.com/ihcsim/promdump#why-this-tool) and [How It Works](https://github.com/ihcsim/promdump#how-it-works).

There is an active issue at https://github.com/prometheus/prometheus/issues/8753 to gather community 
feedback.

### DIFF
--- a/plugins/promdump.yaml
+++ b/plugins/promdump.yaml
@@ -1,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: promdump
+spec:
+  version: v0.2.2
+  homepage: https://github.com/ihcsim/promdump
+  shortDescription: promdump dumps the head and persistent blocks of Prometheus.
+  description: |
+    promdump dumps the head and persistent blocks of Prometheus. It supports
+    filtering the persistent blocks by time range.
+
+    promdump is a tool that can be used to dump Prometheus data blocks. It is
+    different from the 'promtool tsdb dump' command in such a way that its output
+    can be re-used in another Prometheus instance. And unlike the Promethues TSDB
+    'snapshot' API, promdump doesn't require Prometheus to be started with the
+    '--web.enable-admin-api' option. Instead of dumping the entire TSDB, promdump
+    offers the flexibility to filter persistent blocks by time range.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/ihcsim/promdump/releases/download/v0.2.2/kubectl-promdump-darwin-amd64-v0.2.2.tar.gz
+    sha256: b986f3a28435e5e5af0af9e1e29fd15a36f4500ef883471eed58c81b4dc29bbc
+    bin: kubectl-promdump
+    files:
+    - from: "*"
+      to: .
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/ihcsim/promdump/releases/download/v0.2.2/kubectl-promdump-linux-amd64-v0.2.2.tar.gz
+    sha256: 21ef04b686161f713f0e321ed1ab94de42768be2e69dfb0a4647d492fc264837
+    bin: kubectl-promdump
+    files:
+    - from: "*"
+      to: .
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/ihcsim/promdump/releases/download/v0.2.2/kubectl-promdump-windows-amd64-v0.2.2.tar.gz
+    sha256: ae851b130e7871fd1a5fe1dda5a6c91f722b7a0886d51301f5d78e6cf5ae63a7
+    bin: kubectl-promdump.exe
+    files:
+    - from: "*"
+      to: .
--- a/plugins/promdump.yaml
+++ b/plugins/promdump.yaml
@@ -3,9 +3,9 @@ kind: Plugin
 metadata:
   name: promdump
 spec:
-  version: v0.2.2
+  version: v0.2.3
   homepage: https://github.com/ihcsim/promdump
-  shortDescription: promdump dumps the head and persistent blocks of Prometheus.
+  shortDescription: Dumps the head and persistent blocks of Prometheus.
   description: |
     promdump dumps the head and persistent blocks of Prometheus. It supports
     filtering the persistent blocks by time range.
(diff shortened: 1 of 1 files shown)


plugins/promdump.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: promdump
+spec:
+  version: v0.2.2
+  homepage: https://github.com/ihcsim/promdump
+  shortDescription: promdump dumps the head and persistent blocks of Prometheus.
```
corneliusweig, line 8:
```suggestion
  shortDescription: Dumps the head and persistent blocks of Prometheus.
```
No need to repeat the plugin name.

---

Although your current description works, what do you think about talking about the problem it solves instead. For example:

> Helps to move Prometheus TSDB data between clusters

plugins/promdump.yaml
```diff
@@ -0,0 +1,49 @@
...
+  description: |
+    promdump dumps the head and persistent blocks of Prometheus. It supports
+    filtering the persistent blocks by time range.
+
+    promdump is a tool that can be used to dump Prometheus data blocks. It is
+    different from the 'promtool tsdb dump' command in such a way that its output
+    can be re-used in another Prometheus instance. And unlike the Promethues TSDB
+    'snapshot' API, promdump doesn't require Prometheus to be started with the
```
corneliusweig, line 16:
```suggestion
    'snapshot' command, promdump doesn't require Prometheus to be started with the
```

plugins/promdump.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/ihcsim/promdump/releases/download/v0.2.2/kubectl-promdump-darwin-amd64-v0.2.2.tar.gz
+    sha256: b986f3a28435e5e5af0af9e1e29fd15a36f4500ef883471eed58c81b4dc29bbc
+    bin: kubectl-promdump
+    files:
+    - from: "*"
+      to: .
```
corneliusweig, line 29: This is the default. I'd recommend to drop the `files` spec altogether.

plugins/promdump.yaml
```diff
@@ -0,0 +1,49 @@
...
+    different from the 'promtool tsdb dump' command in such a way that its output
+    can be re-used in another Prometheus instance. And unlike the Promethues TSDB
+    'snapshot' API, promdump doesn't require Prometheus to be started with the
+    '--web.enable-admin-api' option. Instead of dumping the entire TSDB, promdump
+    offers the flexibility to filter persistent blocks by time range.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
```
corneliusweig, line 23: `darwin`+`arm64` has become quite popular. Please consider distributing binaries for that platform too.

plugins/promdump.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: promdump
+spec:
+  version: v0.2.2
+  homepage: https://github.com/ihcsim/promdump
+  shortDescription: promdump dumps the head and persistent blocks of Prometheus.
+  description: |
+    promdump dumps the head and persistent blocks of Prometheus. It supports
+    filtering the persistent blocks by time range.
+
+    promdump is a tool that can be used to dump Prometheus data blocks. It is
```
corneliusweig, line 13: DISCLAIMER: I'm not very familiar with Prometheus.

I like that this description describes what the plugin actually does. However, when you describe how to use the dumped data, you are giving very detailed instructions (`--web.enable-admin-api`), but still not complete. I think it's a great idea to have a detailed howto to get this job done. But the `description` does not give enough space for that (and would be the wrong place IMO). So what do you think about writing a more elaborate guide and just link to this page from here?

corneliusweig (commented):
Overall, this is very nice work. Thanks for sharing this!
Regarding the manifest, I have just minor comments.

Does https://github.com/prometheus/prometheus/issues/8753#issuecomment-830733264 mean that this plugin could at some point be a thin wrapper around `promtool`?

---

As far as I can tell, you didn't import the `auth` plugins for cloud providers. So can you include

```go
import _ "k8s.io/client-go/plugin/pkg/client/auth"
```

See https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/ for more details.

ihcsim:
@corneliusweig Thanks for the feedback. 

> Does prometheus/prometheus#8753 (comment) mean that this plugin could at some point be a thin wrapper around promtool?

Not quite. This kubectl plugin is a wrapper around promdump. promdump can be used to move data between Prometheus instances not running on K8s. (See [How It Works](https://github.com/ihcsim/promdump#how-it-works). ) So if anything, promtool will either wrap promdump or use its `pkg`.

> As far as I can tell, you didn't import the auth plugins for cloud providers.

:+1:; will fix - when reading the doc, it didn't cross my mind that it's a required dependency.

plugins/promdump.yaml
ihcsim: I don't want folks to get the idea that this is a migration or backup/restore tool, because there are still things around restoring the WAL at the TSDB level that I don't understand. At the moment, I see this plugin as a SRE tool to copy data from one Prometheus instance to another. promdump itself isn't limited to just K8s.

plugins/promdump.yaml
ihcsim: There is an actual [`snapshot` HTTP endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#snapshot).

plugins/promdump.yaml
ihcsim: As the tool matures and evolves, I can see this changing.

plugins/promdump.yaml
ihcsim: > However, when you describe how to use the dumped data, you are giving very detailed instructions (--web.enable-admin-api), but still not complete. 

The intention of that paragraph is to show how promdump is different from the Prometheus `snapshot` endpoint which requires Prometheus to be started with the `--web.enable-admin-api` option. Open to suggestion on how to make that clearer.

> So what do you think about writing a more elaborate guide and just link to this page from here?

WDYT of adding a link in the description to the [Getting Started](https://github.com/ihcsim/promdump#getting-started) guide in the README?

plugins/promdump.yaml
ihcsim: The promdump README has a [Limitation](https://github.com/ihcsim/promdump#limitations) section describing what it is currently good for, and not.

corneliusweig:
> :+1:; will fix - when reading the doc, it didn't cross my mind that it's a required dependency.

Well it is not required for local installations. But it will be required for cloud providers.

ihcsim:
@corneliusweig thanks again for the review. I have added all your feedback to the code and manifest.

ahmetb:
I think it looks good now. You can iterate over the description in newer versions later if you'd like.
/lgtm
/approve

PR #1280, opened 2021-05-12
added tunnel plugin to krew

omrikiei:
Added a plugin manifest for a tunnel([ktunnel](https://github.com/omrikiei/ktunnel))

### DIFF
--- a/plugins/tunnel.yaml
+++ b/plugins/tunnel.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: tunnel
+spec:
+  version: v1.3.8
+  homepage: https://github.com/omrikiei/ktunnel
+  shortDescription: Establishes a reverse tunnel between a kubernetes cluster and your local machine
+  description: |
+    This plugin establishes a reverse tunnel between a kubernetes cluster and your local machine. 
+    It lets you expose your machine as a service in the cluster or expose it to a specific deployment as a local host.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/omrikiei/ktunnel/releases/download/v1.3.8/ktunnel_1.3.8_Darwin_x86_64.tar.gz
+    sha256: 8b9aef56834aef9a88599334a24a6853de0bdee76cf9a63eeb467944a428e616
+    bin: ktunnel
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/omrikiei/ktunnel/releases/download/v1.3.8/ktunnel_1.3.8_Linux_x86_64.tar.gz
+    sha256: 303acd7d212996da49f7d8d7b60894fe2ecdf9bfe4054e8a97a86cc9362d6ff7
+    bin: ktunnel
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/omrikiei/ktunnel/releases/download/v1.3.8/ktunnel_1.3.8_Windows_x86_64.tar.gz
+    sha256: 8b42102800fd9c21a364addacf2fd2deecbacdc160cdb7278372e9fbbcbc72dd
+    bin: ktunnel.exe
--- a/plugins/tunnel.yaml
+++ b/plugins/tunnel.yaml
@@ -7,9 +7,9 @@ spec:
   homepage: https://github.com/omrikiei/ktunnel
   shortDescription: Reverse tunneling between cluster and your machine
   description: |
-      This plugin establishes a reverse tunnel between Kubernetes and your local machine. 
-      It lets you expose your machine as a service in the cluster or expose it to a
-specific deployment as a local host.
+      This plugin establishes a reverse tunnel between the Kubernetes cluster
+      and your local machine. It lets you expose your machine as a service in
+      the cluster or expose it to a specific deployment as a local host.
   platforms:
   - selector:
       matchLabels:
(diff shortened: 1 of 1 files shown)


plugins/tunnel.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: tunnel
+spec:
+  version: v1.3.8
+  homepage: https://github.com/omrikiei/ktunnel
+  shortDescription: Establishes a reverse tunnel between a kubernetes cluster and your local machine
```
ahmetb, line 8: Please shorten to fit ~65 chars.

plugins/tunnel.yaml
```diff
@@ -0,0 +1,33 @@
...
+  description: |
+    This plugin establishes a reverse tunnel between a kubernetes cluster and your local machine. 
+    It lets you expose your machine as a service in the cluster or expose it to a specific deployment as a local host.
```
ahmetb, line 11: please line wrap at 80 chars.
Kubernetes capitalized.

ahmetb:
The name `tunnel` is a common word, but not so common in Kubernetes context. I think it is fine to assign to first-comer in this case. @corneliusweig @chriskim06 thoughts?

chriskim06:
I second that, the name seems fine to me. Cool plugin btw 😄

plugins/tunnel.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: tunnel
+spec:
+  version: v1.3.8
+  homepage: https://github.com/omrikiei/ktunnel
+  shortDescription: Establish a reverse tunnel between Kubernetes your local machine
+  description: |
+      This plugin establishes a reverse tunnel between Kubernetes and your local
+machine. It lets you expose your machine as a service in the cluster or expose it to a
+specific deployment as a local host.
```
ahmetb, line 12: I think you need to indent this in. see other manifests as an example.

plugins/tunnel.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: tunnel
+spec:
+  version: v1.3.8
+  homepage: https://github.com/omrikiei/ktunnel
+  shortDescription: Establish a reverse tunnel between Kubernetes your local machine
```
ahmetb, line 8: I will make this shorter for you:

> Reverse tunneling between cluster and your machine

ahmetb (changes requested):

corneliusweig:
I haven't looked carefully yet, so just my 2cent for now:

- We already have a few plugins for tunneling/forwarding/proxying requests. I think it would make sense to pick a more descriptive name, such as `reverse-tunnel` (so something with reverse, because this seems to be the main use case)
- telemetry.io is a pretty popular project for the same purpose. How does your plugin compare to that solution?

omrikiei:
> I haven't looked carefully yet, so just my 2cent for now:
> 
> * We already have a few plugins for tunneling/forwarding/proxying requests. I think it would make sense to pick a more descriptive name, such as `reverse-tunnel` (so something with reverse, because this seems to be the main use case)
> * telemetry.io is a pretty popular project for the same purpose. How does your plugin compare to that solution?

Hey @corneliusweig, which other tunneling plugins are listed? I'd like to take a look. could you also share some reference to telemetry.io? I wasn't able to find anything in that context.. did you mean telepresence.io?

corneliusweig:
Meh.. I should not type on phone  :). Sorry for the confusion.

You're right, I meant telepresence.io. The other tunnel plugins are
- net-forward
- socks5-proxy
- tap

omrikiei:
telepresense substitutes a two-way network proxy for your normal pod running in the Kubernetes cluster.. it would start a dedicated shell and proxy all the resource coupled with the deployment to that shell.
ktunnel is a reverse network proxy and has only this functionality - it can be used for various use cases like [remote debugging](https://github.com/omrikiei/ktunnel/tree/master/examples/pycharm-remote-debugging-on-k8s#readme) (which is why I originally built it), metrics population, file transfer and I've seen [other cool usages as well](http://mbuffett.com/posts/ktunnel-ngrok-replace/).
ktunnel is not only limited for use in kubernetes, and it doesn't replace any remote resources with local ones

plugins/tunnel.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: tunnel
+spec:
+  version: v1.3.8
+  homepage: https://github.com/omrikiei/ktunnel
+  shortDescription: Reverse tunneling between cluster and your machine
+  description: |
+      This plugin establishes a reverse tunnel between Kubernetes and your local machine. 
+      It lets you expose your machine as a service in the cluster or expose it to a
+specific deployment as a local host.
```
corneliusweig, line 12:
```suggestion
      This plugin establishes a reverse tunnel between the Kubernetes cluster
      and your local machine. It lets you expose your machine as a service in
      the cluster or expose it to a specific deployment as a local host.
```

corneliusweig (commented):
Ok, thanks for the explanation. Let's add this then. I have a minor wording (and text wrapping) suggestion, otherwise this is good to go!

Thanks for contributing!
/approve

---

Btw, you can shorten https://github.com/omrikiei/ktunnel/blob/6a5d8af7c979f877d63153932c0e032c410bea59/pkg/k8s/common.go#L25-L28 to
```go
```
Then you will also get additional auth provider when they are added.

omrikiei:
> https://github.com/omrikiei/ktunnel/blob/6a5d8af7c979f877d63153932c0e032c410bea59/pkg/k8s/common.go#L25-L28

I like the way you think

corneliusweig:
/lgtm
/approve
Thanks for contributing!

ahmetb:

PR #1376, opened 2021-07-07
add new plugin pvmigrate

laverya:
This should replace/supersede https://github.com/kubernetes-sigs/krew-index/pull/1373

### DIFF
--- a/plugins/pvmigrate.yaml
+++ b/plugins/pvmigrate.yaml
@@ -1,0 +1,57 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pvmigrate
+spec:
+  version: v0.1.0
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/replicatedhq/pvmigrate/releases/download/v0.1.0/pvmigrate_linux_amd64.tar.gz
+    sha256: 0c1b71ed2c012b49e65a2d1c507b797811b2007c8cb694190d22226fff9ee221
+    files:
+    - from: pvmigrate
+      to: .
+    - from: LICENSE
+      to: .
+    bin: pvmigrate
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/replicatedhq/pvmigrate/releases/download/v0.1.0/pvmigrate_darwin_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
@laverya thanks for your submission.

What's your intended use case with this plugin? 

This plugin seems to be an imperative management tool for PVs, and I am not sure if that's something people do frequently as I believe most Kubernetes mgmt happens declaratively through YAML manifests (applied imperatively via kubectl apply).

What's stopping users from doing kubectl edit (or editing YAMLs `storageClass` and applying it to the API via kubectl)? Is that currently not allowed?

Does this migration cause the data to be lost (deleting PV and recreating in a different storage class)? If so, this is probably worth noting in `caveats` section.

If you can quickly answer those and lay out the problem a bit more in detail and why you solved it this way, it would be helpful. I am trying to make sure this is something people actually need and would use.

laverya:
> What's your intended use case with this plugin?

Our internal use is to change clusters from using rook-ceph to longhorn.

> This plugin seems to be an imperative management tool for PVs, and I am not sure if that's something people do frequently as I believe most Kubernetes mgmt happens declaratively through YAML manifests (applied imperatively via kubectl apply).

This is something that people should not have to do frequently. However, it also can't be done easily with kubectl apply, as the involved fields are immutable.

> What's stopping users from doing kubectl edit (or editing YAMLs storageClass and applying it to the API via kubectl)? Is that currently not allowed?

Yes, the storage class for a PV is immutable.

> Does this migration cause the data to be lost (deleting PV and recreating in a different storage class)? If so, this is probably worth noting in caveats section.

This migration does not cause data loss. New PVs are created and the original data is copied over to them. The PVCs pointing to the original PVs are then deleted and new ones are created with the same names but referencing PVs on the destination storage class.

Is that the information you're looking for?

ahmetb:
Yeah that sounds great. The plugin is copying the data I assume.

laverya:
The plugin runs (and waits for) a pod for each PV to be migrated that actually does the copying, yep! I've put a step-by-step description in the repo: https://github.com/replicatedhq/pvmigrate/tree/v0.1.0#process

ahmetb:
Thank you. 🙏 Do you import the client-go Auth plugins?

laverya:
Not at present, but we should - I can put that in the next release! (Which will come approximately as soon as this gets merged, I think)

ahmetb:
/lgtm
/approve
Thanks!

PR #1446, opened 2021-08-05
Add support for the Instana self-hosted CLI

FrenchBen:
This adds an easier install path for those leveraging `krew` for plugin install - Docs: https://www.instana.com/docs/self_hosted_instana_k8s/instana_kubectl_plugin

### DIFF
--- a/plugins/instana.yaml
+++ b/plugins/instana.yaml
@@ -6,11 +6,11 @@ spec:
   shortDescription: Interact with instana
   description: |
     The official kubectl plugin for self-hosted instana.
-  version: v0.31.0
-  homepage: https://kubernetes.github.io/ingress-nginx/kubectl-plugin/
+  version: v205-0
+  homepage: https://www.instana.com/docs/self_hosted_instana_k8s/instana_kubectl_plugin/
   platforms:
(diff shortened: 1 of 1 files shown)


ahmetb:
@FrenchBen thanks for your submission.

It seems that Instana is an IBM project. Are you officially affiliated with it?

We expect the owners of the project to be involved with submission of their software on Krew so that they also can be onboard with updating versions here or installing hooks on their repo to automate releasing new versions.

FrenchBen:
@ahmetb fair enough... I am not, but thought it would be a handy contribution 😅

CodingFabian:
While we, Instana, consider this a great community contribution and addition to krew, we currently have no plans to support krew from our side.

corneliusweig:
@CodingFabian Would you be ok, if someone sends PRs to setup [krew-release-bot](https://github.com/rajatjindal/krew-release-bot) for the instana tools? It automatically sends version bumps for the manifest when a new github release is pushed. For example https://github.com/corneliusweig/ketall/blob/370e67f95fb2d10ca9816a850baf892695e4c7a2/.github/workflows/ci.yml#L83-L87.

I think without having reliably the latest version in krew, this will not work out well.

FrenchBen:
Closing, as it seems to not align with Instana practice.

PR #1566, opened 2021-09-22
added plugin advise-policy

darryk10:

### DIFF
--- a/plugins/advise-policy.yaml
+++ b/plugins/advise-policy.yaml
@@ -1,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: advise-policy
+spec:
+  version: {{ .TagName }}
+  homepage: https://github.com/sysdiglabs/kube-policy-advisor
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    {{addURIAndSha "https://github.com/sysdiglabs/kube-policy-advisor/releases/download/{{ .TagName }}/kube-policy-advisor_{{ .TagName }}_darwin_amd64.tar.gz" .TagName }}
+    bin: kubectl-advise-policy
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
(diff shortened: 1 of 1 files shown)


ahmetb:
The manifests in this repo should not be templatized. That template will be used by the release bot for the next versions of your plugin if you've set up the bot. 

For now, please hand fill those templatized values.

darryk10:
> The manifests in this repo should not be templatized. That template will be used by the release bot for the next versions of your plugin if you've set up the bot.
> 
> For now, please hand fill those templatized values.

fixed

ahmetb:
I took a look and overall looks good. The only point in my mind is that this tool offers functionality beyond "advising PSP/OPA policies":

> Available Commands:
>  * compare:     Compare k8s workload YAML files
>  * completion:  generate the autocompletion script for the specified shell
>  * convert:     Generate a PodSecurityPolicy from a single K8s Yaml file
>  * **inspect**:     Inspect a live K8s Environment to generate a PodSecurityPolicy

If you are ok with this potentially narrow naming of your plugin, we can move forward. But it'll be quite a hassle to change the plugin name later (upgrades won't happen on existing installations, it's essentially a new plugin at that point).

darryk10:
> I took a look and overall looks good. The only point in my mind is that this tool offers functionality beyond "advising PSP/OPA policies":
> 
> > Available Commands:
> > 
> > * compare:     Compare k8s workload YAML files
> > * completion:  generate the autocompletion script for the specified shell
> > * convert:     Generate a PodSecurityPolicy from a single K8s Yaml file
> > * **inspect**:     Inspect a live K8s Environment to generate a PodSecurityPolicy
> 
> If you are ok with this potentially narrow naming of your plugin, we can move forward. But it'll be quite a hassle to change the plugin name later (upgrades won't happen on existing installations, it's essentially a new plugin at that point).

@ahmetb thanks for your feedback. I'd like to keep this name and proceed due to the similarity with the old version of the tool. 
Thanks again

ahmetb:
I see. In this case, what do you plan to do with https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/advise-psp.yaml ? It would be better if you communicate that proactively as we sometimes struggle to keep 150+ plugins in mind.

ahmetb:
@darryk10 ping. #1577 just arrived before we could merge this.
@corneliusweig I am going on a vacation, handing this off to you. :D

PR #1618, opened 2021-10-11
Cm beta.6, a plugin for Red Hat Advanced Cluster Management

itdove:


itdove:
@ahmetb makes sense but the doc mentions
```
The possible values for os and arch come from the Go runtime. Run go tool dist list to see all possible platforms and architectures.
```
at https://krew.sigs.k8s.io/docs/developer-guide/plugin-manifest/

is the ppc64le the only one I listed in the manifest which is not supported or there is some others?
Removing the ppc64le...

itdove:
@ahmetb Retriger because missing the LICENSE

PR #1636, opened 2021-10-15
First push of clusteradm open-cluster-management

itdove:
Signed-off-by: Dominique Vernier <dvernier@redhat.com>

This is the first push of the clusteradm.
Clusteradm is a plugin to manage multi-clusters.

### DIFF
--- a/plugins/clusteradm.yaml
+++ b/plugins/clusteradm.yaml
@@ -1,0 +1,42 @@
+# Copyright Contributors to the Open Cluster Management project
+
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: clusteradm
+spec:
+  version: v0.1.0-alpha.6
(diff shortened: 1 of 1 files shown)


itdove:
@ahmetb Can you check with @corneliusweig what is his thought about this?

ahmetb:
@itdove I took another look again. Correct my understanding if it's incorrect: the clusteradm tool itself targets a different control plane that is not Kubernetes API to manage the clusters/registrations. In this regard it differs from kubectl or a kubectl plugin as kubectl typically targets Kubernetes API or the kubectl itself (e.g. kubeconfig files).

This might not be a good "kubectl plugin" fit based on this, just like how `clusterctl`, `minikube`, `kubeadm` or `kind` are also not kubectl plugins. They are standalone tools that do Kubernetes related tasks, but that doesn't fit into the "extending kubectl". We tend to reject this type of submissions, and it's ok to distribute as a stadalone tool (or in a custom krew index that you can easily host).

PR #1644, opened 2021-10-16
Add node allocations plugin

wshihadeh:
PLUGIN DEVELOPERS: @wshihadeh

Hi Team,

I would like to add my new plugin

Thanks
Al-waleed

### DIFF
--- a/plugins/node-allocations.yaml
+++ b/plugins/node-allocations.yaml
@@ -1,0 +1,55 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: node-allocations
+spec:
+  platforms:
+  - uri: https://github.com/wshihadeh/kubectl-node-allocations/releases/download/v0.0.2/kubectl-node-allocations-Darwin-x86_64.tar.gz
+    sha256: c2820760e002cad30357a756817b17d4592146935b313e7cb7c60d03ae6f33f2
+    bin: kubectl-node-allocations
+    files:
+    - from: kubectl-node-allocations
+      to: .
+    - from: LICENSE
+      to: .
+    selector:
+      matchLabels:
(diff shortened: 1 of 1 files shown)


ahmetb:
Is the output below (showing how many nodes are there on each node) all this plugin does?

```
kubectl node-allocations --all-namespaces
NODE                                               PODS-COUNT
gk3-autopilot-cluster-nap-1nafw4xe-26a7899f-k0ad   11
gk3-autopilot-cluster-default-pool-eeb737f6-uy2z   14
```

If so, I am afraid there's not enough utility to warrant adding it here.
Furthermore, I think the name is not great –node allocation in Kubernetes context typically refer to how the resources like CPU/memory are reserved to the system vs the dynamic workloads. (e.g. see the definition [here](https://unofficial-kubernetes.readthedocs.io/en/latest/tasks/administer-cluster/reserve-compute-resources/)).

wshihadeh:
Yes that is the first command currently implemented (show pods count by Node), However I am going to add more commands and features

corneliusweig:
IMO `statistics` is far too generic, so that's a no-go for me.

My impression is that the plugin is still heavy in the conception phase. As of right now, the added functionality is low, so I'm reluctant to merge already. I'd encourage you to add more functionality and then let's reconsider. Please be aware that it is not great to have too much overlap with existing plugins, so please have a look at the existing plugins https://krew.sigs.k8s.io/plugins/.

/kind gray-area
/hold

ahmetb:
+1 for revisiting later. 

Until then, you can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index. 

/close

PR #1647, opened 2021-10-18
Add kubectl-counts

wshihadeh:
Hi , 
I would like to add my new plugin kubectl-counts

### DIFF
--- a/plugins/counts.yaml
+++ b/plugins/counts.yaml
@@ -1,0 +1,68 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: counts
+spec:
+  platforms:
+  - uri: https://github.com/wshihadeh/kubectl-counts/releases/download/v0.0.1/kubectl-counts-Darwin-x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


chriskim06:
hi @wshihadeh, thanks for your plugin submission!

as far as this one goes, I think the name might be a little too generic (you can refer to [the naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/) on what acceptable plugin names for krew are). also I think this might fall into that category of plugins that don't provide enough functionality. it seems like some of the flags supported by this plugin are a wrapper over normal `kubectl get` commands piped to `wc -l` (although some of these would probably need to use `--field-selector` which can definitely get a little messy sometimes but is functionality that is already baked into kubectl).

/cc @ahmetb 
/cc @corneliusweig 
(I'd like to get your guys' opinions on this as well)

corneliusweig:
I suspect that your plugin will not be used much. It takes effort to discover and become familiar with a plugin. If the same task can be achieved with a bash oneliner, there is simply no need to even look for a solution to this problem, because it is trivial to compose on-the-fly.

As a way out, have you thought about contributing your feature to `kubectl-grep`. If I understand correctly, it would fit into that plugin. So you could make your feature public and available via krew-index in an indirect way.

ahmetb:
Thanks for the discussion here.
/close

PR #1753, opened 2021-11-16
new version v0.6.1 of viewnode

grafvonb:


plugins/viewnode.yaml
```diff
@@ -16,20 +16,20 @@ spec:
...
+      sha256: f08c158d73a01a4b84a8552a5ce80530cb580194e500b087bc2aa4503188ffff
       bin: viewnode
     - selector:
         matchLabels:
           os: linux
           arch: amd64
-      uri: https://github.com/NTTDATA-EMEA/viewnode/releases/download/v0.6.0/viewnode_0.6.0_linux_amd64.tar.gz
-      sha256: b8c986ff3af9402c679a0a828febd8338223d142c8da3b376019b1146d88c023
+      uri: https://github.com/NTTDATA-DACH/viewnode/releases/download/v0.6.1/viewnode_0.6.1_linux_amd64.tar.gz
+      sha256: cae163c884082dcd4ed9c10b9289a0aa760a051a4b4d67420e9865757c8470b4
```
ahmetb, line 27: why are there two different repos here? this looks suspicious from a trust perspective that existing repo is not renamed/transferred but rather a copy of it is created and a new release is made there.

grafvonb: The first repository I used was wrong (an unofficial one) and will be deleted according to our CTO office in Germany. There was some misunderstanding in our organization. The new repo is a part of our NTT DATA EMEA official organization account at GitHub as you can see here https://github.com/enterprises/emea-nttdata. I can provide a statement from our CTO for it if necessary.

ahmetb: > because I was not sure if after that the link from our krew-manifest will work if pointing at the old address.

it would work :) github sets up redirects for transferred repos.

PR #1865, opened 2021-12-23
implement go mod cache

ahmetb:
since we do go-get, 15s (about 50%) of the time of each krew-index PR
is spent there. this solves that.

/cc @chriskim06

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,18 +4,25 @@ on: pull_request
 
 jobs:
   build:
-    name: Build and Tags
+    name: Validate plugin manifests
     runs-on: ubuntu-latest
     env:
       KREW_VERSION: v0.4.0
       GO111MODULE: on
     steps:
 
-    - name: Set up Go 1.14
+    - name: Setup Go
       uses: actions/setup-go@v1
       with:
         go-version: 1.14
       id: go
+    - name: Go module cache
+      uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
 
     - name: Check out PR branch
       uses: actions/checkout@v2


.github/workflows/ci.yml
```diff
@@ -4,18 +4,25 @@ on: pull_request
...
-    - name: Set up Go 1.14
+    - name: Setup Go
       uses: actions/setup-go@v1
       with:
         go-version: 1.14
       id: go
+    - name: Go module cache
+      uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
```
chriskim06, line 22: I see another path `~/.cache/go-build` [in the example](https://github.com/actions/cache/blob/main/examples.md#linux), do we not need that one?

ahmetb: Valid point. I've actually done it in some other projects (e.g. kubectx repo builds about 20 binaries) so here the build time is not as prevalent (since it's a linux build on linux), we mostly spend the time in retrieving the dependencies vs build time. We can try adding that later on? @chriskim06

ahmetb: here's a better snippet than the provided example, but it seemed to complicate things too much just to have a little bit of caching:

```
    - id: go-cache-paths
      run: |
        echo "::set-output name=go-build::$(go env GOCACHE)"
        echo "::set-output name=go-mod::$(go env GOMODCACHE)"
    - name: Go Build Cache
      uses: actions/cache@v2
      with:
        path: ${{ steps.go-cache-paths.outputs.go-build }}
        key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
    - name: Go Mod Cache
      uses: actions/cache@v2
      with:
        path: ${{ steps.go-cache-paths.outputs.go-mod }}
        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
```

chriskim06: sgtm, so this is just for the downloading of the validate-krew-manifest tool right?

ahmetb: yeah

PR #1877, opened 2022-01-04
Example plugin update

seredot:
The new version comes with Arm64 support for Darwin.


ahmetb:

-----

>file `plugins/example.yaml` is not a straightforward version bump: change on line `10` does not seem like a version bump: [`+    - selector:`]
/kind plugin-update

PR #1923, opened 2022-01-20
Add kubectl-vpa-recommendation plugin to Krew index

wI2L:
Hello

I would like to propose a new plugin I've been working on recently: https://github.com/wI2L/kubectl-commitment
The plugin allows users to compare the target recommendations of `VerticalPodAutoscaler` resources with the actual resource requests of the pods managed by the targeted controller.

Release automation with `rajatjindal/krew-release-bot@v0.0.40` and documentation will be added/updated, if the plugin gets added to the index. The `.krew.yaml` template file is already versioned in the repository: https://github.com/wI2L/kubectl-commitment/blob/master/.krew.yaml

The plugin lacks unit tests/coverage, they will be added over time as the plugin mature in features.

Regards,

William

### DIFF
--- a/plugins/vpa-recommendation.yaml
+++ b/plugins/vpa-recommendation.yaml
@@ -1,0 +1,60 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: vpa-recommendation
+spec:
+  version: v0.2.0
+  homepage: https://github.com/wI2L/kubectl-vpa-recommendation
+  shortDescription: Compare VPA recommendations to actual resources requests
+  description: |
+    This plugin prints a table that show the percentage increase/decrease
+    of the selected VerticalPodAutoscaler recommendations compared to the
+    targeted controller's pod resource requests.
+  caveats: |
+    The plugin recognizes only some well-known controllers such as:
+      - CronJob
+      - DaemonSet
+      - Deployment
+      - Job
+      - ReplicaSet
+      - ReplicationController
+      - StatefulSet
+
+    The autoscaling.k8s.io/v1 API is required for the plugin to work:
+    https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/wI2L/kubectl-vpa-recommendation/releases/download/v0.2.0/kubectl-vpa-recommendation_v0.2.0_darwin_amd64.tar.gz
+    sha256: 7753a8098837055b32e0c033e2d4910e17b1a414d46206a62ecc0a74ce0cb6f2
+    bin: kubectl-vpa-recommendation
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: arm64
+    uri: https://github.com/wI2L/kubectl-vpa-recommendation/releases/download/v0.2.0/kubectl-vpa-recommendation_v0.2.0_darwin_arm64.tar.gz
+    sha256: 3f0bc3a357b7154baded836f17bd0935c27f8f846f1e7a452d8ed6ea661a8647
+    bin: kubectl-vpa-recommendation
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/wI2L/kubectl-vpa-recommendation/releases/download/v0.2.0/kubectl-vpa-recommendation_v0.2.0_linux_amd64.tar.gz
+    sha256: bc566300a38da5939f69f5beca35aa6ffac82eb30d667766b1ef2844fbdf4258
+    bin: kubectl-vpa-recommendation
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm64
+    uri: https://github.com/wI2L/kubectl-vpa-recommendation/releases/download/v0.2.0/kubectl-vpa-recommendation_v0.2.0_linux_arm64.tar.gz
+    sha256: dfcb68cce61d19744b9c0a0b61debd52ad4c85d0f7602b229cc42b21ded92103
+    bin: kubectl-vpa-recommendation
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/wI2L/kubectl-vpa-recommendation/releases/download/v0.2.0/kubectl-vpa-recommendation_v0.2.0_windows_amd64.zip
+    sha256: 711e98d062a4bda4c8827e9b441480a71f419cca7753ecbe1128d03c47e76bec
+    bin: kubectl-vpa-recommendation.exe


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
This seems pretty excellent already. The name "commitment" doesn't really signal to me that this is designed to measure VPA status, but if you feel comfortable with it that's fine as it's specific enough –except the only other use case I can imagine this name could be used is probably for a plugin that measures overcommit/undercommit situation (and this would be a perfect name probably for that plugin instead).
I bring this up since there's still time to change this, and the name didn't really ring the "VPA" bell in my mind.
/approve

plugins/commitment.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: commitment
+spec:
+  version: v0.1.1
+  homepage: https://github.com/wI2L/kubectl-commitment
+  shortDescription: Compare VerticalPodAutoscaler recommendations to actual resources requests
```
ahmetb, line 8: this probably won't fit `kubectl krew list` output, it'll be truncated at 50 chars.

wI2L: Got it. I wrapped the `description` field to 80 chas as indicatd in the developper guide, but missed that. Will fix it.

wI2L:
I am open to suggestions regarding the name of the plugin, if you have any.

> –except the only other use case I can imagine this name could be used is probably for a plugin that measures overcommit/undercommit situation

To me, that's what the plugin is doing, but I may be wrong and mistaking it with another thing. Could you clarify what you have in mind when you mention "undercommit/overcommit" ?

ahmetb:
> Is "commitment, under/over" generally used to talk about limits vs. capacity, instead of requests ?

Yeah, I think typically when people talk about "commitment" in the infrastructure space (compute, network etc), we talk about what's provisioned and what's assigned vs what's available. e.g. memory, CPU, network etc. An alternative to this is "oversubscription" or "undersubscription".

Perhaps the name is okay and I'm exaggerating the potential of someone developing a plugin that shows what's assigned vs what's available in terms of resources like memory/CPU. But Kubernetes doesn't really let us overcommit I guess, so maybe that won't happen.

wI2L:
@ahmetb
Looks like `kubectl describe node` already show the informations regarding allocated resources (and overcommitment when limits > 100%). I have mixed feelings regarding the naming, when it comes to VPA, it seems to me that it's also an over-commitment to go beyond the recommendations (in the sens that this is an excess).
I have thought of alternatives terns, such as `excess`, `reservation` (since it's related to resource requests).
Should we get another opinion ? @corneliusweig maybe ?
Again, not against a renaming of the plugin if we find a name that is clearer.

chriskim06:
`commitment` makes me think of node over commitment like you both discussed earlier. I'm not 100% opposed to that name but I think it would be more clear if `vpa` was included in the name (like `vpa-commitment` or something).

wI2L:
@chriskim06 @ahmetb 
What do you think of `recommendation(s)`, as name ? It's closely relatd to the VPA concept. Ideally I'd like to kep a single-word name, but otherwise `vpa-commitment` sounds good.

plugins/commitment.yaml
```diff
@@ -0,0 +1,57 @@
...
+  description: |
+    This plugin prints a table that show the percentage increase/decrease
+    of the selected VerticalPodAutoscaler recommendations compared to the
+    targeted controller's pod resource requests.
+  caveats: |
+    The plugin recognize only some well-known controllers such as:
```
chriskim06, line 14: small typo `recognize` -> `recognizes`

also I think maybe its worth mentioning that the `VerticalPodAutoscaler` needs to be installed in the cluster, especially depending on if the name will contain `vpa` or not

chriskim06:
> @chriskim06 @ahmetb
> What do you think of `recommendation(s)`, as name ? It's closely relatd to the VPA concept. Ideally I'd like to kep a single-word name, but otherwise `vpa-commitment` sounds good.

actually `vpa-recommendation` sounds even better to me, still can't think of a one word plugin name though. but like I said before, if everyone if fine with `commitment` (or `recommendation`) I'm also ok with it.

plugins/commitment.yaml
wI2L: Yep, make sense. Will push a fix including @ahmetb recommendation to wrap the description to 50 chars.

ahmetb:
Recommendations is way broader IMO and can actually be developed as many different plugins.

ahmetb:
Took a closer look at this plugin. I think `vpa-diff` or `vpa-status` are also pretty good alternatives. As a matter of fact, both you and [this plugin](https://github.com/ninlil/kubectl-vpa) (not in krew) use "diff" terminology commonly. That could be something to think about since your plugin is still new.

I don't see "commitment" as a word within the VPA nomenclature anywhere and that's why the word still does not ring a bell for me in this context.

wI2L:
> I don't see "commitment" as a word within the VPA nomenclature anywhere and that's why the word still does not ring a bell for me in this context.

"commitment" is often used as terminology regarding limits, but that could apply to requests as well, since the requests "commited" are taken into account by the scheduler to schedule pods onto nods (a.k.a reservation). Since the VPA computes targets that indicate whether the resources would be wasted or aren't enough, that's why I went with that from the beginning.

I think we can agree on the fact that the plugin would be better named with a `vpa-` prefix. What completes it is the hard part.

> As a matter of fact, both you and this plugin (not in krew) use "diff" terminology commonly. That could be something to think about since your plugin is still new.

I was aware of this plugin before I developped this one. I wasn't satisfied with the approach of binding pods to VPA by name, rather than fetching the target controller referenced by the VPA. What do you mean by "That could be something to think about since your plugin is still new." ? Should we leave the "vpa-diff" name open if this plugin's owner wish to add his plugin to krew ?

To stick closely to the VPA nomenclature, we could go with the name `vpa-target(s), instead of `vpa-commitment` or `vpa-recommendations`. `vpa-diff` is also ok, but that's not taking into account that a VPA a different "kind/type" of recommendations (target, upper/lower bound, uncapped target).

wI2L:
I gave it a thought again, `vpa-targets` is too specific to a type of recommendation provided by a VPA. Since this plugin will be centered on computing the differences from VPA recommendations vs. actual usage, I think `vpa-recommendation` (singular), would fit nicely. @ahmetb Could you please validate it, before I go on and rename the plugin itlsef, release new binaries and update the PR ? Thanks

wI2L:
Thank you both for the feedback and review. I'll notify you once the plugin is renamed and the PR amended to include name change and review fixes.

wI2L:
@ahmetb I just released a new version of the plugin with the new chosed name, and amended the PR to update the krew manifest file. I think it's ready for another review. Thanks.

ahmetb:
/lgtm
/approve
Thanks!

wI2L:
Yep, I didn't committed the new step in the `release` pipeline yet, but will do.

PR #1933, opened 2022-01-24
feat: add m1 support (darwin/arm64) for deprecations (kubepug)

patrickjahns:
# Motivation

Installation for https://github.com/rikatz/kubepug currently fails on m1 macbooks
```
kubectl krew install deprecations
Updated the local copy of plugin index.
Installing plugin: deprecations
W0124 17:06:28.276080   99305 install.go:164] failed to install plugin "deprecations": plugin "deprecations" does not offer installation for this platform
F0124 17:06:28.276136   99305 root.go:79] failed to install some plugins: [deprecations]: plugin "deprecations" does not offer installation for this platform
```

This adds the m1 compatible binary links


chriskim06:
I think you should file an issue/PR in kubepug instead of updating it here. that repo has the krew release automation setup so if this is updated the next routine upgrade for them won't automatically pass the PR checks when the release bot submits a PR

patrickjahns:
@chriskim06 
Thanks for pointing this out - will check with kubepug as well.

PR #1999, opened 2022-02-15
Add pod-node-matrix plugin

ardaguclu:
This PR adds `pod-node-matrix` plugin into krew-index.

`pod-node-matrix` shows [pods x node] matrix in terms of pod statuses in colorful table view.
This plugin can clearly indicate that if there is a general node problem,
or suggest that node has no problem and instead deployment, service, etc. of this pod can be checked.

Project repo: https://github.com/ardaguclu/kubectl-pod-node-matrix

### DIFF
--- a/plugins/pod-node-matrix.yaml
+++ b/plugins/pod-node-matrix.yaml
@@ -1,0 +1,42 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pod-node-matrix
+spec:
+  version: v0.0.2
+  homepage: https://github.com/ardaguclu/kubectl-pod-node-matrix
+  shortDescription: Show [pod x node] matrix in terms of pod statuses
+  description: |
+    This plugin shows [pod x node] matrix in terms of pod statuses 
+    with suitable colors to mitigate troubleshooting effort.
+  caveats: |
+    * Due to the console width limitation, it does not support huge number of nodes.
+  platforms:
+    - selector:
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. However this plugin looks a bit under-baked in design to me. kubectl get pods -o=wide already shows statuses and node name (you can also customize columns) and this plugin won't scale well to double digit nodes and further in terms of output rendering. Am I missing any more features?

If this is the case, I recommend refining the experience a bit more since as it stands today, it doesn't create enough value-add to kubectl users. 

/kind gray-area

ardaguclu:
> Thanks for the submission. However this plugin looks a bit under-baked in design to me. kubectl get pods -o=wide already shows statuses and node name (you can also customize columns) and this plugin won't scale well to double digit nodes and further in terms of output rendering. Am I missing any more features?
> 
> If this is the case, I recommend refining the experience a bit more since as it stands today, it doesn't create enough value-add to kubectl users.
> 
> /kind gray-area

Thanks for review @ahmetb.

TBH, I wasn't sure either this plugin adds **real** value for general use case when I was opening this PR. Because all your concerns are totally valid(limitation of node count, most people can just use `-owide`, etc.). 

This plugin covers very specific use case and in the future if I add something appealing to users, I'll re-open this PR. Thereby, I'm closing now.

Thanks again.

/close

ahmetb:

PR #2145, opened 2022-04-05
Bump cert-manager plugin versions for 1.8

SgtCoDFish:
This process isn't automated because the automation that's provided for the krew-index isn't a great fit for the cert-manager release process.

Even if it was automated, this PR would require manual review in any case since we changed our repo name, updated Windows to use zip and added the `.exe` extension for the Windows binary.

If this needs any changes please let me know!

### DIFF
--- a/plugins/cert-manager.yaml
+++ b/plugins/cert-manager.yaml
@@ -3,54 +3,53 @@ kind: Plugin
 metadata:
   name: cert-manager
 spec:
-  version: v1.7.0
-  homepage: https://github.com/jetstack/cert-manager
+  version: v1.8.0
+  homepage: https://github.com/cert-manager/cert-manager
   shortDescription: Manage cert-manager resources inside your cluster
   description: |
-    The official plugin accompanying cert-manger, a Kubernetes add-on to
-    automate the management and issuance of TLS certificates. Allows for
-    direct interaction with cert-manager resources e.g. manual renewal of
-    Certificate resources.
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

>file `plugins/cert-manager.yaml` is not a straightforward version bump: change on line `4` does not seem like a version bump: [`-  homepage: https://github.com/jetstack/cert-manager`]
/kind plugin-update

ahmetb:
@SgtCoDFish hey, do you mind elaborating what makes it not a good fit to cert-manager release process?

ahmetb:
/lgtm
/approve
Thanks!

SgtCoDFish:
> @SgtCoDFish hey, do you mind elaborating what makes it not a good fit to cert-manager release process?

@ahmetb sure! It's not a problem with the automation per-se, it's that our release process was split across two repos a while ago and that makes it a pain to trigger a full release automatically. We're moving away from that split and simplifying things, but it's a slow process juggling that with other things going on.

We also don't really use GitHub actions (for various reasons) and so there was a bit of a learning curve when I tried to integrate the krew automation a while back, which made it all a bit slow and sluggish.

Our release process is definitely overcomplicated, but with that complication comes power and there's no reason we couldn't have the "publish" step of our releases automatically create a PR here. We have automatic PRs for publishing our helm charts, for example. We just haven't got around to doing the automation bit for krew!

PR #2185, opened 2022-04-17
Add diag plugin

yuval-k:
A plugin to help diagnose problems I encountered when debugging istio.

The main feature is that it allows redirecting traffic from a pod to a laptop (which I use to get envoy to talk with my pilot running on my laptop). There are other tools as well (multi pod log output, and debug shell)

I named this `diag` so it is short to type, but I'm open to suggestions here.
If this will be accepted I will add release automation.

### DIFF
--- a/plugins/diag.yaml
+++ b/plugins/diag.yaml
@@ -22,7 +22,7 @@ spec:
     sha256: b687aaea96a67021b89cfb16d1c07226a7bbb180bc6261fea7a2da80b6a77958
     bin: kubectl-diag
     files:
-    - from: kubectl-diag-darwin-arm64
+    - from: kubectl-diag
       to: kubectl-diag
     - from: LICENSE
       to: .
--- a/plugins/diag.yaml
+++ b/plugins/diag.yaml
@@ -3,11 +3,11 @@ kind: Plugin
 metadata:
   name: diag
 spec:
-  version: v0.0.5
+  version: v0.0.9
   homepage: https://github.com/solo-io/kdiag
   platforms:
-  - uri: https://github.com/solo-io/kdiag/releases/download/v0.0.5/kubectl-diag_v0.0.5_darwin_amd64.tar.gz
-    sha256: a19e2c6e9449fde670d13ab8bda7c0444f4539b83d9faa6020f14cf29dbdafb2
+  - uri: https://github.com/solo-io/kdiag/releases/download/v0.0.9/kubectl-diag_v0.0.9_darwin_amd64.tar.gz
+    sha256: fd38a1a2ace63bcf688879e32d94d0b6a556094b9733477b71c9ac1d7ce377b4
     bin: kubectl-diag
     files:
     - from: kubectl-diag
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
I think an alternative name would be better than assigning `diag` to a plugin only related to Istio. `istio-diag` is appropriate however.

yuval-k:
> I think an alternative name would be better than assigning `diag` to a plugin only related to Istio. `istio-diag` is appropriate however.

while the examples are istio focused, I use this to debug also other client/server systems, where i want to debug how the server responds to a specific client. (i didn't have examples for these as they are internal systems). there is nothing istio specific in the plugin itself.

I plan to add another tool that lets you start a shell in distroless/"scratch" containers.

Perhaps `srv-diag` (i.e. server diagnostics) ?

ahmetb:
I think I see the usefulness, but I am also trying to understand how does this relate to ephemeral containers (and also the builtin `kubectl debug` command). Can you elaborate further?
At any rate I think the shortDescription/description should be clear enough that users don't find themselves asking these questions.

yuval-k:
make a lot of sense! thanks for the feedback.

Ephemeral containers are used to attach a container to a running pod. That container has a component that helps sets up the traffic redirection. specifically this means setting up iptables rules, and proxying connections.

The reason i use ephemeral container and not just exec to the pod is two fold: 1. i can ask an ephemeral container with the capabilities i need and 2. i have all the tools needed (i.e. iptables binary).

I also plan to  add to the ephemeral container a static variant of standlone ash (the busybox shell), and a tool to start in the pod's namespaces, so that you can have a usable shell in scratch containers.

ahmetb:
OK, got it. So you're using ephemeral containers and changing some things in the net ns etc to route traffic.
I'm giving this feedback mostly because your plugin seems to be new:

My observation is that you have 2 sub-commands:
- redir: makes perfect sense, unique functionality. should be a plugin.
- shell: seems to be duplicating functionality with the builtin `kubectl debug --image=` cmd which also uses ephemeral containers. If that is the case, I recommend not doing that (unless you provide an obvious differentiator)

So I recommend instead of trying to build a Swiss army knife, try to build more focused tools. The lack of focus is clear from the current `shortDescription: Diagnostics tools for kubernetes developers.`. (As a user, I can't tell what you're diagnosing from this description. It can be dozens of different things.)

So I'd love to have you think about whether you really want to turn this into a Swiss army knife and use name `diag` which will impede your discoverability OR scope down to `redir` and get that plugin name.

yuval-k:
The main differentiator i plan to add is that the shell will work for scratch containers. I.e. it will feel like  `kubectl shell pod` will behave like `kubectl exec -ti pod -- /bin/bash` but will work even if the pod doesn't have bash (using ephemeral containers and `execveat`).

If I'll complete that feature first would you be ok having both in the plugin? or would you still prefer separate ones? 
if you still prefer separate ones, I'll restructure the code so that `redir` is an independent krew plugin (while keeping the rest of the functionality in a separate binary not published in krew-index).

ahmetb:
> The main differentiator i plan to add is that the shell will work for scratch containers. I.e. `kubectl shell pod` will behave like `kubectl exec -ti pod -- /bin/bash` but will work even if the pod doesn't have bash (using ephemeral containers and `execveat`).

I am still not understanding how is it different than "kubectl debug" because `kubectl debug --image=bash` does what you described to a scratch container?



> If I'll complete that feature first would you be ok having both in the plugin? or would you still prefer separate ones?

I am actually ok with the name diag as it's specific enough that others won't come asking for the same name. I'm more worried about your plugin's discoverability among the catalog of ~200 plugins. Neither the name nor the description clearly explains what the plugin does.

yuval-k:
Hi @ahmetb , finally had some time this weekend and I updated the repo and the description.
Hopefully it's more clear now - let me know what you think!

The common thread of the tools in this package is that I found them useful when debugging istio.

PR #2261, opened 2022-05-17
Add kopy plugin

giorgosdi:

### DIFF
--- a/plugins/kopy.yaml
+++ b/plugins/kopy.yaml
@@ -1,0 +1,46 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kopy
+spec:
+  version: v0.0.1
+  homepage: https://github.com/giorgosdi/kubectl-kopy
+  shortDescription: Copy a Kubernetes object between namespaces
+  Description: |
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for your submission. Since you've very recently authored this plugin, I'm hoping this feedback makes sense. We typically try to encourage composability of the plugins and kubectl. 

As you also noted in your repo, is your plugin basically a shorthand alias for:

    kubectl get KIND/NAME -n SRC --context SRC | kubectl neat | kubectl apply -f- -n DST --context DST

If so, I think we should probably not accept this plugin since it seems "copying" is a special case that can be achieved by  composing existing plugins with kubectl builtin commands. (For the record, `eksporter` plugin also exists and does nearly the same thing as `neat` plugin already.)

giorgosdi:
Hey Ahmet, thanks for the quick reply. Yeah you could say that it also works as an alias. The main reasoning behind this is to avoid these long piped command (i.e you need 3 of the to accomplish what kopy is doing with 1).

If you still believe that its derivative of these two plugins, then im happy for you to close the PR 😄

ahmetb:
/close 
since it's not a fit to the plugin model but more like an alias/bash func.

PR #2263, opened 2022-05-17
Add crd-extractor to plugin

hadar-co:

### DIFF
--- a/plugins/crd-extractor.yaml
+++ b/plugins/crd-extractor.yaml
@@ -1,0 +1,27 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: crd-extractor
+spec:
+  version: v0.0.1
+  homepage: https://github.com/datreeio/CRDs-catalog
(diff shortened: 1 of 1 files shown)


ahmetb:
I'm inclined to reject this because: 

1. This does not warrant a kubectl plugin. The operation at hand here is converting CRDs to jsonschema, which is not a day to day kubernetes operation that makes it a "kubectl" plugin, as we expect plugins to be enhancing kubectl experience. This is more like a small utility that targets CRD developers (e.g. kubebuilder tools are not kubectl plugins), so its audience is not general kubectl audience.

1. The program downloads and executes other programs on user's behalf, which is a big no-no for us.
https://github.com/datreeio/CRDs-catalog/blob/8fbff760d9615629538c73b65c83907843874f24/crd-extractor/kubectl-crd-extractor#L23-L32. The program doesn't add notable value around the actually executed program. It feels like this just needs to be a gist/script somewhere, and not a kubectl plugin.

1. There are actual tools like [kcp crd-puller](https://github.com/kcp-dev/kcp/tree/main/cmd/crd-puller) that might fit this name better.

PR #2294, opened 2022-05-27
add interact-sh plugin

ya-makariy:


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
The idea of the plugin is lgtm. 

Unrelated comment about the plugin itself: I highly recommend you use something like `fzf` to shell out to. :) That's what I do in kubectx and it saves a ton of work. The current interactive picker you have is actually a lot slower than typing out the names by hand or using kubectl shell completion. I found it very difficult to use and scroll down the list etc.

ahmetb:
Oh and I think you have a bug that doesn't parse multiple containers in a Pod onto different lines, so I ended up choosing a line with both container names in one line and as a result could not get this to work.

ya-makariy:
Thanks for the quick response. I will change the scroll mechanism to speed up the whole process. And then I'll be back)

yep, it was a bug)

ya-makariy:
Hey @ahmetb, could you review the plugin again?

PR #2303, opened 2022-05-31
Make cache work on every pull request

JohnTitor:
Because each PRs has a fresh build, caching on GHA rarely happens. And even if it happens, it doesn't show a significant result, see https://github.com/kubernetes-sigs/krew-index/runs/6546909595?check_suite_focus=true and https://github.com/kubernetes-sigs/krew-index/runs/6550071463?check_suite_focus=true for example. One of the reasons is our build time is quite small.
Thus I think we could remove caching modules to simplify. While doing it, this also updates some actions to the latest ones.

**UPDATE**: This PR now enables cache on every pull request as requested in https://github.com/kubernetes-sigs/krew-index/pull/2303#issuecomment-1147644947.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       id: go
 
     - name: Check out PR branch
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0  # disable shallow clones
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,17 +12,10 @@ jobs:
     steps:
 
     - name: Setup Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v3
       with:
-        go-version: 1.14
+        go-version: '1.14'
       id: go
-    - name: Go module cache
-      uses: actions/cache@v2
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
 
     - name: Check out PR branch
       uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: krew-index CI
 
-on: pull_request
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


ahmetb:
I am not understanding. Why's caching worse here?

As far as I know this was reducing test times quite a bit.

That's why I added it not too long ago.

Our cache key is the go.sum file which doesn't change often, so we should be frequently having cache hits.

JohnTitor:
> Our cache key is the go.sum file which doesn't change often, so we should be frequently having cache hits.

I explained it on OP: "Because each PRs has a fresh build, caching on GHA rarely happens."

So, let's see the recent builds:
- https://github.com/kubernetes-sigs/krew-index/runs/6623836837?check_suite_focus=true
- https://github.com/kubernetes-sigs/krew-index/runs/6630132693?check_suite_focus=true
- https://github.com/kubernetes-sigs/krew-index/runs/6669498944?check_suite_focus=true
- https://github.com/kubernetes-sigs/krew-index/runs/6652493159?check_suite_focus=true
- https://github.com/kubernetes-sigs/krew-index/runs/6643796904?check_suite_focus=true

All of them didn't use the cache feature.

Exceptions are:
- https://github.com/kubernetes-sigs/krew-index/runs/6546909595?check_suite_focus=true
- https://github.com/kubernetes-sigs/krew-index/runs/6550071463?check_suite_focus=true

So, if someone opens a PR and pushes changes several times, we hit cache. But I'm not sure if it's worth doing so with introducing complexity as it's not "frequently" to me and it doesn't show any significant results. Both builds, cache ones and no-cache ones, finished in 30-40s (see the above builds).
Alternatively, if you strongly prefer, e.g. we could use [setup-go builtin cache feature](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) instead of using actions/cache as it's less complex than actions/cache.

ahmetb:
Thanks for pointing out. It might be better to understand why it's not working as intended and see if there's a workaround (such as your recommendation). Basically we spend too much time doing a reproducible "go install", which should be cacheable somehow.

ahmetb:
Yeah, making it work intuitively would be preferred. :)

.github/workflows/ci.yml
```diff
@@ -12,24 +16,26 @@ jobs:
...
+        go-version: '1.14'
       id: go
+
     - name: Go module cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-go-${{ env.KREW_VERSION }}
```
ahmetb, line 28: Why do we need to change this again?

JohnTitor: See https://github.com/kubernetes-sigs/krew-index/runs/6550071463?check_suite_focus=true#step:3:4 for example, you can see `key: Linux-go-`, i.e. `${{ hashFiles('**/go.sum') }}` does nothing.

ahmetb: I think that'd be good to figure out why. We copied it from https://github.com/actions/cache/blob/main/examples.md so I expect it to work.

JohnTitor: Because this repo doesn't have go.mod/go.sum files. That example assumes you have a general Go module/package structure.

ahmetb: D'uh I am not sure how that got past the code reviews initially. Thanks. You picked the right key here in that case. I hope we don't have the problem in Krew main repo but caching matters here more, definitely.

PR #2317, opened 2022-06-05
Add clean plugin

shardulsrivastava:
Add clean plugin.

### DIFF
--- a/plugins/clean.yaml
+++ b/plugins/clean.yaml
@@ -1,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  # 'name' must match the filename of the manifest. The name defines how
+  # the plugin is invoked, for example: `kubectl restart`
+  name: clean
+spec:
+  # 'version' is a valid semantic version string (see semver.org). Note the prefix 'v' is required.
+  version: "v0.0.1"
+  # 'homepage' usually links to the GitHub repository of the plugin
+  homepage: https://github.com/shardulsrivastava/kubectl-clean
+  # 'shortDescription' explains what the plugin does in only a few words
+  shortDescription: "Cleans up orphaned pods in the cluster."
+  description: |
+    Cleans up orphaned pods in the cluster.   
+  # 'platforms' specify installation methods for various platforms (os/arch)
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: "os"
+        operator: "In"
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
The name is too generic. Please refer to naming guidelines.

ahmetb:
We also have the rm-standalone-pods plugin, please quantify (as a comment as well as in your plugin description field) what orphaned means.

shardulsrivastava:
@ahmetb thanks for your review, I will change the plugin name to be specific. This plugin is different from `m-standalone-pods` that you are mentioning.  If you're using `Spot instances` in EKS or `Pre-emptible instances` in GKE, when those instances go away, there will be a lot of pods (or just their reference) left, resulting in pods in `Evicted/NodeShutdown/Terminated` states or orphaned pods from jobs in `Completed`. This plugin helps to clean them all up.

ahmetb:
I am not sure if that makes a lot of sense based on what you said. That sounds like the job of the respective workload controller to clean up a Job pod that's completed or a Pod whose Node went away. 

I don't know what this would be a manual task. Are you positive on how k8s workload controllers are supposed to work in this regard?

shardulsrivastava:
@ahmetb here is what the cluster looks like in GKE, whenever the pre-emptible node gets shutdown, these pods don't go away automatically  : 

image

ahmetb:
That sounds like a problem with whatever controller is scheduling those pods (kube-controller-manager seems to be the case here). There shouldn't be a plugin needed to clean up Pods that are managed by a controller.

shardulsrivastava:
@ahmetb agreed this should be handled by the `kube-controller-manager`, however that's not the case in AWS and GKE, that results into periodic checks for these pods and manual cleanup, this plugin attempts to automate that.

ahmetb:
I am not sure I agree. Plugins are not automation tools, they're executed manually by kubectl users, mostly interactively. That's not automation. 

What you're running into might be a bug and accepting a plugin for this sort of situation doesn't make a lot of sense. I recommend you follow up about this somewhere instead of publishing a long term plugin here. 

You can still distribute your plugin via Krew using custom indexes if you want: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index. 

/close

PR #2343, opened 2022-06-10
integrate the ouctl plugin for deploying OpenUnison

mlbiam:
1.  Validated the install through manual kubectl krew install --manifest...
2. Updated to match best practices docs


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

plugins/ouctl.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ouctl
+spec:
+  version: "v0.0.1"
+  homepage: https://github.com/tremolosecurity/openunison-control
+  shortDescription: "Deploy OpenUnison for authentication and automation"
+  description: |
+    Deploys OpenUnison for authentication and automation.  Also deploys satelites for multi-cluster authentication.  See https://openunison.github.io/deployauth/ for detailed instructions
```
ahmetb, line 10: Do you mind wrapping this at 80 chars?

plugins/ouctl.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ouctl
+spec:
+  version: "v0.0.1"
+  homepage: https://github.com/tremolosecurity/openunison-control
+  shortDescription: "Deploy OpenUnison for authentication and automation"
+  description: Deploys OpenUnison, see https://openunison.github.io/deployauth/ for details
```
ahmetb, line 9: The new description reads like the plugin is meant to be invoked one-time. Is this the case? Or are there ongoing management subcommands?

mlbiam: in a single cluster, you can use it to both deploy and update openunison (it does error checking and timing of deployment between multiple helm charts).  In a multi cluster environment, it's used for onboarding new clusters so it would be called for each new cluster to enable authentication on it.

ahmetb:
(Moving discussion to main thread)

Typically, we don't accept plugins that are designed mostly to install/remove some component. (i.e. a script/helm chart could do that.)

I'm afraid the plugin might be more appropriate for the host-your-own Krew-index mode: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/
/kind gray-area

mlbiam:
@ahmetb isn't a tool used to automate repetitive tasks (like integrating clusters for authentication) something that kubectl plugins were built for?  This tool automates a multi-helm chart deployment (among other things).

ahmetb:
Yeah, I think it's arguable if it orchestrates something non-trivial. That said, other plugins that are very closely tied to operators (or similar external systems) like yours typically do more than installation/removal. 

Do you mind pasting examples of commands/flags your plugin supports (sorry I am afk so can't try myself right now)?

mlbiam:
@ahmetb sorry, the plugin wasn't as ready as i thought and has gone through several revisions and updates.  That said:

> Do you mind pasting examples of commands/flags your plugin supports (sorry I am afk so can't try myself right now)?

There are two modes:

`install-auth-portal` - something like `ouctl install-auth-portal -s /path/to/secret /path/to/values.yaml` will:

1.  Create a `Secret`
2. Install the operator helm chart
3. Install the orchestra helm chart
4. install the orchestra-login-portal chart

There are also the `-u` flag which lets you pre-run charts for additional configuration (like custom logos) and `-r` which lets you add additional charts for running after installation.  This would include custom applications that you integrate into OpenUnison for authentication.  Using the `ouctl` command, rather then helm, lets the cluster operator control both installs and updates in a predictable way.  The plugin does error checking and state management amongst the various components.

`install-satelite`- this mode integrates a new cluster into the OpenUnison control plane, providing a single authentication source for all your clusters.  In order to do this, the `ouctl` needs to inspect both the control plane cluster and new satelite cluster, make sure the proper certificates and authorization rules are configured, do the work of `install-auth-portal`  on the satelite and running charts on the control-plane to automatically setup SSO between the two without asking the user any questions.  Each time a new cluster is onbaorded, this process needs to be run making the `ouctl` a utility that simplifies the process and cuts down on the manual work, to your point about a plugin needing to be :" I think it's arguable if it orchestrates something non-trivial".  This integration process requires the inspection of multiple objects in both clusters including `Secret`, `ConfigMap`, `OpenUnison` CRD to generate the appropriate configurations and run the correct charts.

Let me know if you need any other details and thanks!

ahmetb:
I think my position still stands. kubectl plugins are good opportunities to improve the experience of kubectl functionality itself.

However, I think this plugin still appears to be focused on installation. Even though the installation is repetitive for each cluster, you can imagine everyone who distributes something would go down this way and develop a plugin just to get their thing installed ––and that's not what the kubectl plugins are intended for.

I think the fact that the tool is called `ouctl` kind of implies that this is meant to be a tool on its own –not necessarily a kubectl plugin.

PR #2403, opened 2022-07-06
Adding windows-debug plugin

jsturtevant:
Adding new plugin https://github.com/jsturtevant/windows-debug/releases follow up from #2402.  I ran the `kubectl krew install` locally and everything was working.

### DIFF
--- a/plugins/windows-debug.yaml
+++ b/plugins/windows-debug.yaml
@@ -1,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: windows-debug
+spec:
+  version: v0.1.2
+  homepage: https://github.com/jsturtevant/windows-debug
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/jsturtevant/windows-debug/releases/download/v0.1.2/kubectl-windows-debug-v0.1.2.tar.gz
+    sha256: 5c13553af09b98acf0001d1bc94155a7dbbbd1cb54ad84d0faaac770689e2b10
+    bin: kubectl-windows-debug
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/jsturtevant/windows-debug/releases/download/v0.1.2/kubectl-windows-debug-v0.1.2.tar.gz
+    sha256: 5c13553af09b98acf0001d1bc94155a7dbbbd1cb54ad84d0faaac770689e2b10
+    bin: kubectl-windows-debug
+  shortDescription:  Windows node access via kubectl
+  description: |
+      kubectl plugin for launching a Windows pod that will give you access to the specified node
+      and provide a few useful utilities for debugging Windows processes.
+      Access to the node is provided by a Windows Host Process Containers feature in Kubernetes.
+
+      To use this plugin you will need:
+      - kubernetes 1.22+ (with the WindowsHostProcessContainers feature-gate enabled)
+      - containerd 1.6+ as the runtime
+
+      By default it uses container image https://github.com/jsturtevant/windows-debug/pkgs/container/windows-debug
+      which has some useful utilities for debugging Windows processes pre-installed.
+      You can provide your own image by using --image flag.


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

plugins/windows-debug.yaml
```diff
@@ -0,0 +1,28 @@
...
+    sha256: 8d4cf283b80220f8729a0aa759038a52b4a8d46f49c42b1ccabc036b9f92ada9
+    bin: kubectl-windows-debug
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/jsturtevant/windows-debug/releases/download/v0.1.1/kubectl-windows-debug-v0.1.1.tar.gz
+    sha256: 8d4cf283b80220f8729a0aa759038a52b4a8d46f49c42b1ccabc036b9f92ada9
+    bin: kubectl-windows-debug
+  shortDescription:  Launch a Windows host process pod that will give you access to the specified Windows node
```
ahmetb, line 23: Shorter would be better, this would be truncated.

plugins/windows-debug.yaml
```diff
@@ -0,0 +1,28 @@
...
+  description: |
+    Usage:
+      kubectl windows debug <node-name>
```
ahmetb, line 26: we already print this `Usage` string upon installation, avoid. let users call --help and you can show them there. this will go out of date here.

also, I don't think it will be `windows debug`, should be `windows-debug` once installed. we replace `-` with `_` in the filename.

plugins/windows-debug.yaml
```diff
@@ -0,0 +1,28 @@
...
+      krew plugin for launching a Windows host process pod that will give you access to the specified node and provide a few useful
```
ahmetb, line 27: krew plugin → kubectl plugin

plugins/windows-debug.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: windows-debug
+spec:
+  version: v0.1.1
+  homepage: https://github.com/jsturtevant/windows-debug
```
ahmetb, line 7: it seems like the auth plugins are not imported https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/#auth-plugins consider fixing

plugins/windows-debug.yaml
ahmetb, line 27: `description` field is where you get to go into more detail, please go into more detail in terms of features and how does it do what it does.

ahmetb (changes requested):
Thanks for your submission.

plugins/windows-debug.yaml
jsturtevant: > we already print this `Usage` string upon installation, avoid. let users call --help and you can show them there. this will go out of date here.

sounds good.

according to https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#naming-a-plugin 

>For example, a plugin that wishes to be invoked whenever the command kubectl foo bar baz is invoked by the user, would have the filename of kubectl-foo-bar-baz

I tried it and it worked this way after installing locally via `kubectl krew install --manifest=manifest.yaml --archive=kubectl-windows-debug-v0.1.1.tar.gz`

ahmetb: weird. I think I can confidently say that it should not work. we replace the plugins that contain dashes with underscore:

```
ls ~/.krew/bin
kubectl-allctx     kubectl-example  kubectl-images       kubectl-krew  kubectl-neat        kubectl-open_svc   kubectl-rbac_lookup  kubectl-resource_capacity  kubectl-tree              kubectl-who_can
kubectl-ca_cert    kubectl-flame    kubectl-interact_sh  kubectl-kurt  kubectl-nodeinfo    kubectl-print_env  kubectl-rbac_tool    kubectl-resource_view      kubectl-view_allocations  kubectl-whoami
kubectl-eksporter  kubectl-get_all  kubectl-karmada      kubectl-mc    kubectl-oidc_login  kubectl-prof       kubectl-rbac_view    kubectl-tail               kubectl-view_secret
```

and therefore those plugins installed via krew does not work without dashes:

```
$ kubectl oidc login
Error: unknown command "oidc" for "kubectl"
Run 'kubectl --help' for usage.
$ kubectl who can
Error: unknown command "who" for "kubectl"
Run 'kubectl --help' for usage.
```

Perhaps you have another binary in your `%PATH%` with `kubectl-a-b` format.

plugins/windows-debug.yaml
ahmetb: My bad, I assumed this would be in Go, since it's intended for Windows, but I realize it targets Windows clusters/nodes.

plugins/windows-debug.yaml
jsturtevant: updated

plugins/windows-debug.yaml
jsturtevant: updated

jsturtevant: updated

plugins/windows-debug.yaml
jsturtevant: Yes I mostly use WSL so it is usable from windows in that respect.  I realize folks may want to run from powershell which is why I was considering re-writing but wanted some feedback.  I assume lots of folks just use WSL.

plugins/windows-debug.yaml
jsturtevant: After removing the previously installed binary, found a bug with the file not being marked as executable.  It is update and I bumped the version in the manifest file here

jsturtevant:
@ahmetb I've made the requested updates.  could you take another look when you get a chance? Thanks!

ahmetb:
/lgtm
/approve
Thanks!

jsturtevant:
Thank you!

PR #2427, opened 2022-07-15
add kubectl-shell plugin

hojerst:
The plugin allows to spawn a interactive shell in the cluster and also
allows to mount volumes to this shell.

### DIFF
--- a/plugins/shell.yaml
+++ b/plugins/shell.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   homepage: https://gitlab.com/hojerst/kubectl-shell
   shortDescription: "Kubernetes Shell Spawner with Mounting"
-  version: "v1.3.0"
+  version: "v1.3.1"
   description: |
     Allows to spawn a shell image in a kubernetes cluster and attach to it.
     It is similar to "kubectl run" but allows to easily mount hostPath, secret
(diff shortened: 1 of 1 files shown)


ahmetb:
Hi, "shell" name is too generic and goes against the naming guidelines. There are far too many plugins already that does what this plugin does.

hojerst:
Hi, before renaming the plugin: Would `run-shell` be ok, or is this also to generic?

Regarding "There are far too many plugins already that does what this plugin does.": I haven't seen any; thats the reason I developed this plugin in the first place. I know, there is pod-shell, debug-shell and node-shell. But all those have specific use cases which don't easy allow to attach volumes or run with specific service accounts for debugging.

ahmetb:
Yeah I meant those plugins (and I believe there's a couple more). They distinguish from one another with their names, and the names are somewhat qualifying what the plugins do.

Run-shell is also too generic. It's also no surprise that we get everyone's debugging scripts as plugins, as everyone ends up debugging some stuff one way or the other. That's why it's somewhat crucial to not give anyone firstcomers' advantage in naming, prevent the curated plugin list from going out of control with plugins with duplicate functionality with ambiguous names.

hojerst:
So my suggestions would be:

* `mount-run` as it is basically a `kubectl run` with the option to mount stuff. I would probably change the behaviour of the plugin slightly to not spawn a specific image/attach to it by default but rather have a similar behavior that `kubectl run`+mounting.
* `diagnostic-shell` without any changed behavior

If one of the names is approved, I would start implementing the change accordingly (just don't want to do that multiple times)

ahmetb:
I think the first one gives you more chance to distinguish from other similar plugins.

PR #2550, opened 2022-08-29
add execws plugin

jpts:
The plugin implements the exec protocol over websockets, which is missing from kubectl, and looks to be a little way off yet. Has some other functionality which I think is unlikely to ever make it into kubectl, such as directly talking to the kubelet API and not sending the "get pod" preflight request.

### DIFF
--- a/plugins/execws.yaml
+++ b/plugins/execws.yaml
@@ -1,0 +1,56 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: execws
+spec:
+  homepage: https://github.com/jpts/kubectl-execws
+  shortDescription: kubectl exec using WebSockets
+  version: v0.1.0
+  description: A replacement for "kubectl exec" that works over WebSocket connections.
+  caveats: |
+    Node direct-exec functionality requires connection to the kubelet API to work, since it
+    bypassess the apiserver by design.
+  platforms:
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: "https://github.com/jpts/kubectl-execws/releases/download/v0.1.0/kubectl-execws_v0.1.0_linux_amd64.tar.gz"
+    sha256: d9a622e2e90a1185647baa5134ee93634ce17d6813294a8e1c674c67592085f2
+    bin: kubectl-execws
+  - selector:
+      matchLabels:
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. I understand that it is using an alternative approach, but is it introducing net-new user visible functionality or enabling currently blocked users? The answer right now seems to be "no". If that's the case, it might be good to distribute the plugin through a custom index if you want people to try out and give feedback.

jpts:
I would propose that the tool does add user visible functionality, which is not present in kubectl:
- Allows exec into a pod even if the user/serviceaccount does not have "get pods" RBAC permissions (via `--no-sanity-check` flag)
- Allows exec into a pod if the apiserver is not directly reachable, but the nodes are (via `--node-direct-exec` flag)

Further, although the websocket implementation is largely invisible to the user, I do believe it unblocks users. Some loadbalancers do not support SPDY, e.g. AWS ALB, [Envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#http-protocols). As a result, for clusters where the apiserver is behind one such loadbalaner, users can't use exec functionality without a [workaround](https://dev.to/aws-builders/kubectl-execport-forward-with-aws-alb-and-nginx-ingress-controller-4fn0), which often requires deploying more resources. With this tool, users can just exec directly through the loadbalancer to the apiserver, as websockets is more widely supported.

ahmetb:
Got it. I think I'm ok with this in that case (I initially assumed it is identical functionality).
Do you mind expanding on these capabilities in the `description` section?

plugins/execws.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: execws
+spec:
+  homepage: https://github.com/jpts/kubectl-execws
+  shortDescription: kubectl exec using WebSockets
+  version: v0.1.0
+  description: A replacement for "kubectl exec" that works over WebSocket connections.
```
ahmetb, line 9: Shorter is better:

> kubectl exec replacement that works over WebSockets

ahmetb:
/lgtm
/approve
Thanks!

PR #2555, opened 2022-08-30
Fix plugin kc URI

astraw99:
To make the plugin URI consistent with the `.krew.yaml`in https://github.com/sunny0826/kubecm/blob/master/.krew.yaml


ahmetb:

-----

>file `plugins/kc.yaml` is not a straightforward version bump: could not find the old/new version spec in the diff
/kind plugin-update

PR #2827, opened 2022-12-22
Add rsh plugin

nilic:
This is a port of OpenShift client's `oc rsh` command to a `kubectl` plugin.

It opens a remote shell session to a running container (shell is configurable, `/bin/sh` is used by default) but can also be used for running a one-time command on a container instead of a login shell.

It works with pods, deployments, jobs, daemon sets, replication controllers and replica sets. When something other than a pod is targeted, it will be resolved to a ready pod. Container is configurable, by default first container is used.

I find this feature of `oc` very useful for troubleshooting and haven't been able to find a similar plugin which uses `exec` in the background.

### DIFF
--- a/plugins/rsh.yaml
+++ b/plugins/rsh.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: rsh
 spec:
-  version: "v0.0.3"
+  version: "v0.1.0"
   homepage: https://github.com/nilic/kubectl-rsh
   shortDescription: Open a remote shell session to a container
   description: |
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. I took some closer look but I can't tell where the major benefit of this is. It almost seems like it's a shorter way of doing `kubectl exec -i -t POD -- cmd args...`, except it can pick a pod out of deployment?

When a notable value-add is not present, we typically don't allow the plugin into the default registry. So if there's something you'd like to highlight, I most certainly appreciate.

Also worth asking: Have you implemented this yourself or is the code adopted from `oc`?

nilic:
One could say that it's an rsh opinionated alternative to `exec`. I found myself often reaching for it when working with OpenShift and wanted to port it to `kubectl` when interacting with non-OpenShift clusters.

It provides a quicker and arguably a more logical way of getting a remote shell to a container than `kubectl exec`:
* `kubectl rsh mypod` -> `kubectl exec --stdin --tty mypod -- /bin/sh`
* `kubectl rsh --shell /bin/zsh mypod` -> `kubectl exec --stdin --tty mypod -- /bin/zsh`

For consistency with `rsh` it also adopts "--"less syntax for executing commands, eg. `kubectl rsh mydeploy/mypod ls /`.

Code is adopted from `oc` with minimal changes in order to make it work as a root command. This is stated in the [NOTICE](https://github.com/nilic/kubectl-rsh/blob/main/NOTICE) file distributed in the releases as per the conditions of Apache 2.0 license used in the `oc` repo.

PR #2853, opened 2023-01-04
Add nodepools plugin to index

inkel:
This is in response of #2845 where it was stated that the first submission shouldn't be automated. This plugin allows users working with Kubernetes clusters in CSP that have the concept of node pools/groups to list them in an easier way or listing all the nodes in a given node pool/group.

Signed-off-by: Leandro López (inkel) <leandro.lopez@grafana.com>

### DIFF
--- a/plugins/nodepools.yaml
+++ b/plugins/nodepools.yaml
@@ -50,6 +50,18 @@ spec:
     bin: kubectl-nodepools
   shortDescription: List node pools/groups.
   description: |
-    Usage:
-      kubectl nodepools [list]
-      List all the nodepools in the current cluster.
+    List node pools/groups in the current cluster, alongside a count of how
+    many nodes there are in each pool/group and their type.
+
+    You can also list nodes for a given node pool/group by name.
+
+    This plugin supports the standard kubectl flags for environment selection
+    and authentication.
+
+    Example usage:
+      List node pools/groups in cluster:
+        kubectl nodepools
+      List nodes in a node pool/group:
+        kubectl nodepools nodes $nodepool
+
+    More information at https://github.com/grafana/kubectl-nodepools
--- a/plugins/nodepools.yaml
+++ b/plugins/nodepools.yaml
@@ -48,7 +48,7 @@ spec:
     uri: https://github.com/grafana/kubectl-nodepools/releases/download/v0.0.1/kubectl-nodepools_v0.0.1_windows_arm64.zip
     sha256: eeca0e865046bab307364fb303edc096f1e708a7fd4be63f40f736547dfa5bf7
     bin: kubectl-nodepools
-  shortDescription: A kubectl plugin for listing node pools/groups
+  shortDescription: List node pools/groups.
   description: |
     Usage:
       kubectl nodepools [list]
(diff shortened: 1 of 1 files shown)


inkel:
Thank you @ahmetb for the review! I'll go through your comments and push an update later today.

inkel:
@ahmetb in addition to addressing your comments I also figured out why the build was failing and I've fixed it.

plugins/nodepools.yaml
```diff
@@ -0,0 +1,67 @@
...
+    sha256: ebacb232dfd97ddb6ad5e4b710b72e3440e918a5ac1b8836341af75b87e62435
+    bin: kubectl-nodepools.exe
+  - selector:
+      matchLabels:
+        os: windows
+        arch: arm64
+    uri: https://github.com/grafana/kubectl-nodepools/releases/download/v0.0.1/kubectl-nodepools_v0.0.1_windows_arm64.zip
+    sha256: eeca0e865046bab307364fb303edc096f1e708a7fd4be63f40f736547dfa5bf7
+    bin: kubectl-nodepools.exe
+  shortDescription: List node pools/groups.
```
ahmetb, line 51: no need for trailing dot

plugins/nodepools.yaml
```diff
@@ -0,0 +1,67 @@
...
+    This plugin supports the standard kubectl flags for environment selection
+    and authentication.
+
+    Example usage:
+      List node pools/groups in cluster:
+        kubectl nodepools
+      List nodes in a node pool/group:
+        kubectl nodepools nodes $nodepool
+
+    More information at https://github.com/grafana/kubectl-nodepools
```
ahmetb, line 67: no need for example usage. users will run -h/--help
plus we print `kubectl nodepools -h` and the website after installation and the `krew info` command.
typically this info goes out of date, so I recommend omitting.

ahmetb (changes requested):
mostly good thanks.
/lgtm

plugins/nodepools.yaml
inkel: I see, that makes sense. I just borrows inspiration from other plugins (kudo I believe), I'll take it into account for an upcoming update!

ahmetb: I can merge as is but I recommend fixing.

Also it'd be cool for maybe the plugin to work without any args and print all nodes grouped by pool? (Sorry I havent used it yet, if that's the case already)

inkel: I believe you merged it already 🙀

The plugin without any arguments just return a list of node pools/groups. In fact the `list` subcommand is optional. The `nodes` subcommand is an extension that upon a given node pool/group it returns the nodes that are part of it.

ahmetb: uh oh lol , I guess /approve is an org membership thing. feel free to send a PR with the fixes.

unrelated: I think people like tree views so having a cluster-wide view/stats of node pools would be extremely useful (maybe also with some stuff like their differing node annotations)

inkel: > uh oh lol , I guess /approve is an org membership thing. feel free to send a PR with the fixes.

Sure, I'll send another PR later today. Thanks!

> unrelated: I think people like tree views so having a cluster-wide view/stats of node pools would be extremely useful (maybe also with some stuff like their differing node annotations)

In our use case we definitely use the two current views for different purposes, however, that sounds like a nice adding as a flag, something like `kubectl nodepools list -tree` or similar. I'll think about it!

PR #2899, opened 2023-01-29
Add a new plugin "eks-node"

flavono123:
The plugin print and filter AWS EKS nodes easily rather than selecting verbose named labels.

This is the first version and would be improved like following:
- Options for print formats(table, yaml, json)
- Re-implement with Go; remove the yq dependency

Signed-off-by: flavono123 <flavono123@gmail.com>

### DIFF
--- a/plugins/eks-node.yaml
+++ b/plugins/eks-node.yaml
@@ -1,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: eks-node
+spec:
+  homepage: https://github.com/flavono123/kubectl-eks-node
+  shortDescription: Print the informantions on AWS EKS nodes
+  version: v0.1.0
+  description: |
+    Easily print, filter EKS nodes rather than using verbose named labels
+  caveats: |
+    [yq](https://github.com/mikefarah/yq#install) is required for now.
+  platforms:
+  - selector:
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Hey, thanks for the submission. Based on what I'm seeing, this plugin won't be a good addition to krew-index because we typically reject plugins that are meant to replace things like one liners e.g. `kubectl get nodes -l machine-type=x`.

Similarly, you can expect a similar plugin would be developed for other cloud providers seeing this example, and that would further make the curated index more complex and crowded with minimal/no value add to the user (whereas "kubectl get nodes" can handle it all). (You can imagine the command inferring which labels to use from the cluster's cloud provider.)

I think if it was a more generic plugin that handled multiple cloud providers (and did things which would otherwise be complicated kubectl queries with multiple pipes etc), it would be a more appealing submission. (And we typically merge plugins based on their current status; not their planned additions.)

Until then you can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

I hope you can develop this more and resubmit.

/close

flavono123:
Thanks for your comment. It would be ok to use a custom index for now.

PR #2921, opened 2023-02-06
add capsh plugin

dcasati:
kubectl-capsh will automate the following laborious process.

You might need to check what capabilities(7) a running pod has for security, audit or development purposes. You could manualy exec into the pod and check for this information on /proc/pid/status

### DIFF
--- a/plugins/capsh.yaml
+++ b/plugins/capsh.yaml
@@ -1,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capsh
+spec:
+  shortDescription: check the capabilities(7) of a running Pod.
+  homepage: https://github.com/appdevgbb/utils-capsh
+  description: |
+    check the capabilities(7) of a running Pod.
+  caveats: |
+    kubectl-capsh needs capsh to run, so when executed for the first time it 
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
>    so when executed for the first time it 
>    will download that file from this repository and install it under /usr/local/bin/

this part is a bit problematic, we don't want plugins to be doing that ideally (as uninstalling the plugin would not delete that binary).

Also is this problem typically more involved than a bash one-liner like:

        kubectl exec pod -- command | something_that_parses_caps_list_or_bits

If it can be achieved the same way, maybe we shouldn't make it a plugin and rely on unix tools composability principle.

dcasati:
> > so when executed for the first time it
> > will download that file from this repository and install it under /usr/local/bin/
> 
> this part is a bit problematic, we don't want plugins to be doing that ideally (as uninstalling the plugin would not delete that binary).
> 
> Also is this problem typically more involved than a bash one-liner like:
> 
> ```
>     kubectl exec pod -- command | something_that_parses_caps_list_or_bits
> ```
> 
> If it can be achieved the same way, maybe we shouldn't make it a plugin and rely on unix tools composability principle.

PR #2924, opened 2023-02-07
adding kubescape plugin with version v2.1.2

matthyx:
Signed-off-by: Matthias Bertschy <matthias.bertschy@gmail.com>

### DIFF
--- a/plugins/kubescape.yaml
+++ b/plugins/kubescape.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kubescape
+spec:
+  homepage: https://github.com/kubescape/kubescape/
+  shortDescription: Scan resources and cluster configs against security frameworks.
+  version: v2.1.3
+  description: |
+    It includes risk analysis, security compliance, and misconfiguration scanning
+    with an easy-to-use CLI interface, flexible output formats, and automated scanning capabilities.
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/kubescape/kubescape/releases/download/v2.1.3/kubescape-macos-latest.tar.gz
+    sha256: b1f02977ebadc7c626843d171e53d40ddd11a98d1c00aa8664f4f058d5df24e4
+    bin: kubescape
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/kubescape/kubescape/releases/download/v2.1.3/kubescape-ubuntu-latest.tar.gz
+    sha256: 3d41997c6d1a6b694c5fa1bf8c1cedf2a80ef3357cef0d08075fdc7e906a4cb6
+    bin: kubescape
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/kubescape/kubescape/releases/download/v2.1.3/kubescape-windows-latest.tar.gz
+    sha256: 00127cd9eaf8dc0578db76b7548eea28578729f0ecdecdfe397a2f402b3dbd20
+    bin: kubescape


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

matthyx:
sorry I'll add `LICENSE` in the next release

ahmetb:
Do you mind elaborating how the plugin makes the kubectl experience better? Many standalone tools tend to distribute as a kubectl plugin and sometimes it's not a good fit. 

Also the name is against our naming guidelines (due to kube prefix, which is often the indicator of the earlier point).

matthyx:
@ahmetb you're right, distributing a standalone tool should not be the main motivation...

I took a look into existing [plugins](https://krew.sigs.k8s.io/plugins/) and would like to know your thoughts on the following plugins:
- kubesec-scan
- kyverno
- linstor
- liqo
- vela

Now, coming back to kubescape, let's say if I rework the way to interact with our CLI to seamlessly integrate as a kubectl plugin, would you be open to reconsidering this submission?
I was thinking about the `scan` keyword, and I can elaborate how `kubectl scan control C-0168` or `kubectl scan framework MITRE` would make cluster administrators live easier.

Is it possible to have multiple keywords for a plugin?

Thanks in advance for your help :-)

ahmetb:
@matthyx many of those plugins were there prior to instituting naming and plugin functionality guidelines. Though, I think since the plugin is very similar to plugins like `preflight`, and still uploads data about objects to somewhere for validation, I'd be ok with accepting it.

> I was thinking about the `scan` keyword,

This would be definitely against naming guidelines, as you discovered there are almost a dozen plugins that do "scan"ning in some form or another. So, it's not preferable to assign it to a particular plugin.

ahmetb:
I also would like to mention this project is already popular --why does it need to be distributed in yet another form factor (kubectl plugin) in your opinion?

craigbox:
What's the point of krew if it only offers unpopular software?

ahmetb:
> What's the point of krew if it only offers unpopular software?

This is not the argument here. Krew is a tool that aims to make it easy to distribute plugins for kubectl; it's not a general-purpose package manager for any tool related to Kubernetes. 

From a user's perspective, there is little to no value in packaging every single Kubernetes tool as a kubectl plugin (e.g. `kubectl kubeadm`, or `kubectl minikube`) when the project is already a popular standalone binary. 

This applies to the curated index we have here (and you can host your own index fairly easily). If you do that, you can choose your own name (e.g. `kubectl scan`, too).

Typically the `kube*` prefix is an indicator that the tool might as well be a standalone tool, so I assumed it's better off that way. I now looked at the project more closely (the website doesn't load for me), and after reading https://github.com/kubescape/kubescape/blob/master/docs/getting-started.md#install-kubescape, it seems the tool queries resources and performs analysis on Kubernetes resources. This is well within what makes a plugin (as it's mainly about resource management/querying capabilities of kubectl).

I'm happy to accept it, we also have similar plugins, but the name would read: `kubectl kubescape`, which I assume is not very ideal(?). I am still not positive about assigning a generic name like `scan`.

craigbox:
> This is well within what makes a plugin (as it's mainly about resource management/querying capabilities of kubectl).

Definitely; it starts by using your `.kubeconfig` file.

> but the name would read: `kubectl kubescape`

I think this is the best outcome tbh.  There may be use cases that relate to different tasks, and I agree that one generic word isn't the best way to expose them (from anyone's POV).  It just falls on the obvious name happening to be `kube*`, and if you can look past the duplication, so can we 😊

plugins/kubescape.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kubescape
+spec:
+  homepage: https://kubescape.io/
```
ahmetb, line 6: doesn't exist?

plugins/kubescape.yaml
```diff
@@ -0,0 +1,39 @@
...
+  shortDescription: An open-source Kubernetes security platform for your IDE, CI/CD pipelines, and clusters
```
ahmetb, line 7: Try to be more concise as this will be truncated if too long. Try to remove words that don't add value when you have limited space:
e.g.

> Scan resources and cluster configs for seucirty

plugins/kubescape.yaml
```diff
@@ -0,0 +1,39 @@
...
+  version: v2.1.2
+  description: |
+    Kubescape is an open-source Kubernetes security platform.
+    It includes risk analysis, security compliance, and misconfiguration scanning.
+    Targeted at the DevSecOps practitioner or platform engineer,
+    it offers an easy-to-use CLI interface, flexible output formats, and automated scanning capabilities.
+    It saves Kubernetes users and admins precious time, effort, and resources.
```
ahmetb, line 14: IMO this sounds a bit like a marketing pitch and product vision. 

> it offers an easy-to-use CLI interface, flexible output formats, and automated scanning capabilities

seems like the only useful part while trying to understand if I (as a user) should install the plugin or not.

ahmetb (changes requested):

matthyx:
Thanks @ahmetb I'll work on this PR early next week.
@Oshratn PTAL

plugins/kubescape.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: Plugin
+metadata:
+  name: kubescape
+spec:
+  homepage: https://github.com/kubescape/kubescape/
+  shortDescription: Scan resources and cluster configs against security frameworks.
+  version: v2.1.3
+  description: |
+    It includes risk analysis, security compliance, and misconfiguration scanning
+    with an easy-to-use CLI interface, flexible output formats, and automated scanning capabilities.
```
ahmetb, line 11: This could use a better description but you can update later.

plugins/kubescape.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kubescape
+spec:
+  homepage: https://github.com/kubescape/kubescape/
```
ahmetb, line 6: Recommend importing `import _ "k8s.io/client-go/plugin/pkg/client/auth"` as the users typically expect plugins to work on cloud clusters.

ahmetb:
Approving `kube*` prefix in naming with the exception since no other good name exists and the project already has an established name with notable user base.
/lgtm
/approve

plugins/kubescape.yaml
matthyx: hi @ahmetb can you point out an example?

matthyx:
> Also, please consider setting up [Krew release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) which helps you skip manually making updates to your Krew manifests on each new version and send a pull request. It’s a GitHub workflow bot that can run every time you push a new tag. Those PRs are automatically approved.

I have it... it's just the 1st commit that needs to be done manually

matthyx:
thanks!

PR #2927, opened 2023-02-07
add capsh

dcasati:
This adds a refactored  version of capsh. Worth mentioning:

- With this version we do not copy the static linked capsh file to /usr/local/bin. 
- A static version of capsh will be copied to ~/.kube/capsh 
- Having a static version of capsh allows us to run this against a running pod and it has been tested on macOS and Linux.

### DIFF
--- a/plugins/capsh.yaml
+++ b/plugins/capsh.yaml
@@ -1,0 +1,34 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: capsh
+spec:
+  shortDescription: check the capabilities(7) of a running Pod.
+  homepage: https://github.com/appdevgbb/utils-capsh
+  description: |
+    check the capabilities(7) of a running Pod.
+  caveats: |
+    kubectl-capsh needs capsh to run, so when executed for the first time it 
+    will download that file from this repository and install it under .kube/capsh 
+
+    Usage:
+      kubectl capsh <pod>
+  version: v0.1.1
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: os
+        operator: In
+        values:
+        - darwin
+        - linux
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
> * A static version of capsh will be copied to ~/.kube/capsh

This isn't much different than what you were previously proposing.

Is there a reason why you aren't reimplementing this capability parsing functionality yourself? At this point, you're bundling and distributing a GPL binary which requires you to be very careful around licensing (i.e. the GPL license must follow wherever the code goes etc.) 

Similarly, it makes it hard to support macOS and Windows (I assume capsh doesn't have that distribution)?

Also, because it's a binary you're bundling in your release, it will make it difficult for the users to understand where the code for capsh binary itself came from and will be a lot less trusted. 

In general we don't recommend this approach of bundling additional binaries for such reasons.

dcasati:
> > * A static version of capsh will be copied to ~/.kube/capsh
> 
> This isn't much different than what you were previously proposing.
> 
> Is there a reason why you aren't reimplementing this capability parsing functionality yourself? At this point, you're bundling and distributing a GPL binary which requires you to be very careful around licensing (i.e. the GPL license must follow wherever the code goes etc.)
> 
> Similarly, it makes it hard to support macOS and Windows (I assume capsh doesn't have that distribution)?
> 
> Also, because it's a binary you're bundling in your release, it will make it difficult for the users to understand where the code for capsh binary itself came from and will be a lot less trusted.
> 
> In general we don't recommend this approach of bundling additional binaries for such reasons.

You’re right and because of the GPL restrictions it really wouldn’t make sense to have this at all. Linux user could run the command you’ve suggested with kubectl exec and piping that to capsh locally. 

I really don’t have any other ideas here but open to suggestions.

PR #3040, opened 2023-03-29
add kubectl-ai plugin

sozercan:
adds https://github.com/sozercan/kubectl-ai plugin to krew index

### DIFF
--- a/plugins/ai.yaml
+++ b/plugins/ai.yaml
@@ -1,0 +1,39 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ai
+spec:
+  version: v0.0.6
+  homepage: https://github.com/sozercan/kubectl-ai
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. I think the name "ai" takes advantage of being a first-comer. Typically we try not to assign broad terms to any particular plugin.

By assigning this name, it basically takes the opportunity from all and every AI related plugin to kubectl. Hope that makes sense. 

A recently added relevant plugin was called "kopilot" and that was unique enough that it didn't warrant this discussion. 

So if it's not too late, I recommend a more relevant name, like "ai-example" that can also explain the functionality better. You can also always host a custom plugin index to distribute it yourself more freely.

sozercan:
@ahmetb thanks! that makes sense. would repo or binary name have to match the plugin metadata name? for example, plugin name being `ai-example` or similar, while binary name as `kubectl-ai`.

ahmetb:
Repo name or the executable name in the tar/zip archive does not have to match to the `name:` field in the plugin manifest, but the `name` field determines how the symlink is created (and therefore how the plugin is invoked).

We have quite a few plugins that act both as a standalone tool as well as a kubectl plugin (or plugins that have a different name in krew), and they print the right usage/--help outputs based on the invocation method.

PR #3110, opened 2023-05-02
Add plugin gateway

tafkam:
Hi!

kubectl krew install --manifest=gateway.yaml --archive=kubectl-gateway_linux_amd64.tar.gz
Installing plugin: gateway
Installed plugin: gateway
\
 | Use this plugin:
 |      kubectl gateway
 | Documentation:
 |      https://github.com/tafkam/kubectl-gateway
/

### DIFF
--- a/plugins/gateway.yaml
+++ b/plugins/gateway.yaml
@@ -1,0 +1,44 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: gateway
+spec:
+  version: v0.0.1
+  homepage: https://github.com/tafkam/kubectl-gateway
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/tafkam/kubectl-gateway/releases/download/v0.0.1/kubectl-gateway_darwin_amd64.tar.gz
+    sha256: 9896b258d047afa818d82ee7433962c9707e8a601f656e8b38a99aed9f7f2db7
+    bin: kubectl-gateway
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: arm64
+    uri: https://github.com/tafkam/kubectl-gateway/releases/download/v0.0.1/kubectl-gateway_darwin_arm64.tar.gz
+    sha256: cde92a6cdd493a6c9d0b9cbe82325e74f2710031f8c92a847234fefeb4eefedb
+    bin: kubectl-gateway
+  - selector:
+      matchLabels:
+        os: linux
+        arch: amd64
+    uri: https://github.com/tafkam/kubectl-gateway/releases/download/v0.0.1/kubectl-gateway_linux_amd64.tar.gz
+    sha256: 6856e55de1abfe5bebc788cb6764c89fc13f2144feadbbdb39c8d7514e763217
+    bin: kubectl-gateway
+  - selector:
+      matchLabels:
+        os: windows
+        arch: amd64
+    uri: https://github.com/tafkam/kubectl-gateway/releases/download/v0.0.1/kubectl-gateway_windows_amd64.zip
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. The name is too generic and is a first party concept in Kubernetes ecosystem.  We don't assign such names to a particular first comer unless they're an official Kubernetes project plugin.

Please refer to the naming guide in Krew docs. 
/hold

tafkam:
That's alright. I would contribute if I had a clue about creating an official kubernetes kubectl plugin.😀 I guess that's not easy to achieve with the organizational overhead.
Would a rename of the plug-in to "manage-gateway" suffice?

ahmetb:
I recommend proposing a plugin like this to the kubectl project. There are builtin commands like `kubectl create <resource>` and if that is still a pattern actively pursued, it's worth considering.

Overall I'm not sure if this is a plugin that adds value in the centralized krew-index:
- We could have a plugin/cmd for every single resource, but this largely doesn't make sense because Kubernetes already has a resource-agnostic way to manage every type of resource (resource manifests and kubectl/kustomize etc)
- The plugin installs/uninstalls CRDs, which again, I'm not sure if it's a good idea. Why would the users not use YAMLs declaratively and instead manage Gateway resource or parts of it declaratively?

I don't think this is how the user community manages resources? In that case, I don't see the value it adds broadly to the kubectl users.

You can still distribute your plugin via Krew using [custom indexes](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/). (Users will be downloading the plugin manifest from your repository instead of krew-index.)

tafkam:
> I recommend proposing a plugin like this to the kubectl project. There are builtin commands like `kubectl create <resource>` and if that is still a pattern actively pursued, it's worth considering.

True, but the Gateway API is still shipped outside of Kubernetes as CRDs, until someday when it's mature enough. I guess that will stand in the way of directly adding functionality to kubectl.

> Overall I'm not sure if this is a plugin that adds value in the centralized krew-index:
> 
> * We could have a plugin/cmd for every single resource, but this largely doesn't make sense because Kubernetes already has a resource-agnostic way to manage every type of resource (resource manifests and kubectl/kustomize etc)

Also, agreed. The Gateway API is relatively new though. I see value in fiddling around with it (like "kubectl create ingress") to increase adoption. Nobody in the right mind would setup Ingress in production with kubectl. Same with Gateway API objects.

> * The plugin installs/uninstalls CRDs, which again, I'm not sure if it's a good idea. Why would the users not use YAMLs declaratively and instead manage Gateway resource or parts of it declaratively?

As mentioned before, this plugin doesn't aim to be a replacement for a more serious declarative deployment, which should be handled with GitOps, where you shouldn't need kubectl at all.

My main use-case for starting this plugin is not fully implemented yet. I started with the basics to "get going" with the Gateway API and implemented some low hanging fruit with the current functionality. The main function of this plugin will be "patching" (adding, removing, copying with modifications) segments of (mostly) HTTPRoutes to shift traffic (e.g. with defined headers) to another version of a (micro-)service deployment for development purposes. 
Using declarative/GitOps managed resources for "I quickly want to spin up my new microservice featurebranch to test it in a fully deployed environment with hundreds of other microservices" is cumbersome, when you have to modify hundreds or thousands of GitOps managed HTTPRoutes. Using a script or pipeline or devspace or whatever non-declarative way to make those changes with a kubectl plugin makes it easy though.

> I don't think this is how the user community manages resources? In that case, I don't see the value it adds broadly to the kubectl users.

Agreed, the term "manage" in the proposed renaming might be misleading.

> You can still distribute your plugin via Krew using [custom indexes](https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/). (Users will be downloading the plugin manifest from your repository instead of krew-index.)
> 
> /kind gray-area

tafkam:
Im gonna close this until the plugin isn't vaporware anymore, and maybe I'll find a better name on the way :) Thank you for your efforts @ahmetb!

PR #3124, opened 2023-05-10
add forward-exec plugin

tgunsch:
Hi,

not sure if this really worth a plugin and useful for others.

For our integration and functional tests we often needed a smoke test, where we need to 
* create a port-forward to a pod /service
* execute a curl command to  test the pod or get data from them.
   * for security reasons we only have scratch images without a curl command, so we execute the curl locally / in CI system
*  finally we drop the port-forward.

Until now we had a simple helper script to do this. Now I have experimented with kubectl plugins, so I converted the script to a plugin.

Example usage:

```
kubectl forward-exec --context prod -n my-namespace svc/my-service 8080:80 -- curl --connect-timeout 5 --retry 3 --retry-connrefused http://localhost:8080/whatever
```

BTW: We currently implement a go version, which waits until the port-forward is established before starting the command.

### DIFF
--- a/plugins/forward-exec.yaml
+++ b/plugins/forward-exec.yaml
@@ -1,0 +1,24 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: forward-exec
+spec:
+  version: "v0.0.1"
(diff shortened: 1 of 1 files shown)


tgunsch:
We will first finish the migration to go and then create another PR.

PR #3289, opened 2023-08-05
Add otel

agardnerIT:
# This PR

- Adds the `otel` plugin

# This Plugin

Uses [tracepusher](https://github.com/agardnerit/tracepusher) to generate [OpenTelemetry](https://opentelemetry.io) traces for `kubectl` commands.

### DIFF
--- a/plugins/otel.yaml
+++ b/plugins/otel.yaml
@@ -21,3 +21,11 @@ spec:
         - darwin
         - linux
     uri: https://github.com/agardnerIT/kubectl-otel/archive/refs/tags/0.1.0.tar.gz
+    sha256: 345d6260820be1def9c51929fe16ad26bcbefde46794d141b0ba746516ee3eb4
+    files:
+    - from: kubectl-otel
+      to: .
+    - from: LICENSE
+      to: .
+    bin: kubectl-otel
+    
--- a/plugins/otel.yaml
+++ b/plugins/otel.yaml
@@ -1,0 +1,23 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: otel
+spec:
+  homepage: https://github.com/agardnerit/kubectl-otel
+  shortDescription: Trace kubectl commands with OpenTelemetry
+  version: v0.1.0
+  description: |
+    Generate OpenTelemetry traces for any kubectl command
+  caveats: |
+    You must have the tracepusher binary in your path.
+    tracepusher is a standalone binary to generate OpenTelemetry spans.
+    See https://github.com/agardnerIT/kubectl-otel for details.
+  platforms:
+  - selector:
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Hi, unless this is an official otel project, we cannot grant this name to your plugin.

agardnerIT:
`Hi, unless this is an official otel project, we cannot grant this name to your plugin.`

Why? Nothing in the plugin naming guide forbids this? The name `otel` describes what the plugin does. `otel` is an abbreviation. I would understand if it was called `opentelemetry`.

What _would_ a valid plugin name be? `span`? `tp`? `tracepusher`? `tracer`?

`kubectl trace` is a plugin using `bpftrace` and that was excepted?

ahmetb:
> I would understand if it was called `opentelemetry`.

It just so happens `otel` is practically the de facto name for OpenTelemetry project though?

You'll find inconsistencies before we started enforcing this. We strive to not to grant first-comer advantage to project/vendor names --or terms like `trace` that are fairly common in this domain and might  be assigned to more than one project.

It's not too hard to be creative and still have natural-sounding cmds here like: `kubectl traced`, `with-trace`, `with-otel`, `with-telemetry`, ...

agardnerIT:
@ahmetb I've spoken to @austinlparker (an OpenTelemetry maintainer) and he stated that [he doesn't believe](https://cloud-native.slack.com/archives/CJFCJHG4Q/p1691885218094519) the project reserves `otel / OTEL` or any variant for use. So we should be OK to proceed here, right?

austinlparker:
We don't reserve it from a trademark perspective, certainly.

austinlparker:
I would say that I'm not super familiar with this PR but it seems like it only pushes traces? I would probably suggest more specific naming since OpenTelemetry supports metrics, traces, logs, etc.

ahmetb:
Yeah, otel really isn't a good name for this plugin. It's the means to an end, I think we can get creative and find something better here that doesn't give the impression that this is the official kubectl plugin for otel.

agardnerIT:
Correct, this plugin (currently) emits only traces, but there's no reason it couldn't emit metrics and logs (good enhancement).

In terms of naming, `strace` and `trace` are both already taken. What about `oteltrace`?

ahmetb:
I'll suggest a couple more alternatives that are less likely to be taken away by anyone else:

    kubectl traced 
      e.g. kubectl traced get pods 

    kubectl otrace

ahmetb:
Hi, @agardnerIT I've reviewed the source code of the plugin and basically it's setting a couple of environment variables around kubectl around using another command to execute kubectl. Typically we don't allow plugins that can be achieved with one-liners, and this seems to be roughly that. 
/kind gray-area
/close

PR #3326, opened 2023-08-24
Add shell-ctx plugin

glemsom:
Simple plugin to enable each shell-instance to operate in its own isolated Kubernetex context

### DIFF
--- a/plugins/shell-ctx.yaml
+++ b/plugins/shell-ctx.yaml
@@ -1,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: shell-ctx
+spec:
+  version: "v1.0.2"
+  homepage: https://github.com/glemsom/shell-ctx
+  shortDescription: "Kubernetes shell isolated context switcher"
+  description: |
+    Enabled each instance of a shell to operate in its own isolated Kubernetes context.
+    Supported shells: bash, zsh and fish.
+  caveats: |
+    Requirements: bash, awk, sed, fzf and kubectl
+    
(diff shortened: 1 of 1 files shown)


plugins/shell-ctx.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: shell-ctx
+spec:
+  version: "v1.0.2"
+  homepage: https://github.com/glemsom/shell-ctx
+  shortDescription: "Kubernetes shell isolated context switcher"
```
ahmetb, line 8: I think this would be a higher signal:

> Launch isolated shell for context

what do you think?

glemsom: I see the point of the current shortDescription being slightly misleading.
But `shell-ctx` does actually not spawn a new shell.

What about:
> Shell independent context switching

Will that be better ?

Clarification;
The basic idea behind `shell-ctx` is to hook into the users current shell-instance - allowing the user to keep the same current shell-history,  and any other shell-specific settings the user might have.
This will enable Kubernetes administrators to have multible shell windows open, and switch Kubernetes context per shell instance - without risking overwriting currentContext of any other shell instances.

glemsom:
Updated pull request with a better shortDescription https://github.com/kubernetes-sigs/krew-index/commit/ed545703429904a80801235990b504a5ba48cd85

plugins/shell-ctx.yaml
ahmetb: Ah yeah makes sense.

ahmetb:
/lgtm
/approve
Thanks!

PR #3449, opened 2023-10-25
Add backup-diff plugin to index

milapj:
* `backup-diff` creates a backup of a resource and compares the change between the backed-up state at a specific point in time and the current state, showing the state drift. 
* Functions similar to git diff. This can be pretty useful during disaster recovery or debugging to recover previous states of resources that were updated on the fly w/o having to re-release.
* The backup is stored at `/tmp/backups/<resource-type>` by default, but it can be stored at a user-specified path. The resources can also be re-created using the backup files.

* [github repo](https://github.com/milapj/kubectl-backup-diff)

https://github.com/kubernetes-sigs/krew-index/assets/9828402/380264cf-1f04-44f7-b4e3-6c928f4c8fa7

### DIFF
--- a/plugins/backup-diff.yaml
+++ b/plugins/backup-diff.yaml
@@ -1,0 +1,29 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: backup-diff
+spec:
+  version: "v1.0.2"
+  homepage: https://github.com/milapj/kubectl-backup-diff
+  shortDescription: "Backups a resource and diffs the state change with current"
+  description: |
+    Creates a backup of a resource and compares the change 
+    between the backed-up state at a specific point in time 
+    and the current state. Functions similar to git diff. 
+    Useful during diaster recovery.
+  platforms:
+  - selector:
+      matchExpressions:
+      - key: "os"
+        operator: "In"
+        values:
+        - darwin
+        - linux
+    uri: https://github.com/milapj/kubectl-backup-diff/archive/refs/tags/v1.0.2.tar.gz
+    sha256: 86f52e0bba84caa126720668f6c0401bc864bc850b60f2d07a18b6166ccf42f5
+    files:
+    - from: "kubectl-backup-diff-1.0.2/cmd/kubectl-backup-diff"
+      to: "kubectl-backup-diff"
+    - from: "kubectl-backup-diff-1.0.2/LICENSE"
+      to: "."
+    bin: kubectl-backup-diff
--- a/plugins/backup-diff.yaml
+++ b/plugins/backup-diff.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   version: "v1.0.2"
   homepage: https://github.com/milapj/kubectl-backup-diff
-  shortDescription: Backs up a resource, diffs state change with current
+  shortDescription: Back up a resource, shows state drift with current
   description: |
     Resource backup, git-diff like compare state for recovery
   platforms:
--- a/plugins/backup-diff.yaml
+++ b/plugins/backup-diff.yaml
@@ -5,12 +5,9 @@ metadata:
 spec:
   version: "v1.0.2"
   homepage: https://github.com/milapj/kubectl-backup-diff
-  shortDescription: "Backups a resource and diffs the state change with current"
+  shortDescription: Backs up a resource, diffs state change with current
   description: |
-    Creates a backup of a resource and compares the change 
-    between the backed-up state at a specific point in time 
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission. Based on your description, this is not too different than what a few UNIX tools and existing plugins offer, e.g.

```
kubectl get RESOURCE | kubectl neat (or eksporter) > 1.yaml
kubectl edit RESOURCE
kubectl get RESOURCE | kubectl neat (or eksporter) > 2.yaml 
diff -y 1.yaml 2.yaml
```

In these situations, we typically encourage use of existing tools and UNIX tool composability principle instead of creating a separate plugin for every combination tools. These plugins tend to go unmaintained over time, which benefits neither the users nor the maintainer.

milapj:
> Thanks for the submission. Based on your description, this is not too different than what a few UNIX tools and existing plugins offer, e.g.
> 
> ```
> kubectl get RESOURCE | kubectl neat (or eksporter) > 1.yaml
> kubectl edit RESOURCE
> kubectl get RESOURCE | kubectl neat (or eksporter) > 2.yaml 
> diff -y 1.yaml 2.yaml
> ```
> 
> In these situations, we typically encourage use of existing tools and UNIX tool composability principle instead of creating a separate plugin for every combination tools. These plugins tend to go unmaintained over time, which benefits neither the users nor the maintainer.

well, there are quite a few plugins in the index whose behavior can be replicated using stock kubectl plugins + unix commands. This plugin offers a way to keep track of state per resource, which can be accessed via a pointer

ahmetb:
> well, there are quite a few plugins in the index whose behavior can be replicated using stock kubectl plugins + unix commands.

We try to keep those to a minimum (and they might be added before we started questioning usefulness of plugins). We historically rejected _many_ plugins because they could be achieved by putting together 1-2 commands or pipes, or using other kubectl plugins. If you look at label `gray-area` you'll find plenty of examples.

You cited scenarios like disaster recovery, but I am assuming the plugin works on only one resource at a time. I don't think this is all that much different than the snipped I listed above. For example, if I wanted to back up all resources, I would use `kubectl get-all` today. If I wanted to back up a single resource, I probably would use `kubectl get KIND -A`. On top of that, the idea of backup/restore only rarely works (given some resources have ordering relationships between them, or webhooks enforcing a particular state).

In the past, I've given plugin authors advice to do some user research, and listen to feedback to come up with actually useful plugins that solve an actual problem. It's really easy to implement some plugin idea by hacking a few lines of a bash script, but that doesn't make the tool broadly useful.

Hope you understand my point –if we accepted every single plugin, it would be an enormous dumping ground. So we try to do our best to curate the list.  Also, you can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way, users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

milapj:
> > well, there are quite a few plugins in the index whose behavior can be replicated using stock kubectl plugins + unix commands.
> 
> We try to keep those to a minimum (and they might be added before we started questioning usefulness of plugins). We historically rejected _many_ plugins because they could be achieved by putting together 1-2 commands or pipes, or using other kubectl plugins. If you look at label `gray-area` you'll find plenty of examples.
> 
> You cited scenarios like disaster recovery, but I am assuming the plugin works on only one resource at a time. I don't think this is all that much different than the snipped I listed above. For example, if I wanted to back up all resources, I would use `kubectl get-all` today. If I wanted to back up a single resource, I probably would use `kubectl get KIND -A`. On top of that, the idea of backup/restore only rarely works (given some resources have ordering relationships between them, or webhooks enforcing a particular state).
> 
> In the past, I've given plugin authors advice to do some user research, and listen to feedback to come up with actually useful plugins that solve an actual problem. It's really easy to implement some plugin idea by hacking a few lines of a bash script, but that doesn't make the tool broadly useful.
> 
> Hope you understand my point –if we accepted every single plugin, it would be an enormous dumping ground. So we try to do our best to curate the list. Also, you can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way, users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

I do understand your point and I agree with it. Wouldn't want krew to become like npm hah. Well, I guess I'll come back with some other new idea for a plugin. Cheers!

milapj:
closing this PR, thanks @ahmetb, for your feedback

PR #3588, opened 2024-01-16
Add searchcm.yaml

pbletard:
This tool comes without pretention, not sure if this really worth a plugin and useful for others.
Until now we were looking through our gitlab and with classical 'kubectl get cm' trials. Decided to create a simple helper script to do this.
Example usage: kubectl searchcm -s valuetosearch


ahmetb:
As you noted, we prefer existing tools to be composed instead of their combination to be offered as a new kubectl plugin. 

I think the task described here can be achieved with kubectl|jq.

SamuelBECK1:
:+1:

PR #3680, opened 2024-03-07
Update Popeye v0.20.5

mstoops:
This update not only bumps the version, but allows for more platforms to be supported. The Popeye plugin already has all the respective platforms (linux/darwin/windows) and architectures (Intel and ARM) compiled, but the current manifest doesn't reflect them all.

### DIFF
--- a/plugins/popeye.yaml
+++ b/plugins/popeye.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   homepage: https://popeyecli.io
   shortDescription: Scans your clusters for potential resource issues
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

>file `plugins/popeye.yaml` is not a straightforward version bump: change on line `7` does not seem like a version bump: [`+      files:`]
/kind plugin-update

ahmetb:
Please update https://github.com/derailed/popeye/blob/master/.krew.yaml instead. That file is the source of truth for the file you're trying to update.

PR #3908, opened 2024-06-18
Add kubectl-kl plugin file.

AmitGujar:
I followed the naming guide and I was able to install the plugin locally.
Please include my plugin.


ahmetb:
I'm sorry neither the plugin name is descriptive enough to reflect the plugin's functionality, nor the description clearly explains what the plugin does, and the repo has little to detail about what this plugin is about and what it looks like. I'm afraid we can't accept it in its current shape.

AmitGujar:
Updated the following things:
- The plugin name is descriptive now
- Long description

plugins/kswitch.yaml
```diff
@@ -0,0 +1,26 @@
...
+      sha256: b2e06e31730f9d24eca528ce853c2ee94dc610f06169a6c5ebd3f338fbffce8b
+      bin: kubectl-kswitch
+      files:
+        - from: kubectl-kswitch
+          to: .
+        - from: LICENSE
+          to: .
+  shortDescription: Switch AKS cluster on & off
+  description: |
+    kswitch lets you to turn AKS cluster on & off, Switch between multiple namespaces with use command.
```
ahmetb, line 23: Isn't this ctx/ns plugins already do?

plugins/kswitch.yaml
```diff
@@ -0,0 +1,26 @@
...
+    Additonally, it also provides the ability to create and delete namespaces.
```
ahmetb, line 24: Isn't this already  what "kubectl delete|create namespace" already do? Why would people need an AKS specific plugin?

AmitGujar: I am planning to create simple ways to perform tasks on aks. Started from the cluster on & off. I am working on adding more features.

ahmetb:
Currently there are multiple plugins tools that allow switching between clusters and namespaces. I'm currently failing to see how's AKS related to these tasks and why the plugin is specific to AKS.

Maybe when you develop the plugin more, a better theme can emerge, and you can pick a better name to submit. Right now, it's unclear what the plugin does, there's not enough documentation about it, and it seems you're not sure where to take the plugin.

I recommend developing the plugin further and re-submitting later. 

/close

PR #4064, opened 2024-09-12
Add inject-shell plugin

inerplat:
### Add `inject-shell` Plugin

**Description:** `kubectl-inject-shell` is a tool for injecting a BusyBox environment into a running Kubernetes container, even if it lacks a shell. It sets up a privileged container on the same node as the target pod, allowing you to debug and access the container's filesystem. It supports specifying various options like namespace, image, and container for customized execution.

**Features:**
- Injects BusyBox into running containers
- Supports containers without shells (e.g., distroless images)
- Creates a privileged container on the same node for debugging
- Allows customization through options for namespace, image, and container

**Usage:**
```sh
kubectl inject-shell <pod-name> [options]
```

For detailed usage and additional information, please refer to the [GitHub repository README](https://github.com/inerplat/kubectl-inject-shell).

### DIFF
--- a/plugins/inject-shell.yaml
+++ b/plugins/inject-shell.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: inject-shell
 spec:
-  version: "v0.0.5"
+  version: "v0.0.6"
   homepage: https://github.com/inerplat/kubectl-inject-shell
   shortDescription: "Injects a BusyBox environment into running Kubernetes containers for debugging."
   description: |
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Hi, thanks for the submission. Can you please elaborate how is this different than these options:

- `kubectl debug` builtin command
- plugin https://github.com/kubernetes-sigs/krew-index/blob/f227a0853c7988486d71ab261ab055ff39badf85/plugins/pod-shell.yaml
- plugin https://github.com/kubernetes-sigs/krew-index/blob/f227a0853c7988486d71ab261ab055ff39badf85/plugins/pod-shell.yaml

If it can be achieved with one of these means, we probably should not accept this.

inerplat:
Hi, thanks for reviewing my PR. 🤗 @ahmetb 

I believe the `inject-shell` plugin serves a different purpose compared to existing tools:

### `kubectl debug` command
The built-in `kubectl debug` command includes a `--share-processes` flag, which allows sharing the process namespace with existing containers. However, it only supports sharing the process namespace and does not provide access to the container’s filesystem without a volume mount. Since volume mounts cannot be added to a running pod, kubectl debug cannot modify the existing filesystem.

In contrast, the `inject-shell` plugin is particularly useful for troubleshooting distroless or minimal containers that lack a shell. Instead of creating a new container, the `inject-shell` plugin injects BusyBox binaries and libraries directly into the existing container, allowing access to the container’s filesystem. This approach simplifies troubleshooting for containers without a shell or those with minimal setups.

### `pod-shell` plugin
The `pod-shell` plugin, on the other hand, simply provides access to the pod's shell, but does not offer the advanced functionality of injecting binaries or libraries like the `inject-shell` plugin does.

Since these plugins address different use cases, I believe each has its own value depending on the situation.

PR #4077, opened 2024-09-16
adds ranching.farm plugin.

fbettag:
This adds the plugin for my ranching.farm project. Tested in terms of installation as well. Ideally i wanted "ranching.farm" as a command but apparently having dots is invalid -.-

### DIFF
--- a/plugins/ranching_farm.yaml
+++ b/plugins/ranching_farm.yaml
@@ -10,7 +10,7 @@ spec:
           os: linux
           arch: amd64
       uri: https://github.com/ranching-farm/kubectl-addon/releases/download/v0.1.0/kubectl-ranching.farm_linux_amd64.zip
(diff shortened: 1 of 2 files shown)


plugins/ranching_farm.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: ranching_farm
```
ahmetb, line 4: sorry for the late review. have you thought of picking a name like `ranching`, `ranching-farm` which are a bit more straightforward?

fbettag: i can do that, i'm not 100% married to it yet :)

ahmetb:
The project appears to be abandoned, closing due to inactivity. Feel free to resubmit later.

PR #4227, opened 2024-11-17
add krew manifest for snapshot-api-objects

oguzhan-yilmaz:
## About

This plugins loops through namespaces, kubernetes api object types and gets the `-o yaml` output, pipes it to `kubectl neat -` create a snapshot of the currently installed K8s API objects to filesystem.

Then it creates a `.tar.gz` file with cluster name and timestamp values.

### Verify you can install your plugin locally: kubectl krew install --manifest=[...] --archive=[...]

```bash
$ kubectl krew install --manifest=plugins/snapshot_api_objects.yaml
Installing plugin: snapshot_api_objects
Installed plugin: snapshot_api_objects
\
 | Use this plugin:
 |      kubectl snapshot_api_objects
 | Documentation:
 |      https://github.com/oguzhan-yilmaz/kubectl-snapshot-api-objects
 | Caveats:
 | \
 |  | Requires:
 |  |   - kubectl neat
 |  |   - yq
 | /
/

```


oguzhan-yilmaz:
I've updated my Github Account with the commit author email. We can run the EasyCLA again

plugins/snapshot_api_objects.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  # 'name' must match the filename of the manifest. The name defines how
```
ahmetb, line 4: Let's remove the boilerplate comments.

plugins/snapshot_api_objects.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: snapshot_api_objects
```
ahmetb, line 5: Is there a reason you picked _ over - ? It's rather not typical plugin naming.

plugins/snapshot_api_objects.yaml
```diff
@@ -0,0 +1,43 @@
...
+spec:
+  # 'version' is a valid semantic version string (see semver.org). Note the prefix 'v' is required.
+  version: "v0.0.5"
+  # 'homepage' usually links to the GitHub repository of the plugin
+  homepage: https://github.com/oguzhan-yilmaz/kubectl-snapshot-api-objects
+  # 'shortDescription' explains what the plugin does in only a few words
+  shortDescription: "Snapshot Kubernetes API Objects"
+  description: |
+    Snapshot all Kubernetes Objects and save them as yaml files in .tar.gz archive.
```
ahmetb, line 14: Does kubectl get-all  plugin let us do this? If so, it might be good to not create another plugin and instead compose them.

plugins/snapshot_api_objects.yaml
oguzhan-yilmaz: Removed with the new commit

plugins/snapshot_api_objects.yaml
oguzhan-yilmaz: This was an oversight, I've updated it to `snapshot-api-objects` now

oguzhan-yilmaz:
> Does kubectl get-all plugin let us do this? If so, it might be good to not create another plugin and instead compose them.

Thanks for the insight @ahmetb. I wasn't aware of the `get-all` plugin and it does a good job at data collection with more CLI options. 

My main goal was to have a way to snapshot API Objects to a point-in-time and then have another way to diff the snapshots to get an idea of what's changed.

So I created this plugin to have the data, and was planning to create a complementary plugin for **diff**ing two snapshots using the [dyff tool](https://github.com/homeport/dyff). Which is a great tool with semantic diff features.

I've decided to collect the two features into this plugin to make it different than `get-all` plugin.

My current plan is:
- utilize `kubectl get-all` for data collection
- add the diff functionality
- Have `collect` and `diff` cmds:
  - `kubectl snapshot-api-objects collect -n default, -r pods,configmaps`
  - `kubectl snapshot-api-objects diff snap1.tar.gz snap2.tar.gz`

Would adding the diff functionality make this plugin acceptable?

ahmetb:
Yeah I think it would be much better if the plugin has a distinguishing feature over a plugin that already exists.

PR #4537, opened 2025-04-07
add kubectl resource-backup plugin

zak905:
kubectl plugin that backs up Kubernetes objects (including CRDs) to the local file system (like Velero). Before saving any resource, the plugin does some additional processing to remove:
* the status stanza if the object has any.
* the server generated fields from the object metadata.
* any field with a null value.

The plugin aims to make the saved objects look like the original creation request.

Usage example: `kubectl resource-backup [<flags>] <resource>` where resource can be any resource kind (single name, lower case). 

The saved object files are named as follow: NAME_TYPE_NAMESPACE.yaml. For example, deployment1_deployment_ns.yaml

if the resource is not namespaced the namespace is omitted.

More details can be found on project README: https://github.com/zak905/kubectl-backup

### DIFF
--- a/plugins/resource-backup.yaml
+++ b/plugins/resource-backup.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   version: v0.1.0
   homepage: https://github.com/zak905/kubectl-resource-backup
-  shortDescription: back up K8 resources (like CRDs) to local disk
+  shortDescription: Save Kubernetes resources to disk
   description: |
     Backs up Kubernetes objects (including CRDs) to the local file system. 
     Before saving any resource, some additional processing is done to remove:
--- a/plugins/resource-backup.yaml
+++ b/plugins/resource-backup.yaml
@@ -1,10 +1,10 @@
 apiVersion: krew.googlecontainertools.github.com/v1alpha2
 kind: Plugin
 metadata:
-  name: backup
+  name: resource-backup
 spec:
   version: v0.1.0
-  homepage: https://github.com/zak905/kubectl-backup
+  homepage: https://github.com/zak905/kubectl-resource-backup
   shortDescription: back up K8 resources (like CRDs) to local disk
   description: |
     Backs up Kubernetes objects (including CRDs) to the local file system. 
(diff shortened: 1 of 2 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
How's it exactly different than eksporter/neat plugins? Just the filename part?

zak905:
Hi @ahmetb, 

I aware of the neat plugin, I even mentioned it in the readme. To compare between this plugin and neat plugin
- the neat plugin does not remove null values
- the neat plugin uses kubectl binary underneath when doing a get (which seems like an anti-pattern to me, a kubectl plugin that calls kubectl), while this plugin uses the go client directly. 
- this plugin does lists only and saves each item as a file (off course this is something that can be achieved using bash tricks and utils like `yq`, but the goal of a plugin is to make our life easier right ? )
- this plugin does not make any assumptions when it comes to fields default values. Even if some fields may have the default as a value, it is not possible to distinguish between a value set by a controller/admission webhook, and a value set explicitly upon creation, and therefore it does not remove fields with default values. 
- this plugin aims to be a lightweight alternative to Velero. The plan is to add some functionalities like generating an archive, and backing a resource cluster wide. 

I am already finding this useful when working with CRDs. For example, when testing a new version of a controller, I like doing a quick backup of the state, to be able to restore in case something goes wrong.

ahmetb:
Makes sense. From a naming perspective we try not to assign broadly applicable verbs to a specific plugin. https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/ let's try to find an alternative name?

ahmetb:
Latter is probably better. You'll see plugins with the same prefix.

zak905:
I would not be surprised, as this is a quite standard name:) Plugin renamed. I also fixed the missing license from the tar balls.

plugins/resource-backup.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: resource-backup
+spec:
+  version: v0.1.0
+  homepage: https://github.com/zak905/kubectl-resource-backup
+  shortDescription: back up K8 resources (like CRDs) to local disk
```
ahmetb, line 8: > Save Kubernetes resources to disk

is better. 

K8 is not a thing.

zak905: Sure

ahmetb: Did you fix this part? This is not showing up in the diff.

zak905: oops sorry, I still don't know what happened. the commit was in my fork's branch but still it did not show up here in the PR. Fixed by rebasing against master and force pushing.

ahmetb:
/lgtm
/approve
Thanks!

zak905:
welcome! thank you for considering the plugin. I will look into Krew release automation asap.

ahmetb:
@zak905 many hundred repos use the over engineered automation so if you have a better proposal, you can probably improve that. 

Other than that, it shouldn't matter much in this repo. There are other tools like Goreleaser capable of generating manifests as well.

PR #4745, opened 2025-07-16
Update: net-forward to 1.1.3

antitree:
This is a minor upgrade to the latest version of net-forward. This adds the following few features:

* An optional --image tag which will let you set custom images (thanks to @killerkenobi)
* Support randomized image values


antitree:
another bad pr. Sorry folks.

PR #4787, opened 2025-08-05
add the openunison-cli plugin, deprecate the oulogin plugin

mlbiam:


ahmetb:
IMO the name is really weird, typically the CLI tools don't end in `-cli`.
Do you really wanna keep that?

mlbiam:
@ahmetb sorry didn't see your comment.  Yes, we have users using it and changing that would cause issues.

ahmetb:
/lgtm
/approve
Naming exception

PR #5699, opened 2026-05-06
Fix inspect.yaml file syntax

irenedo:
The syntax in the merged file is bad and I cannot use the automation to release new versions


ahmetb:

-----

>file `plugins/inspect.yaml` is not a straightforward version bump: could not parse version string `"v1.0.0 # This will be replaced with the actual version during release"`
/kind plugin-update

PR #5761, opened 2026-05-22
Add pvu plugin

kha7iq:
## Summary

This PR adds `pvu` to the Krew index.

`pvu` is a kubectl plugin for checking pod volume space usage, similar to `df`. It helps you see how much space is used and how much is still available for a volume directly from the CLI.

## Why this plugin

This plugin focuses on a gap that is not easy to see directly in Kubernetes today. It gives a quick view of volume usage and available space without needing to jump into extra monitoring or separate tooling.

- Repository: https://github.com/kha7iq/pvu
- Release: v0.1.1
- Manifest: `plugins/pvu.yaml`

pvu demo
## Validation

- [x] Local install tested successfully with Krew.
- [x]  Remote install from the GitHub release assets tested successfully.
- [x]  Plugin executes correctly after install.
- [x]  Release archives include the expected binary and LICENSE file.

###  Notes
`pvu` is a kubectl plugin for inspecting pod volumes. It provides an interactive TUI for browsing pods in a namespace, and it can also print volume details for a specific pod.


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Thanks for the submission, @kha7iq!

Just a heads-up that initial plugin submissions can take a while to get reviewed --- we're currently receiving a high volume of new plugin PRs (including quite a few vibe-coded ones), and reviews are done on a best-effort basis by volunteers. Please refrain from directly pinging the reviewers to check on status.

When a reviewer gets to this, they'll assess whether the plugin is a good fit for the index and whether there are already plugins that cover the same use case.

On that note --- there are a few existing plugins that have meaningful overlap with `pvu`'s pod volume inspection functionality:

-   [`mounts`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/mounts.yaml) --- lists all Pods in the current namespace along with their Volumes and VolumeMounts. Very similar to what `pvu` describes.
-   [`browse-pvc`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/browse-pvc.yaml) --- browse PVC contents from the command line.
-   [`debug-pvc`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/debug-pvc.yaml) --- debug pods with access to PVC-backed volumes; also includes an interactive Bubble Tea TUI mode.

It would be worth trying those out. If `pvu` brings something meaningfully different (especially the interactive TUI for browsing pod volumes that the others don't offer), it'd be helpful to highlight those specific differences in the PR description to help reviewers understand why a new plugin is warranted rather than a contribution to an existing one.

kha7iq:
Thank you @ahmetb for the detailed feedback and for taking the time to look into it.

I checked the plugins you mentioned, and `pvu` is focused on a different part of the problem.

 `mounts` shows the volumes and volume mounts for pods, but `pvu` is meant to show volume space usage  basically a `df`-style view for pod volumes, so you can see how much space is used and how much is still available.

That information is not exposed directly in Kubernetes in an easy way, so right now people usually have to rely on prometheus metrics with grafana dashboard  to check it. `pvu` is meant to make that visible directly from the CLI.

`browse-pvc` and `debug-pvc` are useful too, but they work differently. They create or use a pod to inspect a PVC, while `pvu` is focused on checking available space and usage for the volume itself from the pod side without spinning a new pod to attach volume.

I’ll update the PR description to make that distinction clearer. 
Thanks again for the review and the suggestions.

ahmetb:
sounds good!

plugins/pvu.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: pvu
+spec:
+  version: "v0.1.1"
+  homepage: https://github.com/kha7iq/pvu
+  shortDescription: Inspect Kubernetes pod volumes
```
ahmetb: if it's focused on size, can we say size/usage maybe?

kha7iq: Sure. 
I have updated the description.

omerXfaruq (approved):

ahmetb:
/lgtm
/approve
Thanks!

PR #5933, opened 2026-06-29
Add kxdiff plugin

fevzisahinler:
Adds kxdiff — a read-only diff between two Kubernetes environments (context + namespace), as a kubectl plugin.

Homepage: https://github.com/fevzisahinler/kxdiff
Release: v1.0.0

Validated locally with `kubectl krew install --manifest=plugins/kxdiff.yaml` — installs and runs as `kubectl kxdiff`.

### DIFF
--- a/plugins/kxdiff.yaml
+++ b/plugins/kxdiff.yaml
@@ -1,0 +1,67 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kxdiff
+spec:
+  version: v1.0.0
+  homepage: https://github.com/fevzisahinler/kxdiff
+  shortDescription: Read-only diff between two Kubernetes environments
+  description: |
+    kxdiff compares two live Kubernetes environments — a context and a namespace
+    on each side — and reports what differs, as text, JSON or markdown.
+
+    It discovers every resource type the cluster serves (built-ins, CRDs and
+    custom resources), filters out server-managed noise (resourceVersion, uid,
+    managedFields, status, controller-owned objects, high-churn types), masks
+    Secret values, and matches list items by name so reordering is not a false
+    diff.
+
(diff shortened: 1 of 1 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Sorry this looks exactly like the ctx-diff plugin. 
/hold

fevzisahinler:
@ahmetb you're right, there's real overlap with ctx-diff. Thanks for the pointer, I hadn't come across it. I went through its source, and here's where kxdiff actually differs (you can judge if it's enough):

- List matching by name, not index. ctx-diff flattens lists positionally (spec.containers[0], [1], ...), so reordering env vars or containers, or inserting one, shows up as a false diff. kxdiff matches list items by their name field, so only real changes show.
- CI drift gate. kxdiff has a stable exit-code contract (0 = no diff, 1 = diff, 2 = error) plus --quiet. ctx-diff returns 0 even when there are differences, so it can't gate a pipeline.
- Multiline values. ctx-diff collapses long or multiline values to a hash. kxdiff shows a line-level diff of the changed lines, e.g. a ConfigMap data block.
- Noise filtering. kxdiff skips any object with ownerReferences (operator or controller created), not just a fixed pods/replicasets/events list.
- Markdown output for PRs and chat, on top of text and JSON.

The core idea (two-environment diff, CRD-aware, secret-safe) does overlap, so I get it if this is too close for the index. Happy to defer to you on that. And if there's anything you'd want changed, in the manifest or the plugin itself, I'm glad to do it.

ahmetb:
I recommend contributing these to ctx-diff, they seem reasonable to have. 


fevzisahinler:
Makes sense, thanks for the quick review and for pointing me to custom indexes. Closing this out. Appreciate your time.

PR #5934, opened 2026-06-29
feat: add kubectl-rename plugin v0.1.0

akashbhujbalwebsite:
### What does it do?

`kubectl rename` safely renames Kubernetes ConfigMaps and Secrets by creating a copy under the new name and deleting the original.

### Why is it useful?

Kubernetes has no built-in way to rename a ConfigMap or Secret. Users currently have to manually `kubectl get`, edit the YAML, `kubectl apply`, then `kubectl delete` — and risk breaking dependent workloads if they forget to check what references the resource.

This plugin makes that workflow safe:
- **Pre-flight permission check** before touching anything
- **Dependency scanner** — shows which Pods and Deployments reference the resource
- **`--dry-run` mode** — preview the plan without making changes
- **Partial-failure recovery** — re-running detects an interrupted rename and completes it
- **Confirmation prompt** (bypass with `-y` for scripts/CI)

### Usage

```bash
# Rename a ConfigMap
kubectl rename configmap old-name new-name -n my-namespace

# Rename a Secret
kubectl rename secret old-secret new-secret -n my-namespace

# Dry run — preview without changes
kubectl rename configmap old-name new-name --dry-run

# Skip confirmation prompt
kubectl rename configmap old-name new-name -y
```

### Plugin details

| Field | Value |
|-------|-------|
| Repository | https://github.com/akashbhujbalwebsite/kubectl-rename |
| Version | v0.1.0 |
| License | MIT |
| Platforms | linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, windows/amd64, windows/arm64 |

### Checklist
- [x] Plugin manifest added at `plugins/rename.yaml`
- [x] All platform binaries published in GitHub release v0.1.0
- [x] SHA256 checksums verified from `checksums.txt`
- [x] `--dry-run` supported
- [x] README with usage examples in repo
- [x] MIT License

PR #5958, opened 2026-07-02
feat: add safe-rename plugin v0.2.0

akashbhujbalwebsite:
### What it does

`kubectl safe-rename` renames Kubernetes ConfigMaps and Secrets — the only resources where renaming is safe. These resources are stateless data stores whose names carry no runtime identity; Kubernetes imposes no live binding on their names at the API level.

Other resources (Pods, Deployments, Services, ServiceAccounts, PVCs) are intentionally not supported — their names are tied to DNS, pod identity, or runtime bindings where a create+delete cycle would break running workloads.

### Why it is useful

Kubernetes has no built-in way to rename a ConfigMap or Secret. Users currently have to manually `kubectl get`, edit the YAML, `kubectl apply`, then `kubectl delete` — with no safety checks, no dependency scan, and no recovery if interrupted halfway.

### Features

- **Pre-flight permission check** — verifies get+create+delete before touching anything
- **Dependency scanner** — shows which Pods and Deployments reference the resource
- **`--dry-run` mode** — preview the full rename plan without making changes
- **Partial-failure recovery** — re-running detects an interrupted rename and completes it
- **Confirmation prompt** — bypass with `-y` for scripts/CI

### Usage

```bash
kubectl safe-rename configmap old-name new-name -n my-namespace
kubectl safe-rename secret old-secret new-secret -n my-namespace
kubectl safe-rename configmap old-name new-name -n my-namespace --dry-run
```

### Note on previous submission

This supersedes PR #5934 (`kubectl rename`) which was closed. The plugin has been renamed to `safe-rename` to make the intentional resource scope explicit.

Homepage: https://github.com/akashbhujbalwebsite/kubectl-safe-rename

### DIFF
--- a/plugins/safe-rename.yaml
+++ b/plugins/safe-rename.yaml
@@ -1,0 +1,93 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: safe-rename
+spec:
+  version: v0.2.0
+  homepage: https://github.com/akashbhujbalwebsite/kubectl-safe-rename
+  shortDescription: Safely rename ConfigMaps and Secrets
+  description: |
+    kubectl safe-rename renames ConfigMaps and Secrets. These resources are
+    stateless data stores whose names carry no runtime identity — Kubernetes
+    imposes no live binding on their names at the API level. Other resources
+    such as Pods, Deployments, Services, ServiceAccounts, and PVCs are
+    intentionally not supported: their names are tied to DNS, pod identity,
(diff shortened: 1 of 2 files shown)


ahmetb:

-----

Thanks for your patience!
/kind new-plugin

ahmetb:
Sorry, I dont think this is a good fit. Renaming secrets/configmaps is definitely not safe as it would block pod startup for any workload resource referencing these as mounts etc. Also the plugin is incredibly specific to some resources for no apparent reason. Also, this is just not a good resource management model at all. 
/close

PR #6092, opened 2026-07-19
Add new plugin: idx

jzills:
Adds **kx** — a kubectl wrapper with index-based resource selection.

Run `kubectl kx get pods` once to list resources with numbered indexes, then reference any resource by number in follow-up commands (`describe`, `logs`, `exec`, `delete`, `edit`, `scale`, `rollout`, `port-forward`, `events`, `yaml`, `tree`, …). A navigable history of listings keeps indexes resolving across resource kinds and namespaces.

- Homepage: https://github.com/jzills/kx
- Platforms: linux/amd64, linux/arm64, darwin/amd64, darwin/arm64 (self-contained PyInstaller builds; no runtime dependencies beyond kubectl itself)
- Archives are built and checksummed by CI on each release ([release workflow](https://github.com/jzills/kx/blob/main/.github/workflows/release.yml)); future versions will be submitted via krew-release-bot.
- Verified locally: manifest sha256s match the release's SHA256SUMS, and the `bin` path (`kx/kx`) extracts and runs (`kx 0.0.7`).

About the name: `kx` is short for "kubectl, indexed" — the plugin's whole purpose is minimizing keystrokes, so a compact name mirrors existing short plugins like `ctx` and `ns`.

### DIFF
--- a/plugins/idx.yaml
+++ b/plugins/idx.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: idx
 spec:
-  version: v0.0.7
+  version: v0.0.8
   homepage: https://github.com/jzills/kx
   shortDescription: Select resources by index instead of typing names
   description: |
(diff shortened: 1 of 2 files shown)


ahmetb:
Thanks for submitting **kx** to the Krew index! A human reviewer will make the final decision on whether this plugin fits the curated index.

> Due to the increased rate of plugin submissions driven by agentic coding tools, we're seeing a lot of pressure on our small reviewer community. Please do not tag maintainers directly — we'll get to your plugin eventually.

Here's my initial review with a few items to address:

---

### 1. Similar plugins — please clarify the difference

There are a couple of existing plugins that overlap significantly in functionality:

- **[ipick](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/ipick.yaml)** — "A kubectl wrapper for interactive resource selection." Lets you pick the resource to act on interactively (e.g. `kubectl ipick exec` shows a list of pods and executes `kubectl exec <selected-pod>`).
- **[fuzzy](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/fuzzy.yaml)** — "Fuzzy and partial string search for kubectl." Uses fzf-like fuzzy-finder to select resources from an interactive list instead of typing full names.

Could you try out these plugins and leave a comment explaining how **kx** differs? The index-based (numbered) approach and navigable history are interesting, but we want to make sure the distinction is clear for users.

---

### 2. Naming — consider a more descriptive name

`kx` is very compact, but the `x` doesn't naturally communicate "indexed." Existing short names like `ctx` (context) and `ns` (namespace) are immediately recognizable abbreviations of well-known kubectl concepts, whereas `kx` requires the reader to already know what the plugin does. Consider whether a name like `ix` (indexed) or something that more clearly conveys "select by index" would be more discoverable.

Additionally, per the [naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/), the `k` prefix is something we generally suggest avoiding (e.g. prefer `debug` over `kdebug`). Since `kubectl kx` already starts with `k`, the leading `k` in the plugin name is somewhat redundant. This isn't strictly enforced, but it's worth considering.

---

### 3. Caveats — remove usage/alias instructions

The `caveats` field currently includes:

```
For the short form, add:
  alias kx="kubectl kx"
```

Krew already instructs users to run `kubectl <plugin>` and links to your homepage. Usage instructions like alias suggestions should live in your README/homepage rather than in the manifest. Please remove this from the `caveats` field.

---

Once these items are addressed (or clarified), a human reviewer will take a final look. Thanks again for your contribution!

/hold

<sup> :robot: *This comment is an automatic review provided by [`krew-review-agent`(https://github.com/ahmetb/krew-review-agent)].

jzills:
Thanks for the thorough review! All three points addressed:

**1. Difference from ipick/fuzzy:** Both attach an interactive picker to a single command — each invocation shows a selection UI, you pick, and the selection is discarded. idx inverts that model: `kubectl idx get pods` prints a plain, non-interactive listing once and assigns stable numeric indexes; those indexes then resolve across any number of follow-up commands — `idx logs 2`, `idx describe 2`, multi-select like `idx delete 1 3` — with no re-listing or re-selection in between. Selections persist as a navigable history (`idx back`, `idx state`), so earlier listings of other kinds/namespaces stay addressable. Practically: with ipick/fuzzy the selection cost is paid per command and needs TTY interaction; with idx it's paid once per listing and follow-ups are plain typed commands with pipe-friendly output. idx also includes commands with no picker equivalent (ownership `tree`, health `diagnostic`, log aggregation across a workload's pods).

**2. Naming:** Agreed — renamed to `idx`, which reads as "index" directly and drops the redundant `k`. The PR, manifest, and title are updated (`plugins/idx.yaml`).

**3. Caveats:** Removed the alias/usage instructions (and the kubectl mention); caveats now carry only the state-directory disclosure. The alias tip lives in the project README instead.

jzills:
Validation failure addressed: the darwin archives contained symlinks from the PyInstaller bundle (`_internal/Python` → `Python.framework/...`), which krew's extractor rejects ("file type 50"). Archives are now built with `tar --dereference`, and our CI gained a step that extracts each produced archive, fails on any symlink, and runs the binary from the extracted tree — verified on all four platforms.

Manifest updated to v0.0.8, whose assets are symlink-free (confirmed: `tar -tvzf` shows zero link entries).

ahmetb:
/hold cancel
/lgtm
/approve
Thanks!

PR #6123, opened 2026-07-24
Add new plugin: cost-scan

DPS0340:
**What does this plugin do?**

kubeoptimizer is a read-only Kubernetes cost waste scanner. It scans a cluster with get/list access only and reports estimated monthly waste (overprovisioned requests, underutilized nodes, idle GPUs, unused PVs, idle LoadBalancers, zombie workloads, missing requests). No agent, no telemetry, no mutations.

- Repo: https://github.com/DPS0340/kubeoptimizer
- Release: https://github.com/DPS0340/kubeoptimizer/releases/tag/v0.1.0
- License: Apache-2.0

**Checklist**

- [x] Read the [plugin naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/)
- [x] Manifest generated by GoReleaser from the tagged release; sha256 sums match the release assets
- [x] Tested locally with `kubectl krew install --manifest=plugins/kubeoptimizer.yaml` on darwin/arm64
- [x] Plugin is read-only against the cluster (get/list verbs only)


ahmetb:
Thanks for submitting `kubeoptimizer`! Before a human reviewer takes a look, I've done an initial review of your plugin manifest against the Krew plugin guidelines.

> Due to the rise of agentic coding tools, we're seeing an increased rate of plugin submissions, which puts stress on our small reviewer community. Please do not tag maintainers directly — we'll get to your plugin eventually.

Here are the items to address:

### Naming — "kube" prefix

The plugin name `kubeoptimizer` has an unnecessary `kube` prefix. When invoked as `kubectl kubeoptimizer`, the `kube` part is redundant. We suggest removing it for a more natural reading experience.

On its own, `optimizer` would be too generic (and slightly misleading since the plugin is read-only — it scans and reports rather than optimizing). Consider a name that captures the verb and focus more precisely, e.g. `cost-waste-scan`, `waste-scanner`, or `cost-scan`. See the [naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/) for more guidance.

### Short Description — redundant "Kubernetes"

> `shortDescription: Read-only Kubernetes cost waste scanner`

The word "Kubernetes" is unnecessary here — it's already implied since this is a kubectl plugin. Consider trimming it to something like:

```
shortDescription: Read-only cluster cost waste scanner
```

### Overlap with existing plugins

Several existing plugins cover related functionality. Please try them out and clarify in a comment how `kubeoptimizer` differs:

- [`cost`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/cost.yaml) — View cluster cost information (uses Kubecost's API; requires a cluster-side daemon)
- [`popeye`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/popeye.yaml) — Scans your clusters for potential resource issues, including over/under allocations
- [`kor`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/kor.yaml) — Discovers unused Kubernetes resources
- [`crane`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/crane.yaml) — FinOps platform that analyzes cluster state and provides optimization recommendations
- [`lambda-g`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/lambda-g.yaml) — 6D resource imbalance scanner that detects stranded/underutilized resources

In particular, it would help to explain how your plugin differs from `popeye` (which also scans for overprovisioned/underutilized resources read-only) and `kor` (which finds unused resources like idle PVs and services).

---

This is a new plugin submission, so it will require human review before it can be merged. A maintainer will take a look after you've had a chance to address the points above.

/hold

<sup> :robot: *This comment is an automatic review provided by [`krew-review-agent`(https://github.com/ahmetb/krew-review-agent)].

DPS0340:
Thanks for the review — all three points are addressed. Pushed as `2e16489`; the PR now adds a single file, `plugins/cost-scan.yaml`.

## 1. Naming

Renamed to `cost-scan`, which was your suggestion. I agree with the reasoning on both halves: `kubectl kubeoptimizer` stutters, and `optimizer` oversells a tool that only reads and reports. `cost-scan` says what it does and reads naturally as `kubectl cost-scan`.

The GitHub repository, the Go module path, the binary, and the Homebrew cask were all renamed to match, so nothing still refers to the old name. `kubectl cost-scan --help` now prints:

```
cost-scan scans a cluster with get/list access only and reports
estimated monthly waste. It never mutates anything and never phones home.
```

I checked `cost-scan` was free in the index before taking it (`plugins/cost-scan.yaml` did not exist; `cost` does, and I address the overlap below).

## 2. Short description

Now `Read-only cluster cost waste scanner`. Changed in the GoReleaser krew stanza that generates this manifest, not just in the manifest, so regenerating cannot reintroduce it.

## 3. Overlap with existing plugins

I read the source of the closest ones rather than going by their descriptions. The short version: those tools answer *"what is misconfigured or unused?"*, this one answers *"what is that costing me per month?"*

**`popeye`** is a linter, and a good one — it grades ~30 resource kinds. Two concrete differences:

- No pricing model anywhere in `internal/`, so an over-provisioned deployment and an idle GPU node surface as comparable findings. Ranking by dollars is the entire point of this tool.
- Its LoadBalancer rule (`internal/lint/svc.go`, `checkType`) fires on *every* service of type LoadBalancer. Mine fires only when a LoadBalancer has zero ready EndpointSlice endpoints — a billed LB routing to nothing. Different question: "is this type risky?" versus "is this specific LB idle?"

**`kor`** finds unused objects by kind and is object-centric. It has no cost model, and it ships `DeleteResource`/`DeleteResourceWithFinalizer` — it can mutate. This plugin cannot: there is no `Create`/`Update`/`Delete`/`Patch` call anywhere in the codebase (only `List`, 8 call sites), and the shipped RBAC is `[get, list]` only. It also does not look at nodes, GPUs, or LoadBalancers.

**`cost`** (Kubecost) is more accurate on price than I can be, because it reads real billing data — but per its own description it "uses Kubecost's API", so it needs that stack running in-cluster. This trades accuracy for zero install: a kubeconfig is enough. Every finding prints its `CostBasis` (the arithmetic behind the number) so an estimate can be checked rather than trusted.

**`crane`** is the CLI for the Crane FinOps platform, and **`lambda-g`** is a resource imbalance auditor. I have not run either, so I will not characterise them beyond that; both are broader than a single read-only command with no daemon and no state.

The check I have not found elsewhere is idle-GPU detection — GPU nodes with allocatable GPUs that no pod requests. That is usually the largest single line item in a cluster bill. Neither `popeye` nor `kor` mentions `nvidia.com/gpu` anywhere in their source; I have not audited `crane` or `lambda-g` closely enough to make the same claim about them.

Not asking for the hold to be lifted — just noting the points are addressed whenever a reviewer has time.

DPS0340:
Verified the manifest end-to-end before asking for the hold to come off:

```
$ kubectl krew install --manifest=cost-scan.yaml
Installed plugin: cost-scan

$ kubectl cost-scan version
cost-scan 0.4.0 (commit 4032bd7, built 2026-07-26T05:36:34Z, darwin/arm64)
```

`sha256` for `darwin_arm64` in the manifest matches the published release
asset (`9e5f35f7...fd3df`), and `deploy/rbac.yaml` referenced in the caveats
resolves on the default branch.

The three review points are addressed in the comment above.

/hold cancel

ahmetb:
you've merged your last plugin only 3 days ago, so I hope you spend a bit more time honing these plugins instead of serially creating new ones, but approving since this one looks ok.
/lgtm
/approve

PR #6148, opened 2026-08-04
Add costwatch plugin

ArpitRajputGithub:
This adds `costwatch`, a kubectl plugin that prices requested CPU/memory/storage against a bundled price table and diffs against the previous run to surface cost drift per namespace — no backend, agent, or cloud credentials.

- Homepage: https://github.com/ArpitRajputGithub/kubectl-costwatch
- Release: v0.1.0 (linux/darwin/windows × amd64/arm64, sha256 pinned)

Tested locally against the release artifacts:

```
kubectl krew install --manifest-url=https://raw.githubusercontent.com/ArpitRajputGithub/kubectl-costwatch/main/.krew.yaml
kubectl costwatch
```

### DIFF
--- a/plugins/costwatch.yaml
+++ b/plugins/costwatch.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   version: v0.1.0
   homepage: https://github.com/ArpitRajputGithub/kubectl-costwatch
-  shortDescription: Show which namespaces got more expensive since the last run.
+  shortDescription: Track cost drift per namespace since the last run
   description: |
     Prices requested CPU/memory/storage against a bundled price table and diffs
     against the previous run to show cost drift per namespace — no backend.
--- a/plugins/costwatch.yaml
+++ b/plugins/costwatch.yaml
@@ -1,0 +1,36 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: costwatch
+spec:
+  version: v0.1.0
+  homepage: https://github.com/ArpitRajputGithub/kubectl-costwatch
+  shortDescription: Show which namespaces got more expensive since the last run.
+  description: |
+    Prices requested CPU/memory/storage against a bundled price table and diffs
+    against the previous run to show cost drift per namespace — no backend.
+  platforms:
+    - selector: { matchLabels: { os: linux, arch: amd64 } }
+      uri: https://github.com/ArpitRajputGithub/kubectl-costwatch/releases/download/v0.1.0/kubectl-costwatch_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for submitting the `costwatch` plugin! This is a new plugin submission, so it will require human review before it can be merged.

> Due to agentic coding tools, we see an increased rate of plugin submissions. This puts a stress on the small reviewer community. Please do not tag maintainers directly — we'll get to your plugin eventually.

Here are a few things to address before the human review:

### 1. `shortDescription` exceeds the 50-character limit

The current value:

> Show which namespaces got more expensive since the last run.

…is 60 characters. Krew requires `shortDescription` to be **50 characters or less**, and it should be a tagline rather than a full sentence. Consider something like:

> Track cost drift per namespace since the last run

(49 chars) or a similarly concise alternative.

### 2. Potential overlap with existing cost plugins

There are existing plugins in the cost-analysis space that you should try out and clarify how `costwatch` differs:

- [`cost`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/cost.yaml) — *View cluster cost information.* Uses the Kubecost API (requires a cluster-side daemon) to gather and format cost data by namespace, deployment, and label.
- [`cost-scan`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/cost-scan.yaml) — *Read-only cluster cost waste scanner.* Scans a cluster with get/list access only and reports estimated monthly waste; no agent, no telemetry, no mutations.

Could you try out both and leave a comment explaining how `costwatch` is different — particularly how the "diff against the previous run" workflow and the bundled price table set it apart from the above?

For naming guidance in general, you can refer to https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/.

/hold

ArpitRajputGithub:
Thanks for the review! I've shortened `shortDescription` to 49 chars ("Track cost drift per namespace since the last run").

On the overlap with the existing cost plugins — here's how `costwatch` differs from each:

- **`cost`** depends on the Kubecost API, so it needs Kubecost running in-cluster to return data. `costwatch` has no such dependency: it reads only the requested CPU/memory/storage from objects you already have access to and prices them against a bundled on-demand table, so it works read-only with zero cluster-side install and no credentials.
- **`cost-scan`** is the closest neighbor — also agentless — but it reports a *current* snapshot of estimated waste/over-provisioning. `costwatch`'s distinguishing feature is temporal: it saves each run and **diffs against the previous run**, so the output is the *change* per namespace ("what got more expensive since last time, and why") rather than an absolute figure. It's designed to be run repeatedly to catch cost drift over time, not as a one-shot audit.

Happy to adjust the naming or description further if that positioning still overlaps too much.

ahmetb:
/lgtm
/approve
Thanks!

PR #6149, opened 2026-08-04
Update vigil to v1.5.0

msambare:
Updates the `vigil` plugin ([KubeVigil](https://github.com/stribog-cloud/KubeVigil)) from **v0.5.0 to v1.5.0**.

The entry has been pinned at v0.5.0 for five months and seven releases — the manifest was only ever updated by hand, once. The repository is now wired to `rajatjindal/krew-release-bot`, so subsequent updates will be opened automatically on release; this one-off PR closes the existing gap.

### Checksums

All five platform `sha256` values were produced by krew-release-bot's own renderer from the published release assets, then **independently verified against `kubevigil_checksums.txt` in the v1.5.0 release** — all five match exactly.

### Other changes

- **Description: 110 → 150 checks.** The tool now ships 150 checks across 12 categories; the published description understated it.
- **`caveats` added**, noting that a live-cluster scan uses the current kubeconfig context and pointing at manifest-only (offline) usage.
- **Platform entries re-indented to two spaces**, matching the output krew-release-bot generates. This is why the diff looks larger than a version bump — doing it now means every future automated PR is a version-only diff rather than a reformat.

No change to the plugin name, homepage, `shortDescription`, binary names, or the set of supported platforms (linux/darwin amd64+arm64, windows amd64).

### DIFF
--- a/plugins/vigil.yaml
+++ b/plugins/vigil.yaml
@@ -3,48 +3,52 @@ kind: Plugin
 metadata:
   name: vigil
 spec:
-  version: v0.5.0
+  version: v1.5.0
   homepage: https://github.com/stribog-cloud/KubeVigil
   shortDescription: Scan clusters for misconfigurations
   description: |
     KubeVigil scans Kubernetes clusters and YAML manifests for security
-    misconfigurations. It runs 110 security checks across 12 categories,
+    misconfigurations. It runs 150 security checks across 12 categories,
     maps findings to compliance frameworks (CIS, MITRE ATT&CK, NSA/CISA),
     and outputs reports in 8 formats. Includes auto-remediation with a
     five-ring safety model.
+  caveats: |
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for the update! This looks like a straightforward version bump with a minor description adjustment (110 → 150 checks) and an added `caveats` field.

One small suggestion for a future update: the `caveats` field currently includes a usage string (`kubectl vigil scan --help`). Krew already directs users to run `kubectl vigil` and links to the plugin's homepage, so usage strings in the manifest fields are generally discouraged. Consider keeping the caveat to just the behavioral note about kubeconfig context, and moving the offline-usage hint to the project's README.

Otherwise, everything looks good — version is moving forward, all URIs and checksums point to the same repository, and the platform set is unchanged.

/lgtm
/approve
/kind plugin-update

PR #6167, opened 2026-08-10
Add kubesplaining plugin (v1.2.0)

0hardik1:
This adds **kubesplaining**, a Kubernetes security scanner whose differentiator is an RBAC privilege-escalation graph: it searches from every RBAC subject toward sinks such as cluster-admin, kube-system secrets, node escape, and token minting, and reports the full hop chain for each route, including whether the route survives the recommended fix. Further analyzers cover overbroad RBAC, pod security, NetworkPolicy coverage, admission webhooks, secrets exposure, service accounts, certificate-signing abuse, and EKS IAM mappings. It scans live clusters or offline snapshots and emits scored findings as HTML, JSON, CSV, or SARIF.

- Source: https://github.com/0hardik1/kubesplaining
- Version: v1.2.0 (Apache-2.0), archives built by GoReleaser with sha256 checksums
- Platforms: darwin/{amd64,arm64}, linux/{amd64,arm64}, windows/amd64

Checklist from the template:

- I have read the Plugin Naming Guide. On the "avoid repeating kube" point: kubesplaining is the tool's established name (a wordplay the project is known by), consistent with existing index entries such as kubescape and kubesec-scan.
- Verified locally with `kubectl krew install --manifest=plugins/kubesplaining.yaml`: the archive downloads, the sha256 verifies, and the installed binary runs (`version` and an offline `scan-resource` were exercised on darwin/arm64).

### DIFF
--- a/plugins/kubesplaining.yaml
+++ b/plugins/kubesplaining.yaml
@@ -28,11 +28,6 @@ spec:
     policy APIs it inspects. Forbidden resources are skipped and recorded
     as collection warnings, so partial scans of locked-down clusters
     still work.
-
-    Quick start:
-      kubectl kubesplaining scan
-      kubectl kubesplaining download -o snapshot.json
-      kubectl kubesplaining scan --input-file snapshot.json
   platforms:
   - selector:
       matchLabels:
--- a/plugins/kubesplaining.yaml
+++ b/plugins/kubesplaining.yaml
@@ -1,0 +1,96 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kubesplaining
+spec:
+  version: v1.2.0
+  homepage: https://github.com/0hardik1/kubesplaining
+  shortDescription: Scan for security risks and RBAC escalation paths
+  description: |
+    Kubesplaining scans a Kubernetes cluster (live, or offline from a
+    snapshot file) and emits scored security findings as HTML, JSON, CSV,
+    or SARIF.
+
+    Its differentiator is an RBAC privilege-escalation graph: it searches
+    from every subject toward sinks such as cluster-admin, kube-system
+    secrets, node escape, and token minting, and reports the full hop
+    chain for each route it finds, including whether the route survives
+    the recommended fix.
+
+    Additional analyzers cover overbroad RBAC grants, pod security
+    (privileged containers, host namespaces, hostPath), NetworkPolicy
+    coverage, admission webhook weaknesses, secret and ConfigMap
+    exposure, service-account token handling, certificate-signing abuse,
+    and EKS-specific IAM mappings. Raw Secret values are never read, and
+    ConfigMap values are redacted at collection time.
+  caveats: |
+    This plugin needs read (get/list) access to the RBAC, workload, and
+    policy APIs it inspects. Forbidden resources are skipped and recorded
+    as collection warnings, so partial scans of locked-down clusters
+    still work.
+
+    Quick start:
+      kubectl kubesplaining scan
+      kubectl kubesplaining download -o snapshot.json
+      kubectl kubesplaining scan --input-file snapshot.json
+  platforms:
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: arm64
+    uri: https://github.com/0hardik1/kubesplaining/releases/download/v1.2.0/kubesplaining_1.2.0_Darwin_arm64.tar.gz
+    sha256: 27b1a495ac3ab8d1c175bf5b4c841c4de581c958f115ef28109f6557e00200f4
+    bin: kubesplaining
+    files:
+    - from: kubesplaining
+      to: .
+    - from: LICENSE
+      to: .
+  - selector:
+      matchLabels:
+        os: darwin
+        arch: amd64
+    uri: https://github.com/0hardik1/kubesplaining/releases/download/v1.2.0/kubesplaining_1.2.0_Darwin_x86_64.tar.gz
+    sha256: c1dc4121507e46cc37981303150588e263c4fcc7e93851d6a6d0e46db2b489d2
+    bin: kubesplaining
+    files:
+    - from: kubesplaining
+      to: .
+    - from: LICENSE
+      to: .
+  - selector:
+      matchLabels:
+        os: linux
+        arch: arm64
+    uri: https://github.com/0hardik1/kubesplaining/releases/download/v1.2.0/kubesplaining_1.2.0_Linux_arm64.tar.gz
+    sha256: 0d51cdb825155b05583bc8cd2b1c0a405a9fe5314b2434f2ebb77f4835a7254a
+    bin: kubesplaining
+    files:
+    - from: kubesplaining
+      to: .
+    - from: LICENSE
+      to: .
+  - selector:
+      matchLabels:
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for submitting **kubesplaining**! This is a new plugin submission, so it will require human review before it can be merged.

> Due to the increased rate of plugin submissions driven by agentic coding tools, our small reviewer community is under significant stress. **Please do not tag maintainers directly** — we'll get to your plugin eventually.

I've completed an initial review and have the following feedback:

---

### Issues to Address

**1. Usage strings in `caveats`**

The `caveats` field currently includes a "Quick start" section with example commands:

```
Quick start:
  kubectl kubesplaining scan
  kubectl kubesplaining download -o snapshot.json
  kubectl kubesplaining scan --input-file snapshot.json
```

Krew already instructs users to run `kubectl <plugin>` and links to your plugin's `homepage` for documentation, so usage examples should not be included in the manifest. Please remove this "Quick start" block. The first paragraph of your caveats (about required RBAC read access and forbidden-resource handling) is exactly the kind of note that belongs there.

**2. Naming — "kube" prefix**

While the strict naming rule only prohibits `kube-` with a hyphen (which your name doesn't use), the [Plugin Naming Guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/) recommends avoiding redundant "kube" in plugin names since these are already kubectl plugins. We understand this is an established wordplay and there are precedents like `kubescape` and `kubesec-scan`. This isn't a blocker, but we'd encourage you to consider whether an alternative name (e.g., `splaining`, `rbac-trace`, or similar) might be even better. The human reviewer can discuss this further.

---

### Similar Existing Plugins

Before the human review, please try out the following existing plugins and clarify in a comment how **kubesplaining** differs from each:

- **[kubescape](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/kubescape.yaml)** — Scan resources and cluster configs against security frameworks
- **[kubesec-scan](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/kubesec-scan.yaml)** — Scan Kubernetes resources with kubesec.io
- **[vigil](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/vigil.yaml)** — Scan clusters for misconfigurations (150 security checks, 12 categories, compliance mapping)
- **[clusterproof](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/clusterproof.yaml)** — Scan workloads for security posture and evidence (SARIF output)
- **[rbac-tool](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/rbac-tool.yaml)** — Analyze RBAC permissions and generate policies
- **[who-can](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/who-can.yaml)** — Shows who has RBAC permissions to access Kubernetes resources

/hold

0hardik1:
Thanks for the initial review. The "Quick start" usage block has been removed from caveats in 340b080; the RBAC read-access note stays as the only caveat.

On naming: we'd prefer to keep kubesplaining since it is the tool's established name (the binary, repo, and docs all use it) and it follows the existing kubescape / kubesec-scan precedent, but happy to discuss if the reviewer feels strongly.

0hardik1:
Good call to check for overlap. I installed all six via `kubectl krew install` and ran them against one throwaway kind cluster seeded with RBAC privilege-escalation fixtures (co-located service accounts, rolebinding-write grants, impersonation, cluster-admin bindings). Summary of what each does and where kubesplaining sits relative to it:

The six fall into two groups, and kubesplaining overlaps with both without duplicating either.

**Config / posture scanners** (`kubescape`, `kubesec-scan`, `vigil`, `clusterproof`) evaluate resources against a rule or framework catalog, one object at a time.

- **kubescape** — framework/compliance scanner (NSA, MITRE, CIS). My `scan framework nsa` run produced per-control results such as "Administrative Roles", "Allow privilege escalation", "Automatic mapping of service account". Each control is an independent per-resource check mapped to a compliance clause. It reports that a cluster-admin binding exists and that a pod allows privilege escalation; it does not connect a specific non-admin subject to cluster-admin through intermediate steps.
- **kubesec-scan** — kubesec.io risk scoring of a single workload object (`deployment`/`pod`/`daemonset`/`statefulset`). Scope is the workload securityContext only. No RBAC, no cluster-wide view.
- **vigil (KubeVigil)** — KSPM scanner, 153 checks including a genuine RBAC family (`rbac-escalation-verbs`, `rbac-secret-access`, `rbac-cluster-admin`, `rbac-wildcard-*`). Each check is per-object, though. Pointed at my privesc-fixtures namespace (which is full of chainable service accounts) it returned a 97/100 posture score with the top risk being `network-policy-missing`; it flags the individual risky grants but never assembles them into a path.
- **clusterproof** — manifest posture scanner whose distinctive output is signed evidence bundles / SARIF for attestation. It has `CP-RBAC-001..007`; e.g. `CP-RBAC-003` "Workload creation privilege" flags "this subject can create workloads" with remediation "restrict workload creation". Same per-grant model, so it does not chain create-pod into token theft into cluster-admin. Its niche (evidence/attestation) is orthogonal to kubesplaining.

**RBAC query / analysis tools** (`rbac-tool`, `who-can`) answer point questions about permissions.

- **rbac-tool** — the closest neighbor and genuinely complementary. `analysis` flags risky principals per rule (Secret Readers, etc.) and `visualize` renders a subject→binding→role relationship graph. But that graph is a static wiring diagram of who holds what; it is not an attacker-reachability search with defined sinks and per-hop cost. rbac-tool also does policy generation and lookup, which kubesplaining does not.
- **who-can** — answers exactly one query: which subjects can VERB TYPE (e.g. `create pods`). One verb, one resource, one hop. It is a lookup primitive; kubesplaining effectively runs the transitive closure of many such queries and stitches the answers into named routes.

**What none of the six produce, and kubesplaining's reason to exist:** a multi-hop RBAC reachability graph. kubesplaining searches from every non-system subject toward named sinks (`cluster-admin`, `kube-system` secrets, `node-escape`, `token-mint`, `system:masters`, and an `aws_iam_role` sink from IRSA/aws-auth), emits one finding per (source, sink) with the ordered hop chain, attenuates severity by summed per-hop difficulty, and then re-runs the search with the binding its own recommended fix would cut removed, so each finding states whether the path survives that fix (`privesc:survives-first-cut`).

Concrete example from the test cluster, a route none of the six reported as a route:

```
ServiceAccount deepchain-tenant/deepchain-binder  ->  cluster-admin  (3 hops)
  1. modify_role_binding      create/update/patch rolebindings in deepchain-tenant
  2. colocated_sa_token_theft steal token of co-located SA deepchain-privileged
  3. wildcard_permission      deepchain-privileged is bound to cluster-admin (*:*:*)
```

Each hop in isolation is a finding the per-object scanners above would emit separately (a role with rolebinding-write; a co-located privileged SA; a cluster-admin binding). The thing that matters to a defender, that `deepchain-binder` can reach cluster-admin by composing all three, only appears when you chain them.

Happy to fold a short version of this into the plugin description if that would help future reviewers.

fenio:
I know my opinion doesn't carry any weight here, but it seems to me that kubesplaining looks really useful and unique, and I'd love to see it available in Krew.

ahmetb:
/lgtm
/approve
Thanks!

PR #6213, opened 2026-08-18
Add safe-rollout plugin

HarnageaGabriel:
Submitting `safe-rollout`, a plugin that explains why a rollout failed or stalled.

- `kubectl safe-rollout check <kind>/<name>` — pre-flight analysis of a live workload and the state of its namespace.
- `kubectl safe-rollout watch <kind>/<name>` — observes a rollout through the Watch API and, when it fails or stalls, classifies the cause from the evidence collected and proposes remediation.

The plugin only reads from the cluster; it has no mutating operation and no `--apply-fix`. Classification is deterministic and uses no language model: when the evidence confirms a rollout is stuck but cannot distinguish between the known causes of a category, the finding is reported with an explicit `*-undetermined` cause listing what was observed, rather than the most likely cause presented as certain.

Repository: https://github.com/HarnageaGabriel/kubectl-safe-rollout
Release: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/tag/v0.2.0
License: Apache 2.0

### Naming

`safe-rollout` follows the naming guide: lowercase, hyphen-separated, no `kube-` prefix, and specific rather than a bare verb. It does not collide with an existing plugin name in the index.

### Verification performed

- `validate-krew-manifest` reports all 5 `spec.platforms` install cleanly.
- Installed locally from this manifest and run: `kubectl krew install --manifest=safe-rollout.yaml` followed by `kubectl safe-rollout version`, which reports `0.2.0` with the release commit.
- Each release archive contains the binary plus `LICENSE` and `README.md`.
- The published `sha256` values were taken from the release `checksums.txt` and verified against the downloaded archives.

The tool itself is exercised by 17 end-to-end scenarios against a kind cluster (Kubernetes v1.36.1, containerd 2.2), one per classified cause, plus one that asserts a slow-starting application completes with no finding at all.

### DIFF
--- a/plugins/safe-rollout.yaml
+++ b/plugins/safe-rollout.yaml
@@ -1,0 +1,50 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: safe-rollout
+spec:
+  version: v0.2.0
+  homepage: https://github.com/HarnageaGabriel/kubectl-safe-rollout
+  shortDescription: Explain why a rollout failed or stalled
+  description: |
+    Use check for pre-flight analysis of a live workload and its namespace.
+    Use watch to observe a rollout and classify its failure cause. The tool
+    never modifies cluster state. Classification is deterministic; ambiguous
+    evidence is reported as an explicit *-undetermined cause with the observed
+    evidence, rather than as a guess presented as certain.
+  platforms:
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/download/v0.2.0/kubectl-safe_rollout_v0.2.0_linux_amd64.tar.gz
+      sha256: 1813299a28b771fae5daf5ffc8e15195336b8de3ee3ae47c6a800df950cd4d5f
+      bin: kubectl-safe_rollout
+    - selector:
+        matchLabels:
+          os: linux
+          arch: arm64
+      uri: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/download/v0.2.0/kubectl-safe_rollout_v0.2.0_linux_arm64.tar.gz
+      sha256: 34c5ae5ddd41896cc06f74ea1fc24e99609c13d91ebe5b26634af17039166adf
+      bin: kubectl-safe_rollout
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/download/v0.2.0/kubectl-safe_rollout_v0.2.0_darwin_amd64.tar.gz
+      sha256: 87c4daa03aa429ada8e40a64e75b3816bf832c32e3372287ad06c350ebd0e6d1
+      bin: kubectl-safe_rollout
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: arm64
+      uri: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/download/v0.2.0/kubectl-safe_rollout_v0.2.0_darwin_arm64.tar.gz
+      sha256: fb3beb4f06d8f22dddcbafa28040a8130b5da3ecdaf133fec5dc7d1ad9af245b
+      bin: kubectl-safe_rollout
+    - selector:
+        matchLabels:
+          os: windows
+          arch: amd64
+      uri: https://github.com/HarnageaGabriel/kubectl-safe-rollout/releases/download/v0.2.0/kubectl-safe_rollout_v0.2.0_windows_amd64.zip
+      sha256: 094545a64b7647cfe8dcbc24a164daeb8a7b7f714c1adadad99fd6069b63427e
+      bin: kubectl-safe_rollout.exe
--- a/plugins/safe-rollout.yaml
+++ b/plugins/safe-rollout.yaml
@@ -3,7 +3,7 @@ kind: Plugin
 metadata:
   name: safe-rollout
 spec:
-  version: v0.2.0
+  version: v0.3.0
   homepage: https://github.com/HarnageaGabriel/kubectl-safe-rollout
   shortDescription: Explain why a rollout failed or stalled
   description: |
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for submitting `safe-rollout`! This is a new plugin submission, so it will require human review before it can be merged.

> Due to agentic coding tools, we see an increased rate of plugin submissions. This puts a stress on the small reviewer community. **Please do not tag maintainers directly** — we'll get to your plugin eventually.

Here's my initial review against the Krew plugin guidelines:

## Manifest Checks

- ✅ `shortDescription` is 41 characters (under the 50-char limit)
- ✅ Plugin name `safe-rollout` is lowercase kebab-case with no `kube-`/`kubernetes-`/`kubectl-` prefix
- ✅ No redundant terms ("plugin", "Kubernetes") in `shortDescription`
- ✅ `version` (`v0.2.0`) is valid semver with no pre-release tags
- ✅ No usage strings in `description` or `caveats`
- ✅ All platform archives originate from the same repository
- ✅ No redundant `files` block

## Potentially Similar Plugins

Before the human review, could you try out the following existing plugins and clarify in a comment how `safe-rollout` differs from each?

- **[watch-rollout](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/watch-rollout.yaml)** — "Monitor deployment rollouts with live progress" — monitors rollouts in real-time with progress bars, pod status, and completion estimates.
- **[why-fail](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/why-fail.yaml)** — "Explain why a pod is failing, in plain language" — diagnoses failing pods and explains the cause, covering common failure modes with deterministic rules.
- **[triage](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/triage.yaml)** — "Fast diagnostics for failed pods" — provides quick diagnostic snapshots for failed pods including status, events, and logs.

Your plugin's focus on **rollout-level** failure classification (as opposed to pod-level) and its deterministic, evidence-based approach with explicit `*-undetermined` reporting seem to differentiate it, but it would help the human reviewers to have your perspective on the distinction.

For naming guidance, you can also refer to https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/.

/hold

HarnageaGabriel:
Thanks for the review. I installed all three plugins from krew and ran them against the same failures on a kind cluster (Kubernetes v1.36.1, containerd 2.2) rather than comparing them from their READMEs. Here is what I found, including where they do the job better than mine.

## The structural difference: rollout-level vs pod-level

`why-fail` and `triage` take a pod name. `safe-rollout` takes a workload and watches the rollout, correlating Pods, ReplicaSets and Events together. That is not a rewording of the same thing — there are failures that have no pod to point at.

The clearest case is a `ResourceQuota` that blocks part of a rollout. Deployment with 3 replicas, namespace quota `pods: "1"`:

```
$ kubectl why-fail quotahit-787b8fdc7d-n7xgx -n cmpq
  quotahit-787b8fdc7d-n7xgx  (namespace: cmpq)
   ✔ LOOKS HEALTHY
  This pod is Running, all its containers are ready and nothing has restarted.
  whyfail has nothing to report.
```

That answer is correct: the one pod that exists is genuinely healthy. But the rollout is stuck at 1 of 3, and the two missing pods were rejected by the quota admission plugin, so they never existed and there is nothing to pass to a pod-level tool.

```
$ kubectl safe-rollout watch deployment/quotahit -n cmpq
HIGH  quota-exceeded           ReplicaSet/quotahit-787b8fdc7d
      cause: ReplicaSet/quotahit-787b8fdc7d cannot create pods: the namespace ResourceQuota is exhausted
      evidence: event: Error creating: pods "quotahit-787b8fdc7d-hwx76" is forbidden: exceeded quota: tight, requested: pods=1, used: pods=1, limited: pods=1
      remediation: increase the namespace ResourceQuota, reduce workload requests, or free capacity ...
        $ kubectl describe resourcequota -n cmpq
```

The signal lives on the ReplicaSet, not on any Pod.

## `why-fail` — the closest one, and genuinely good

For a single failing pod you have already identified, `why-fail` often explains it better than I do. Its readiness explanation was more useful than mine: it named the probe URL, said how long the pod had been running without becoming ready, and pointed at the 0.0.0.0-versus-127.0.0.1 bind as a likely cause. I am not going to pretend otherwise.

Two differences that are not about quality:

1. **Coverage boundary.** On a container crash-looping with a plain non-zero exit it reported `No rule matched this pod`, then offered an optional LLM fallback (`OPENROUTER_API_KEY`). `safe-rollout` has no LLM path at all — by design, not by omission. Where the evidence confirms a rollout is stuck but does not distinguish between the known causes of a category, it emits an explicit `*-undetermined` cause listing what was observed, and stops there.
2. **When it runs.** `why-fail` inspects a pod that is already failing. `safe-rollout watch` follows the rollout through the Watch API from the moment it starts, so it reports the cause as the rollout stalls rather than after someone notices and goes looking for the right pod.

## `triage` — complementary, not overlapping

`triage` produces a diagnostic snapshot for one pod: status, filtered warning events, previous-crash logs, current logs. It aggregates the raw material well and deliberately does not classify a cause or propose a remediation. I would reach for it after a classification, not instead of one.

## `watch-rollout` — same scope, different question

This is the closest in scope, since it is also rollout-level. It answers "how is this rollout progressing" with progress bars, pod counts and an ETA; mine answers "why did it stop". Its output is a live terminal UI and I could not capture it in my non-interactive container, so I am describing it from its help text rather than from output I actually saw.

The distinction I would draw: `watch-rollout` tells you a rollout is failing, `safe-rollout` tells you which of a fixed set of causes it is and what evidence says so.

## What else `safe-rollout` does that these do not

- A pre-flight `check` before the rollout: PodDisruptionBudget headroom against real replica counts, ResourceQuota headroom against the calculated surge, declared requests against real usage from metrics-server.
- Exit code non-zero if and only if a `high` severity finding exists, with stable cause IDs and a JSON output, so it can gate a CI pipeline.
- No cluster mutation of any kind, and no `--apply-fix`.

## Verification

17 end-to-end scenarios run against a real kind cluster, one per classified cause, using real kubelet, scheduler and containerd messages rather than fixtures. One of them asserts *silence*: an application that takes 25 seconds to become ready, failing its readiness probe the whole time, must complete with no finding at all. The README also lists what has **not** been verified — CRI-O, the other supported Kubernetes minors, restricted RBAC — because a diagnosis tool that overstates its coverage is worse than one that reports less.

Happy to answer anything else, and no rush on the review.

HarnageaGabriel:
Updated the manifest to v0.3.0, which fixes two defects a user would have hit:

- a workload reference with more than one slash (`deployment/foo/bar`) was accepted and resolved to a workload named `foo/bar` instead of being rejected;
- a rollout blocked by a namespace `ResourceQuota` repeated the same cause and remediation once per rejected pod, and since the ReplicaSet controller keeps retrying, the output grew without bound. Those now collapse into one finding whose evidence names the rejected pods, capped with an explicit count of what was omitted.

Most of the release is verification rather than features. Since v0.2.0 the suite runs under the race detector, the event-message matching is fuzzed (six targets, no crasher over roughly 28 million executions), the exit code and JSON field names have tests, and the end-to-end suite now runs in full against **three** Kubernetes minor versions — v1.36.1, v1.35.5, v1.34.8 — rather than one, with 18/18 scenarios passing on each and no pattern needing a change.

One of those scenarios runs `check` as a deliberately restricted ServiceAccount against a real API server, to prove that a check which cannot read a resource degrades to a visible `SKIP` instead of failing the run or reporting a clean result.

CRI-O is still unverified and the README says so plainly: the runtime is what produces the messages this tool matches, so that is the largest remaining gap.

Manifest re-validated with `validate-krew-manifest` (all 5 platforms install) and with a real `kubectl krew install --manifest` in a Linux container. No rush on the review.

PR #6242, opened 2026-08-26
Add drain-check plugin

mjavedali:
## Description

`drain-check` is a read-only CLI that gives you a pre-flight safety
report before running `kubectl drain` on a node. Instead of finding out
mid-drain that a PodDisruptionBudget is blocking eviction, or that you
just took down the only replica of something, it flags these risks
up front.

It checks every pod on a node for:
- PodDisruptionBudgets currently allowing zero disruptions
- Single-replica workloads with no PDB protecting them
- Pods relying on local storage (emptyDir/hostPath) that won't follow
  them to a new node
- "Naked" pods with no controller — drain deletes these permanently
  instead of rescheduling them
- Finalizers that could hang eviction, and pods already stuck in
  `Terminating`

It never modifies the cluster — strictly read-only.

## Checklist

- [x] Source is open source (MIT licensed): https://github.com/mjavedali/drain-check
- [x] Tagged a semantic version release: v0.2.0
- [x] Tested locally via `kubectl krew install --manifest=drain-check.yaml`
      and validated output against a live `kind` cluster
- [x] LICENSE file is included in each release archive
- [x] Reviewed the [plugin naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide//)
      and [development best practices](https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/)

## Testing done

```
$ kubectl drain-check kind-control-plane
drain-check report for node "kind-control-plane"

RISK  NAMESPACE           POD                                      REASONS
HIGH  local-path-storage  local-path-provisioner-855c7b7774-d5pc5  single-replica workload with no PodDisruptionBudget
...
```


ahmetb:
Thanks for submitting `drain-check`! Before we dive in, a quick note:

> Due to the rise of agentic coding tools, we've seen an increased rate of plugin submissions, which puts stress on our small reviewer community. Please do not tag maintainers directly — we'll get to your plugin eventually.

This is a new plugin submission, so it will require a human reviewer to make the final call on whether it fits the curated index. In the meantime, here's the automated review:

**Naming & Description**
- ✅ Name `drain-check` is clear, specific, and follows kebab-case conventions. The verb ("check") and context ("drain") make it obvious what the plugin does.
- ✅ `shortDescription` is 49 characters — within the 50-char limit.
- ✅ Description is informative without containing usage strings.

**Manifest**
- ✅ Six platforms covered (darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64, windows/arm64) — all sourced from the same repository.
- ✅ Version `v0.2.0` has no pre-release tags.
- ✅ No redundant `files` block (defaults apply).

**Uniqueness**
I checked against all existing plugins. The closest related ones are:
- [`debug-pdb`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/debug-pdb.yaml) — debugs PodDisruptionBudgets and pod relationships, but doesn't provide a drain pre-flight report.
- [`confirm`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/confirm.yaml) — prompts for confirmation before destructive kubectl commands (including drain), but is a general-purpose guard, not a drain-specific safety analyzer.
- [`slowdrain`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/slowdrain.yaml) — actually performs the drain with delays, a different use case.

`drain-check` fills a distinct niche as a read-only pre-flight safety report for node draining. No duplicates found.

**Minor note:** The file is missing a trailing newline at the end. Not a blocker, but you may want to add one for cleanliness.

Everything looks good from the automated review. A human reviewer will take a look next.

/hold

ahmetb:
you can just `kubectl drain --dry-run=server`? I don't think this fits the general usage pattern of Kubernetes.

mjavedali:
> you can just `kubectl drain --dry-run=server`? I don't think this fits the general usage pattern of Kubernetes. You can still distribute your plugin via Krew using custom indexes: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/ That way users can still install your plugin via Krew, but they'll be downloading the plugin manifest from your repository instead of krew-index.

Fair point on --dry-run=server — for the PDB-blocking case there's genuinely some overlap, no argument there.

Where I think this still earns its keep: single-replica-with-no-PDB. Dry-run won't error on that one, it just quietly dies later — that's actually the exact incident that got me to build this in the first place. Finalizer and already-terminating checks are in the same boat. And it's meant to be run ahead of a maintenance window, not during the drain attempt itself, so it's more of a planning tool than a dry-run replacement

ahmetb:
> single-replica-with-no-PDB.

if it needs availability it needs PDB? sounds a bit like you're solving an artificial problem that's not broadly applicable to the general usage patterns in Kubernetes, hence it's probably not a good fit for this plugin index.

PR #6268, opened 2026-09-02
kuberecord: v0.3.0

yelzhy:
Adds `kuberecord` at v0.3.0.

`kubectl kuberecord` answers questions about recorded Kubernetes state changes —
who changed what, when, and what the object looked like before — reading history a
kuberecord operator streamed to a sink.

The manifest is generated from the release archives by
[`hack/krew-manifest.sh`](https://github.com/kuberecord/kuberecord/blob/v0.3.0/hack/krew-manifest.sh)
and is published as a release asset, covered by `checksums.txt` and by a keyless
cosign signature over it:
https://github.com/kuberecord/kuberecord/releases/tag/v0.3.0


ahmetb:
Thanks for submitting **kuberecord**, @yelzhy! 👋

First, a note on how this will proceed: due to the rise of agentic coding tools, we're seeing a much higher rate of plugin submissions, which puts a lot of stress on our small reviewer community. **Please do not tag maintainers directly** — we'll get to your plugin eventually. This comment is an initial automated review to help get the manifest into shape; a human reviewer will still make the final call on whether the plugin fits the curated index.

A few changes to consider before that final review:

### 1. Usage strings in `description` and `caveats`

Krew already instructs users to run `kubectl kuberecord` and links to your `homepage`, so command invocations don't belong in the manifest text. Please remove the example-command block from the `description`:

```
kubectl kuberecord timeline deploy/checkout -n payments
kubectl kuberecord diff deploy/checkout -n payments --since 1h
kubectl kuberecord get deploy/checkout -n payments --at 2026-08-28T13:00:00Z
kubectl kuberecord blame deploy/checkout -n payments
```

A link to `docs/CLI.md` (which you already include) is a good replacement. Likewise, the `caveats` shouldn't contain `brew install kuberecord/tap/kuberecord` — that's an install instruction for a different distribution channel, not a caveat. The part of the caveats that *does* belong — that the plugin only reads what a kuberecord operator has already recorded, so with no operator running (and no `--source` archive) there is nothing to query yet — is genuinely useful and should stay.

### 2. `shortDescription` — "Kubernetes" is redundant

"Query recorded Kubernetes state changes" is within the 50-character limit, but since every plugin in this index is already a kubectl plugin, the word "Kubernetes" adds no information. Something like `Query recorded state changes` reads cleaner.

### 3. Overlap with existing plugins

Part of curating this index is avoiding near-duplicates. The following existing plugins cover adjacent ground — please try them out and clarify in a comment how kuberecord is different:

- [blame](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/blame.yaml) — Show who edited resource fields (uses `managedFields`)
- [track](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/track.yaml) — Tracking the changes between resource versions
- [history](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/history.yaml) — List & diff revisions of workload resources

From the PR description, your differentiator seems to be durable, operator-recorded history that can be queried even when the original cluster no longer exists — spelling that out explicitly would help the human reviewer.

### 4. Naming (non-blocking)

"kuberecord" doesn't violate the "kube-" prefix rule (there's no dash), and since it's your project's actual name it's a defensible choice. Just be aware that users will type `kubectl kuberecord ...`, and our [naming guide](https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/) suggests avoiding unnecessary k/kube prefixes where possible so the plugin name reads natively after `kubectl`.

Once the requested changes above are addressed, a human reviewer will take the final look at whether the plugin fits the curated index. The `needs-human-review` label added to this PR is how it gets attention — there's no need to ping anyone.

ahmetb:
Yeah AI review is pretty spot on, we recommend no kube- prefix, and you should probably remove usage strings from here and move them to your --help. Other than that LGTM.

yelzhy:
> Yeah AI review is pretty spot on, we recommend no kube- prefix, and you should probably remove usage strings from here and move them to your --help. Other than that LGTM.

Thanks for the review. I've pushed an update to the manifest. Also, if it is possible, I'd prefer to keep the kuberecord name as it's the exact name of the standalone binary and the project ecosystem.

ahmetb:
/lgtm
/approve
Please consider setting up [Krew release automation](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) which helps you skip manually making updates to your Krew manifests on each new version and send a pull request. It’s a GitHub workflow bot that can run every time you push a new tag. Those PRs are automatically approved.

PR #6274, opened 2026-09-03
Add upgrade-check plugin

pvishwa3:

### DIFF
--- a/plugins/upgrade-check.yaml
+++ b/plugins/upgrade-check.yaml
@@ -1,0 +1,113 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: upgrade-check
+spec:
+  version: v0.1.3
+  homepage: https://github.com/runtimez-com/kube-upgrade-check
+  shortDescription: Find what breaks before you upgrade Kubernetes
+  description: |
+    Reports what an upgrade to a target Kubernetes version would break in the
+    cluster you are pointed at: objects still written at an API version the
+    target removes, control-plane and kubelet settings it refuses to start
+    with, in-tree volume plugins that stop mounting, node runtime and version
+    skew, and whether add-ons such as Karpenter, ingress-nginx, cert-manager,
+    Argo CD, CoreDNS and Istio support the version you are moving to.
+
+    It reads the cluster and writes nothing. No account and no agent, and no
+    data leaves the machine.
+
+    Removed APIs are found through managed fields, so an object is caught
+    whatever applied it, including server-side apply, operators and GitOps
+    controllers that leave no last-applied annotation behind.
+
+    Every check that could not run is printed with the reason and a command to
+    run yourself, so a clean report cannot be one where half the checks were
+    silently skipped.
+
+    Examples:
+      kubectl upgrade-check --target 1.34
+      kubectl upgrade-check --target 1.34 -o json
+      kubectl upgrade-check --fail-on high --strict
+  caveats: |
+    Some checks need permissions the built-in view role does not grant:
+
+      * nodes/proxy, to read each node's live kubelet configuration
+      * the /metrics endpoint, for one of the removed-API evidence sources
+
+    Without them the affected checks are reported as unavailable rather than
+    skipped silently. A read-only ClusterRole covering everything is in the
+    repository under rbac/.
+  platforms:
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: amd64
+      uri: https://github.com/runtimez-com/kube-upgrade-check/releases/download/v0.1.3/kube-upgrade-check_0.1.3_darwin_amd64.tar.gz
+      sha256: 887c48d0063aa5099a40e2bfbf427ff3e86b75d78c4ef76889d43659cb5df1b9
+      bin: kube-upgrade-check
+      files:
+        - from: "kube-upgrade-check"
+          to: "."
+        - from: LICENSE
+          to: "."
+    - selector:
+        matchLabels:
+          os: darwin
+          arch: arm64
+      uri: https://github.com/runtimez-com/kube-upgrade-check/releases/download/v0.1.3/kube-upgrade-check_0.1.3_darwin_arm64.tar.gz
+      sha256: 8e2b73a67032f453cc6f4f3c71f8013240f707f9ec7718a8470558910925a090
+      bin: kube-upgrade-check
+      files:
+        - from: "kube-upgrade-check"
+          to: "."
+        - from: LICENSE
+          to: "."
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/runtimez-com/kube-upgrade-check/releases/download/v0.1.3/kube-upgrade-check_0.1.3_linux_amd64.tar.gz
+      sha256: d790dac352e92df9e0334e9121169e82cfa51c97e64975b22e50110cd21d1a43
+      bin: kube-upgrade-check
+      files:
+        - from: "kube-upgrade-check"
+          to: "."
+        - from: LICENSE
+          to: "."
+    - selector:
+        matchLabels:
+          os: linux
+          arch: arm64
+      uri: https://github.com/runtimez-com/kube-upgrade-check/releases/download/v0.1.3/kube-upgrade-check_0.1.3_linux_arm64.tar.gz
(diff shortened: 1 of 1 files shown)


ahmetb:
Hi @pvishwa3, thanks for submitting the `upgrade-check` plugin!

Before the details, a quick note: due to the increased rate of plugin submissions driven by agentic coding tools, our small reviewer community is under a lot of stress. Please do not tag maintainers directly — we'll get to your plugin eventually. Also note that every new plugin requires human approval before it can be added to the curated index, so a maintainer will make the final call after this initial review.

From the initial manifest review, here are a few things to address:

## 1. Remove usage strings from `description`

The `Examples:` block at the end of the description contains command invocations (`kubectl upgrade-check --target 1.34`, etc.). Please remove it — Krew already instructs users to run `kubectl <plugin>` after installing, and links to the plugin's `homepage` for examples and documentation. Keeping usage strings in the manifest duplicates that information for every user. Feel free to move those examples into your repository's README instead.

## 2. Minor: `shortDescription` redundancy

"Find what breaks before you upgrade Kubernetes" is within the 50-character limit, but the word "Kubernetes" is redundant here — every plugin in the index is already a kubectl plugin. A shorter alternative like "Find what breaks before you upgrade" (35 characters) reads just as well.

## 3. Overlap with existing plugins

Part of `upgrade-check`'s functionality (detecting objects that use API versions removed by the target Kubernetes release) sounds similar to existing plugins in the index:

- [`deprecations`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/deprecations.yaml) — *Checks for deprecated objects in a cluster* (explicitly aimed at verifying deprecated objects before upgrading the cluster)
- [`schemagen`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/schemagen.yaml) — *Generate, migrate, and scaffold Kubernetes resources from cluster schema* (its `migrate` subcommand checks manifests for deprecated or removed APIs)

Could you please try these out and leave a comment clarifying how `upgrade-check` differs from them (e.g., your checks for kubelet configuration, in-tree volume plugins, node version skew, and add-on compatibility, and detection via managed fields)? This will really help the human reviewers assess whether the plugin is a good fit for the curated index.

Once these points are addressed, a human reviewer will follow up. Thanks for helping make Kubernetes upgrades safer! 🚀

pvishwa3:
> Hi @pvishwa3, thanks for submitting the `upgrade-check` plugin!
> 
> Before the details, a quick note: due to the increased rate of plugin submissions driven by agentic coding tools, our small reviewer community is under a lot of stress. Please do not tag maintainers directly — we'll get to your plugin eventually. Also note that every new plugin requires human approval before it can be added to the curated index, so a maintainer will make the final call after this initial review.
> 
> From the initial manifest review, here are a few things to address:
> 
> ## 1. Remove usage strings from `description`
> The `Examples:` block at the end of the description contains command invocations (`kubectl upgrade-check --target 1.34`, etc.). Please remove it — Krew already instructs users to run `kubectl <plugin>` after installing, and links to the plugin's `homepage` for examples and documentation. Keeping usage strings in the manifest duplicates that information for every user. Feel free to move those examples into your repository's README instead.
> 
> ## 2. Minor: `shortDescription` redundancy
> "Find what breaks before you upgrade Kubernetes" is within the 50-character limit, but the word "Kubernetes" is redundant here — every plugin in the index is already a kubectl plugin. A shorter alternative like "Find what breaks before you upgrade" (35 characters) reads just as well.
> 
> ## 3. Overlap with existing plugins
> Part of `upgrade-check`'s functionality (detecting objects that use API versions removed by the target Kubernetes release) sounds similar to existing plugins in the index:
> 
> * [`deprecations`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/deprecations.yaml) — _Checks for deprecated objects in a cluster_ (explicitly aimed at verifying deprecated objects before upgrading the cluster)
> * [`schemagen`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/schemagen.yaml) — _Generate, migrate, and scaffold Kubernetes resources from cluster schema_ (its `migrate` subcommand checks manifests for deprecated or removed APIs)
> 
> Could you please try these out and leave a comment clarifying how `upgrade-check` differs from them (e.g., your checks for kubelet configuration, in-tree volume plugins, node version skew, and add-on compatibility, and detection via managed fields)? This will really help the human reviewers assess whether the plugin is a good fit for the curated index.
> 
> Once these points are addressed, a human reviewer will follow up. Thanks for helping make Kubernetes upgrades safer! 🚀
> 
> 🤖 _This comment is an automatic review provided by [`krew-review-agent`](https://github.com/ahmetb/krew-review-agent). If you see a problem with this, please file an issue._ 🤖

Thanks for the review, and for the note about reviewer load — I've kept this to one comment.

Points 1 and 2 are done in the manifest: the `Examples:` block is gone from `description` (the examples now live in the README), and `shortDescription` is now **"Find what breaks before you upgrade"** (35 chars).

On point 3, I installed both plugins and ran them rather than comparing docs.

## `deprecations` (kubepug)

Same cluster (k3s v1.33.5), same target:

```
$ kubectl deprecations --k8s-version v1.34.0
No deprecated or deleted APIs found

$ kubectl upgrade-check --target 1.34
  BREAKS (1)
    HIGH  DaemonSet/ingress-ingress/ingress-nginx-controller — Ingress-NGINX v1.13.3
          does not support Kubernetes 1.34
  COULD NOT SEE (1)
    - control-plane flags — 276 rules not checked
```

Both are right about APIs — this cluster has none. The difference is what else is asked. `upgrade-check` found a real blocker for that upgrade (the running ingress controller does not support 1.34, per the vendor's own table), and reported that 276 control-plane rules could not be evaluated because k3s runs the control plane as one process rather than as static pods.

kubepug answers "which API versions are going away" thoroughly, from Kubernetes API markers. That is one of seven families here; the others are control-plane and kubelet settings the target refuses to start with (382 rules, including feature gates locked to a new default), in-tree volume plugins that stop mounting (17), node runtime and kubelet/kube-proxy skew, add-on compatibility against vendor-published matrices (7 catalogs: Karpenter, ingress-nginx, cert-manager, Argo CD, CoreDNS, Istio, kube-proxy), and version-advisory items no API call can settle (34).

## `schemagen migrate`

This reads YAML and compares against the connected cluster's OpenAPI schema, so it answers "does this manifest use API versions this cluster still serves". It is a manifest tool, and a good one — `upgrade-check` has no equivalent and I point people at file-based tools for pre-merge checks.

Its README suggests `kubectl get deploy -o yaml | kubectl schemagen migrate -`, and that path runs into the conversion problem. I set up a CRD serving `v1beta1` (deprecated) and `v1` (storage), applied an object at `v1beta1` server-side, and compared what each source reports:

```
# 1. what the API server returns, i.e. what a piped manifest tool reads
$ kubectl get widget legacy-widget -o yaml | grep apiVersion
apiVersion: demo.upgradecheck.io/v1          # <- converted, not what was written

$ kubectl get widget legacy-widget -o yaml | kubectl schemagen migrate -
  OK  <stdin> demo.upgradecheck.io/v1/Widget

# 2. the last-applied annotation, which kubent and Pluto read
$ kubectl get widget legacy-widget -o jsonpath='{.metadata.annotations}'
                                              # <- empty; server-side apply writes none

# 3. managedFields
$ kubectl get widget legacy-widget -o jsonpath='{.metadata.managedFields[*]}'
manager=some-operator apiVersion=demo.upgradecheck.io/v1beta1 op=Apply time=2026-09-03T01:56:02Z
```

To be precise about what that does and does not show: schemagen is correct in both runs, since it validates against the schema rather than tracking deprecation lifecycle, and this demo CRD is not in anyone's catalog including mine. What it demonstrates is the mechanism. The API server converts on read, so the written version is not recoverable from the object body; server-side apply, operators, Flux and Argo CD leave no `last-applied` annotation; and `managedFields` retains the version, the writer and the date regardless of how the object was applied.

That is the source `upgrade-check` uses for the 97 catalogued removed APIs, alongside `last-applied` and the apiserver's own `apiserver_requested_deprecated_apis` metric, and each finding says which one produced it. I could not find a reference to managed fields in kubent or Pluto, and neither documents it as a source.

## The other difference

Every check that could not run is printed with a reason and a command to run yourself, and `--strict` turns any gap into a non-zero exit. On a managed control plane a few hundred config rules genuinely cannot be evaluated, and I would rather the report say so than return a shorter list of findings that reads as a pass. A clean report that was clean because half the checks silently did not run is the failure mode I built this to avoid.

Happy to add any of this to the README, or to narrow the plugin's scope if the overlap on removed APIs is still a concern.

PR #6320, opened 2026-09-14
Add crashcause plugin

ryckakas:
- **Add crashcause plugin**
- **file cleaned-up**


ahmetb:
Hi @ryckakas, thanks for submitting **crashcause**! 🎉

A quick heads-up before the details: due to the recent rise of agentic coding tools, we're seeing a much higher rate of plugin submissions, and this puts real stress on our small reviewer community. Please **do not tag maintainers directly** — we'll get to your plugin eventually.

Since this is a **new plugin submission**, it requires a human approval before merging: a maintainer will make the final curation review to decide whether it fits the curated index. Until then, here's my initial review of the manifest with a couple of things to fix.

### Requested changes

1. **Usage strings in `description`**: The manifest contains a usage block:

   ```
   Usage:
     kubectl crashcause inspect <pod> -n <namespace>
     kubectl crashcause inspect <pod> -n <namespace> --output json
   ```

   Krew already instructs users to run `kubectl <plugin>` and links to your plugin's `homepage`, so usage strings should be removed from the manifest. Please keep the `description` focused on what the plugin does and what makes it useful (the cause-code taxonomy, confidence + evidence output, and exit codes are great content to keep).

2. **`shortDescription` redundancy**: "Explain why a Kubernetes pod crashed" — the word "Kubernetes" is unnecessary here, since every plugin in this index is already a kubectl plugin operating on Kubernetes. Something like `Explain why a pod crashed` is shorter and reads better.

### For the human review — overlap with existing plugins

The "why did my pod crash?" problem space is already served by several plugins in the index. Before the final review, please try out the ones below and clarify in a comment how `crashcause` differs from them:

- [crashloop](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/crashloop.yaml) — Inspect pod crash history and previous logs (correlates warning Events, last termination state, and previous logs into one report)
- [why-fail](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/why-fail.yaml) — Explain why a pod is failing, in plain language (deterministic rules for the most common failure modes, optional AI fallback)
- [triage](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/triage.yaml) — Fast diagnostics for failed pods (pod status, warning events, previous crash logs, current logs)
- [kurt](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/kurt.yaml) — Find what's restarting and why
- [crashwatch](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/crashwatch.yaml) — Monitor CrashLoopBackOff pods

Once you've addressed the two items above and commented on how your plugin is different from the existing ones, a human reviewer will take the final look. Thanks for your patience!

ryckakas:
Thanks for the review! I've pushed both fixes:
- Removed the `Usage:` block from `description`
- Trimmed `shortDescription` to "Explain why a pod crashed"

On overlap with the existing pod-failure plugins, happy to clarify how `crashcause` differs from each:

- **crashloop** / **triage**: both produce a one-shot aggregated report (events + last termination state + logs, or a raw status snapshot). `crashcause` classifies the failure into one of 17 stable, versioned cause codes with a confidence score and cited evidence, rather than leaving correlation to the reader.
- **why-fail**: closest in spirit, but its rules are single-purpose plain-language text with an optional OpenRouter-only AI fallback. `crashcause` exposes a structured, machine-readable cause taxonomy (`--output json`) and supports three AI providers (Anthropic, OpenAI, local Ollama) with a fail-closed per-namespace allowlist and log redaction — meant to be safe to point at a real cluster, not just a local demo.
- **kurt**: cluster-wide breadth — which pods are restarting and where. `crashcause` is deep on a single pod: why *this* pod failed, with evidence.
- **crashwatch**: continuous CrashLoopBackOff monitoring with its own dashboard. `crashcause` has an equivalent always-on mode (`watch`), but exports to Prometheus + Loki + Grafana instead of a bespoke UI, and covers all 17 cause codes rather than just CrashLoopBackOff.

Net: same problem space, different bet - structured, versioned classification + evidence, one-off CLI or continuous observability-stack integration, rather than raw aggregation or a single failure mode.

PR #6338, opened 2026-09-19
Add survive-zone plugin

SaiPisey2:
`survive` answers whether a cluster actually survives losing a failure domain.

It reads **real pod placement** rather than declared intent, removes a domain,
and reports which workloads stop serving. A `maxSkew: 3` constraint over three
replicas permits all three in one zone, and `whenUnsatisfiable: ScheduleAnyway`
is advisory, so a cluster can report every workload healthy while a single zone
loss takes two of them down. It separately detects PodDisruptionBudgets that can
never be satisfied, which hang node drains indefinitely.

It also generates remediation, and shows a patch only when two proofs pass
against the cluster's real state: the upstream scheduler's Filter plugins accept
the mutated pod on the cluster's actual nodes, and the survivability computation,
re-run on where the replicas would then land, no longer reports the workload
lost. A fix that schedules but does not solve the problem is shown separately and
labelled as not surviving.

The plugin is read-only: it never evicts, cordons or deletes anything.

### Checklist

- Source is open source, Apache-2.0, and the LICENSE file ships inside each archive.
- Tagged release: `v0.2.0`.
- Installation tested locally from this manifest with `kubectl krew install --manifest=survive.yaml`, then run against a live cluster.
- Manifest is generated by GoReleaser from the release artifacts, so the URLs and `sha256` values come from the published archives.
- Six platforms: linux, darwin and windows on amd64 and arm64.

### On the name

The naming guide discourages generic verbs. `survive` is a verb, so to be
explicit about the reasoning: it is not a verb applied to a Kubernetes resource
and it does not overlap with an existing plugin's space — the nearest entries in
the index are `debug-pdb` and `topology`, which display information rather than
evaluate domain loss. Happy to rename if you would prefer something like
`zone-survival`.

Repo: https://github.com/SaiPisey2/kubectl-survive


ahmetb:
Thanks for the submission, @SaiPisey2! 🎉 Simulating failure-domain loss against real pod placement (rather than declared intent) is a genuinely interesting reliability angle.

**Heads up on process first:** every *new* plugin submission requires a human maintainer's final approval before it can be merged — this comment is just the automated first pass, and a human will later make the call on whether it fits the curated index. Also, due to the rise of agentic coding tools, we're seeing a much higher volume of plugin submissions lately, which puts real stress on our small reviewer community. **Please do not tag maintainers directly** — we'll get to your plugin eventually.

While you wait, here are a few things to address:

### 1. `shortDescription` exceeds the 50-character limit

The current value is 58 characters:

> Report which workloads lose availability if a zone is lost

Krew shows `shortDescription` in tight terminal real estate (`kubectl krew search` / `list` output), which is why we cap it at 50 characters — it's meant to be a tagline, not a full sentence. For example:

- `Check which workloads survive a zone loss` (41 chars)
- `Simulate zone loss to find at-risk workloads` (45 chars)

### 2. Please reconsider the name

I saw (and appreciate) your pre-emptive reasoning in the PR description, but we'd still prefer something more specific than the bare verb `survive`. `kubectl survive` doesn't tell the user *what* survives *what* — node loss? zone loss? an upgrade? a pod eviction? Since the name could be stretched to cover many future use cases, granting it also hands the first submitter a very broad term; this is the same reason we're cautious with generic verbs like `login`, `setup`, or `ai`.

Your own suggestion of **`zone-survival`** reads much better — it names the scope and the evaluation, and `kubectl zone-survival` is self-explanatory. `survive-zone` would also work. The general guidance lives at https://krew.sigs.k8s.io/docs/developer-guide/develop/naming-guide/.

### 3. Closest existing plugins

These aren't duplicates, but they're the nearest neighbors in the index — please give them a try and leave a short comment clarifying how your plugin differs:

- [`topology`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/topology.yaml) — Explore region topology for nodes or pods
- [`debug-pdb`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/debug-pdb.yaml) — Debug Pod Disruption Budgets (PDB)
- [`can-schedule`](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/can-schedule.yaml) — Check whether workloads fit on a cluster

Your PR description already covers the first two (they display state rather than evaluate domain loss), which is helpful. It would also be great if you could address `can-schedule`, since it likewise runs the upstream scheduler's filter plugins against a live, read-only cluster snapshot — the human reviewers will want to understand the boundary between "can this land?" and "does the workload survive a zone loss?".

Once you've shortened the `shortDescription` (and, ideally, renamed the plugin), a human reviewer will take the final call. Thanks again for the detailed PR description — it made reviewing this much easier!

SaiPisey2:
Both points addressed in `v0.2.1`, and I tried the three neighbours.

**1. `shortDescription`** is now 44 characters: `Show which workloads die when a zone is lost`.

**2. Renamed to `survive-zone`.** You were right that the bare verb did not say what survives what. The binary is `kubectl-survive_zone` — kubectl only resolves the hyphenated command from the underscored filename, which I confirmed before settling on it. Repository and Go module path are unchanged.

**3. `can-schedule`** is the one worth being precise about, since it runs the same upstream Filter plugins. Both run against the same three-zone cluster, same moment:

```
$ kubectl can-schedule -f checkout-api.yaml
│ Deployment │ checkout-api │ 6/6 │ ... │ 3/3 │ - │
Result: SCHEDULABLE — 3 of 3 replicas fit

$ kubectl survive-zone
Losing us-east-1a  ->  2 lost, 0 degraded
  checkout-api   us-east-1a:3  LOST all 3 available replicas are in us-east-1a
```

`can-schedule` answers "will this fit?" about a workload you hand it, and it is correct: all three replicas fit on 6 of 6 nodes. `survive-zone` never takes a manifest. It reads where the replicas already are and answers what stops serving when a domain disappears — which no manifest can answer, because constraints bind at scheduling time and placement drifts afterwards.

The scheduler appears here only as a guardrail on remediation: before printing a patch, `survive-zone fix` proves the mutated workload still schedules, because advice that causes Pending pods is the outage the tool was run to prevent. So scheduling feasibility is `can-schedule`'s output and our precondition.

**`debug-pdb`** enumerates coverage; `kubectl debug-pdb pods session-store` listed the one pod. We evaluate satisfiability instead: `minAvailable resolves to 1 of 1 pods; no pod can ever be evicted`. That budget hangs any drain touching it, which is reported separately from domain loss because a PDB offers no protection when a zone's nodes simply vanish.

**`topology`** I could not run: it offers no darwin/arm64 build. From its manifest it displays node and pod topology; it does not evaluate the loss of a domain.

No rush from my side on the human review.

SaiPisey2:
Updated the manifest to `v0.4.0`.

Since the last review the plugin follows Services to their EndpointSlices, so it now flags a workload spread across every zone when a Service it calls has backends in only one. It also gains an optional exporter mode. I kept the name, `shortDescription` and binary as they were, and the manifest installs cleanly with `kubectl krew install --manifest`.

Happy to adjust anything whenever someone has a moment.

